PR #85, opened 2014-11-27
Initial istft implementation.

jfsantos:
I wrote an `istft` function to bring a real signal from its per-frame STFT representation to a time-domain signal by using overlap-add. This version is unoptimized as I am having some trouble to create an `FFTW.Plan` for a backward RFFT. I can add my non-working code to this branch in case you want to take a look. 

I understand the PR is in a crude shape right now but I need some help moving forward, and I figured it would be better to submit a PR than try to solve this via e-mail.

Any suggestions for improvement or tips on how I can use the `FFTW.Plan` interface to do what I want here?

### DIFF
--- a/src/periodograms.jl
+++ b/src/periodograms.jl
@@ -162,19 +162,6 @@ function fft2oneortwosided!{T}(out::Array{Complex{T}}, s_fft::Vector{Complex{T}}
     out
 end
 
-# Evaluate a window function at n points, returning both the window
-# (or nothing if no window) and the squared L2 norm of the window
-compute_window(::Nothing, n::Int) = (nothing, n)
-function compute_window(window::Function, n::Int)
-    win = window(n)::Vector{Float64}
-    norm2 = sumabs2(win)
-    (win, norm2)
-end
-function compute_window(window::AbstractVector, n::Int)
-    length(window) == n || error("length of window must match input")
-    (window, sumabs2(window))
-end
-
 ## PERIODOGRAMS
 abstract TFR{T}
 immutable Periodogram{T,F<:Union(Frequencies,Range)} <: TFR{T}
@@ -207,7 +194,7 @@ function periodogram{T<:Number}(s::AbstractVector{T}; onesided::Bool=eltype(s)<:
     onesided && T <: Complex && error("cannot compute one-sided FFT of a complex signal")
     nfft >= length(s) || error("nfft must be >= n")
 
-    win, norm2 = compute_window(window, length(s))
+    win, norm2 = Util.compute_window(window, length(s))
     if nfft == length(s) && win == nothing && isa(s, StridedArray)
         input = s # no need to pad
     else
@@ -286,7 +273,7 @@ function welch_pgram{T<:Number}(s::AbstractVector{T}, n::Int=length(s)>>3, nover
     onesided && T <: Complex && error("cannot compute one-sided FFT of a complex signal")
     nfft >= n || error("nfft must be >= n")
 
-    win, norm2 = compute_window(window, n)
+    win, norm2 = Util.compute_window(window, n)
     sig_split = arraysplit(s, n, noverlap, nfft, win)
     out = zeros(fftabs2type(T), onesided ? (nfft >> 1)+1 : nfft)
     r = fs*norm2*length(sig_split)
@@ -369,7 +356,7 @@ function stft{T}(s::AbstractVector{T}, n::Int=length(s)>>3, noverlap::Int=n>>1,
                  window::Union(Function,AbstractVector,Nothing)=nothing)
     onesided && T <: Complex && error("cannot compute one-sided FFT of a complex signal")
 
-    win, norm2 = compute_window(window, n)
+    win, norm2 = Util.compute_window(window, n)
     sig_split = arraysplit(s, n, noverlap, nfft, win)
     nout = onesided ? (nfft >> 1)+1 : nfft
     out = zeros(stfttype(T, psdonly), nout, length(sig_split))
--- a/src/util.jl
+++ b/src/util.jl
@@ -21,7 +21,8 @@ export  unwrap!,
         rmsfft,
         unsafe_dot,
         polyfit,
-        shiftin!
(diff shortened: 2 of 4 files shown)


jfsantos:
Oh, it seems I did something terrible while rebasing. I thought I had squashed everything instead of making this mess. Any tips on how to fix this?

src/util.jl
```diff
@@ -79,6 +79,57 @@ function hilbert{T<:Real}(x::AbstractArray{T})
...
+function istft{T<:Union(Float32, Float64)}(S::AbstractMatrix{Complex{T}}, wlen::Int, overlap::Int; nfft=nextfastfft(wlen), window::Union(Function,AbstractVector,Nothing)=nothing)
+    winc = wlen-overlap
+    win, norm2 = compute_window(window, wlen)
+    if win != nothing 
+      win² = win.^2
+    end
+    nframes = size(S,2)-1
+    outlen = nfft + nframes*winc
+    out = zeros(T, outlen)
+    tmp1 = similar(S[:,1])
```
simonster: This would should be `Array(eltype(S), size(S, 1))` to avoid making an unnecessary copy and to ensure that the array is actually an Array (which is not necessarily the case if `S` is an AbstractMatrix).

src/periodograms.jl
```diff
@@ -162,17 +162,18 @@ function fft2oneortwosided!{T}(out::Array{Complex{T}}, s_fft::Vector{Complex{T}}
...
+    sumabs2(x) = Base.sumabs2(x)
+else
+    function sumabs2(s)
+        x = zero(eltype(s))
+        for i = 1:length(s)
+            @inbounds x += abs2(s[i])
+        end
+        x
+    end
 end
```
simonster: I removed this in 5c016833c48cd622f949ea458a4a886c9f590741 since we now require 0.3+ and so we don't need it. It looks like it came back from the dead during the rebase? :dizzy_face:

jfsantos: Grab the shotguns, there's zombie code in this PR! Sorry about that. I'll remove it again.

src/util.jl
jfsantos: Thanks, will fix!

simonster:
Not entirely sure what's going on with the history here, but the diff looks right. I guess the nuclear option is:

```sh
git diff master > patch.diff
git reset --hard master
patch -p1 < patch.diff
rm patch.diff
```

and then commit and force-push.

jfsantos:
This should do the trick, thanks @simonster! Let's wait until the tests pass and then finally merge this?

jfsantos:
Any more comments on this? @simonster @gummif @r9y9

standarddeviant:
What ever happened to this?  This came up in a discourse thread re: `istft` in DSP.jl.

jfsantos:
@standarddeviant I ended up never being able to finish this PR and in the meantime, Julia changed a lot and I was not using it on a daily basis anymore. Please feel free to do whatever is needed to get this to work with newer versions of Julia and submit another PR.

davidavdav:
I started to write an istft as well, today, until I found out this PR and the MusicProcessing reference mentioned above.  The latter turned out to have an alternative implementation of [MFCC](https://github.com/davidavdav/MFCC.jl) as well.  

It would be useful, indeed, to have this `istft` functionality added to DSP.  

While implementing my own little `istft`, I noticed that `DSP.arraysplit()` seems to only keep "whole" segments, dropping the last samples of the source array.  If we would strive for perfect reconstruction possibility with `istft`, `arraysplit` should probably have an option to augment the input.

While we're at it, it would also be great to have an option to `arraysplit` in such a way that the overlap is symmetric on the left and right of each frame, even for the first and last frame.   It may be best to illustrate this with an example.  Suppose I have 1000 samples, and I want frame-shifts of 100 samples and an analysis window of 300 samples.  Then the first frame would be samples -100..200, the second frame would be sampes 0..300, etc until the last frame 800..1100 (give or take a Julia base 1 index offset).  Negative and >size indexes would refer to augmented data, e.g., zeros. 

The advantage would be that such a scheme would have 1000/100 = 10 frames, so that the frames are more easy to align with the original samples (first frame aligned with 0..100, the middle of the analysis window, etc.).  The analysis would be symmetric w.r.t. start and end, of the signals, and of the frames themselves. 

In [Kaldi](https://github.com/kaldi-asr/kaldi) a modern speech recognition toolkit, this is known as the analysis option `snip-edges=true`.

galenlynch:
It seems a little bit unnatural to me to circularly wrap a signal by default. When I apply stfts to data, its usually because it's non-stationary data, and the first and the last frames end up being quite different from each other.

@davidavdav, would you have any interest in continuing this PR?

bafonso:
gist [here](https://gist.github.com/bafonso/2ea72ca3e03e0f2d249591437347563c)
Hi, I'd like to help to get the istft implemented in DSP. I am trying to bring back MusicProcessing and I need a working istft and I see others have shown interest in having this in the DSP library. I am trying to familiarise myself with the DSP.jl code and I am sharing the code inspired on this PR and made some changes so it plays nicely with current DSP.jl code. I tried to follow the same approach but I am sure I am missing several details.. help please? :-)

Right now it appears to work ok using a hanning window but wrongly scaled at the beginning when window is "nothing". I am not sure how to properly address this, perhaps the scaling I see in older code?

```julia
@test x1 ≈ Y1
```
fails even with hanning window, I believe because of discrepancies at the the first sample.

davidavdav:
> It seems a little bit unnatural to me to circularly wrap a signal by default. When I apply stfts to data, its usually because it's non-stationary data, and the first and the last frames end up being quite different from each other.
> 
I don't think signals under a stft transform should be circularly wrapped, either.  Where is that suggested and/or happening?

> @davidavdav, would you have any interest in continuing this PR?

I think that if we have a `stft`, there should be a potentially perfectly reconstructing `istft`.

galenlynch:
> I don't think signals under a stft transform should be circularly wrapped, either. Where is that suggested and/or happening?
Good question, I don't know where I got that from.

PR #90, opened 2014-12-21
Merge Multirate into DSP

JayKickliter:
This is not even close to working yet. I'm just opening this pull request to get the process started.  I will continue to integrate when I can find time.

There are two separate sets of functionality in is pull request:
* Streaming multirate filtering
* Window based filter design

I already know that the DSP.jl developers are trying to get away from enums, which is used in the filter design code. I have no strong opinion on that code, and would happy if someone else wants to run with it.

The actual streaming code is my priority to fully merge. There's some redundant code for the different filt[!] methods. In the future it might make sense to combine them all into one with different work subroutines, but for now I will concentrate on getting the code to work within DSP.jl.


JayKickliter:
Progress:

FIR Filter Design
- [x] Get firdes to run and export
- [ ] Write tests for firdes
- [ ] Implement suggested changes

Streaming FIR Filter Execution
- [x] Get filt[!] to run and export
- [x] Write fit[!] tests
- [ ] Implement suggested changes

simonster:
We should figure out a better way to name things. As I noted [on julia-dsp](https://groups.google.com/forum/#!topic/julia-dsp/GYPcaUSCR-8), we need to differentiate between filter coefficients and filter implementations (i.e. coefficients, state, and an implied algorithm). Having TFFilter/ZPKFilter/SOSFilter objects that don't have any state and an FIRFilter object that holds a state seems like it would be confusing. I'm not sure what the best naming convention is, but I'm open to suggestions.

JayKickliter:
I struggled with this myself. I don't know what to call it. Since the current implementation of FIRFilter is just a small wrapper around a FIRKernel, maybe just rename the whole thing FIRKernel?

simonster:
I'm leaning toward renaming the existing types that hold coefficients without state. We could replace `Filter` with `Coefficients` or spell out the whole names (`TFFilter` -> `TransferFunction`, `ZPKFilter` -> `ZeroPoleGain`, `BiquadFilter` -> `Biquad`, `SOSFilter` -> `SecondOrderSections`). Then you could leave `FIRFilter` as you have it now, and we could eventually have `TDF2Filter` and `TDF2SOSFilter` (or maybe `DF2TFilter` and `DF2TSOSFilter`?) objects that hold state for IIR filters.

jcbsnd:
What a great addition to the DSP package, props to @JayKickliter! If the existing types should be renamed as suggested by @simonster, is there anything preventing that this PR is accepted as is?

JayKickliter:
Thanks @jcbsnd. I have almost 0 time to work on this at the moment, but I was hoping some other people could us it first.

PR #96, opened 2015-02-21
Rename filter types to distinguish coefficients from implementations

simonster:
Per my comment [here](https://github.com/JuliaDSP/DSP.jl/pull/90#issuecomment-68806557)

ZPKFilter -> ZeroPoleGain
TFFilter -> TransferFunction
BiquadFilter -> Biquad
SOSFilter -> SecondOrderSections


simonster:
Or should we do TFFilter -> PolynomialTF or PolynomialRatio?

goretkin:
+1 for something like PolynomialRatio

PR #101, opened 2015-04-08
Merge Multirate, pt. 2

simonster:
@JayKickliter I did some work to try to get your multirate changes ready to merge. Overall, the code looks really good. The biggest thing I did was to change around the filter design API to make it more like the IIR design APIs. I also changed the code style to match what the rest of the package (I hope you don't mind) and fixed some deprecation warnings. If you disagree with any of this or if you see anything I've broken, let me know.

I'll look over this again and I may make a few more changes. Thanks again for all your work on this!


simonster:
Now that I can get the Travis build to complete, it also looks like there are a lot of code paths that aren't covered by the tests. (It seems this wasn't apparent in the Coveralls report for #90 because we didn't used to consider lines in completely untested functions as relevant in the coverage report, but now we do.) @JayKickliter Do you have tests for these in Multirate.jl that I could move here?

JayKickliter:
@simonster Thanks for the help. I have a bit of reading problem, which is why I have those ascii banners and extra white space between functions. However, I value constancy over my personal style, so I don't mind the changes at all if it will help to get the code in the library.

I'll add some tests to bring the coverage back up. I also really need to generate some reference results from a third party tool using identical input and coefficient vectors.

Thanks again for the assistance. I will try to get the tests written this weekend.

simonster:
I have MATLAB and can help with the reference results, unless you already have a strategy. Just let me know.

JayKickliter:
Thanks. Using Matlab as a reference would be preferable. I think I still have Matlab access, if not, I'll just let you know what we need.

JayKickliter:
I just cut out a bunch of code on a different branch. It should increase coverage significantly. The biggest contributor to the low coverage % was the lack of tests for the Farrow filter. It seems to perform worse both numerically and speed-wise than FIRArbitrary, so I got rid of it.

I also got rid of some helper functions that weren't being used.

I tried to push it to JuliaDSP, but it appears I don't have push access. I'm still an idiot when comes to git/GitHub, and can't seem to get travis/coveralls to run it's test from my repo. So I'm not entirely sure what the new coverage is.

simonster:
@JayKickliter Sorry, I thought I already gave you push access, but apparently not. You should have it now.

JayKickliter:
@simonster what else needs to be done before we can merge? I'll have a little free time this week to work on it.

simonster:
It would be good to have correctness tests for the window-based FIR filter functions. Otherwise, this looks good to go.

JayKickliter:
I forgot about that. I'll add them.

JayKickliter:
I'm in the process of adding those tests. I noticed almost everyone uses β for kaiser windows, but we use α. Should we switch to β to more consistent?

simonster:
I don't have a strong opinion on this. The Wikipedia page uses α, but it looks like MATLAB and NumPy use β.

simonster:
I looked into FIR filter design myself so as not to keep pestering you. It seemed like bandpass and bandstop design were effectively swapped, and I now avoid computing `sinc` in places where it will only be 0 or 1. Please double-check to make sure that this is right, but I guess it should be since it appears to match SciPy.

I think we're almost ready to pull the trigger, but there are a few other questions that have arisen in my mind:

- As I guess you've noticed, SciPy has a `scale` kwarg that rescales the filter so that it gives unity response somewhere, and it's enabled by default. Is this something we should have as well?
- Perhaps `FIRWindow` is a better name than `WindowFIR`? What do you think?
- I noticed that `naivefilt` isn't documented. Am I correct in assuming it's only really useful for checking the correctness of the more efficient multirate implementations, and we could just move it to the tests?

JayKickliter:
It looks like my poor git skills created a conflict. I'm sorry about that. I should have been working on another branch since you created this one.

I'll have to look closer later, but it looks like we pretty much implemented the same exact stuff. It's like you read my mind.

JayKickliter:
Nevermind. Of course git wouldn't let me push if there was a conflict. The test I was seeing was from my personal repo. One thing I did was to add the python code to the repo for generating the reference taps and writing the files, for future reference. 

And yes to all your questions. Scaling should be done by default. `FIRWindow` is a good idea. And `naivefilt` is only for testing. I haven't checked Matlab in a while, but last time I did (on an older version) I couldn't find any good options for polyphase filtering with identical behavior to my code.

simonster:
Sorry, I didn't realize you had made further changes! I think you should be able to `git pull` and get the remote branch merged into yours, although if we changed the same lines, there will probably be conflicts that need to be resolved. You can also make a new branch using `git checkout -b newbranch` and push that to DSP and leave the merge to me. Whichever you prefer.

simonster:
I've implemented those three changes. Do you want me to push now, or will that cause further issues?

JayKickliter:
Push away. Thanks.

JayKickliter:
Those changes look good. Thanks. I don't think I have anything immediate to add. Unless you did and I missed it, I'd like to add a python script for generating the reference vectors. But I'm neck-deep in work so that can wait till later, if you want to get this merge out of the way.

simonster:
The Python script would be great to have, but I agree it can wait until after the merge. Since this is almost entirely your work, I think you should have the honor of pressing the big green button :smile:. Thanks for sticking with this, and sorry it took so long!

JayKickliter:
Thanks. I'm pretty sure it's mostly on me for taking so long. I ain't mad.

PR #102, opened 2015-04-22
Use new stateful filter API for filtfilt

simonster:
This provides reduces the memory allocation and provides a small speed boost (on 0.4; on 0.3, where SubArrays are slow, it's probably substantially slower). We could conceivably use this to provide in-place filtfilt, although that is not yet implemented.

I removed support for non-1 initial values of a in filt_stepstate, since we don't need it if the filter is normalized before it is used.


simonster:
It seems there is a bug in SubArrays on 0.3 that prevents this from working at all there.

PR #105, opened 2015-05-06
Speed up FIR filtering

simonster:
- Implement fast path for stateless filtering
- Use `BLAS.dot()` to compute sufficiently large dot products (this also benefits stateful/multirate filters)
- Make `filt(h, x)` use `fftfilt` when it should be faster than naive convolution, replacing `firfilt`

Here's a performance comparison with `DSP.filt()` from the original multirate commit (top) and `Base.filt()` (bottom). It's slower than `Base.filt()` only for very small filter lengths (is this worth optimizing further?), and much faster for filters with 16+ taps.

perf

### DIFF
--- a/doc/filters.rst
+++ b/doc/filters.rst
@@ -99,7 +99,9 @@ Filter application
     to zeros). If ``f`` is a ``PolynomialRatio``, ``Biquad``, or
     ``SecondOrderSections``, filtering is implemented directly. If
     ``f`` is a ``ZeroPoleGain`` object, it is first converted to a
-    ``SecondOrderSections`` object.
+    ``SecondOrderSections`` object.  If ``f`` is a Vector, it is
+    interpreted as an FIR filter, and a naïve or FFT-based algorithm is
+    selected based on the data and filter length.
 
 .. function:: filt!(out, f, x[, si])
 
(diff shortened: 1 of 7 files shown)


jfsantos:
I think this is definitely worth having. Regarding the small filter slowdown, did you check the native code for filters with those sizes to see what the backend is doing in Base.filt and DSP.filt? Maybe some kind of optimization with SIMD instructions?

simonster:
There's probably just too much constant overhead in the outer loop for it to be fast for small filters. There are branches for filter size that would be better placed outside the outer loop and we have to load things that don't change from memory because there could be a function call. If we want to optimize the small cases, there are a few ways we could make things faster:

- Generate specialized code for the small cases. I think we could turn the filter into a tuple and pass that to a function that gets specialized on the tuple, and LLVM should completely unroll the inner loop and keep the filter coefficients in registers.
- Just call `Base.filt` for the small cases. This is obviously the easiest to implement, and at least ensures we don't do any worse than Base.
- Branch outside the outer loop for small cases and use an inner loop with less overhead. I'm not sure if this would actually be faster than `Base.filt`, but we could see.

simonster:
Here it is with specialized code for sizes 1-15.

perf

PR #108, opened 2015-05-13
Allow scaling and composing filter coefficients using *

simonster:
Also fill in some convert methods and promotion types

### DIFF
--- a/src/Filters/coefficients.jl
+++ b/src/Filters/coefficients.jl
@@ -2,6 +2,11 @@
 
 abstract FilterCoefficients
 
+realtype(x::DataType) = x
+realtype{T}(::Type{Complex{T}}) = T
+complextype(T::DataType) = Complex{T}
+complextype{T}(::Type{Complex{T}}) = Complex{T}
+
 #
 # Zero-pole gain form
 #
@@ -11,6 +16,19 @@ immutable ZeroPoleGain{Z<:Number,P<:Number,K<:Number} <: FilterCoefficients
     p::Vector{P}
     k::K
 end
+
+Base.promote_rule{Z1,P1,K1,Z2,P2,K2}(::Type{ZeroPoleGain{Z1,P1,K1}}, ::Type{ZeroPoleGain{Z2,P2,K2}}) =
+    ZeroPoleGain{promote_type(Z1,Z2),promote_type(P1,P2),promote_type(K1,K2)}
+Base.convert{Z,P,K}(::Type{ZeroPoleGain{Z,P,K}}, f::ZeroPoleGain{Z,P,K}) = f
+Base.convert{Z,P,K}(::Type{ZeroPoleGain{Z,P,K}}, f::ZeroPoleGain) =
+    ZeroPoleGain{Z,P,K}(f.z, f.p, f.k)
+
+*(f::ZeroPoleGain, g::Number) = ZeroPoleGain(f.z, f.p, f.k*g)
+*(g::Number, f::ZeroPoleGain) = ZeroPoleGain(f.z, f.p, f.k*g)
+*(f1::ZeroPoleGain, f2::ZeroPoleGain) =
+    ZeroPoleGain([f1.z; f2.z], [f1.p; f2.p], f1.k*f2.k)
+*(f1::ZeroPoleGain, fs::ZeroPoleGain...) =
+    ZeroPoleGain(vcat(f1.z, [f.z for f in fs]...), vcat(f1.p, [f.p for f in fs]...), f1.k*prod([f.k for f in fs]))
 
 #
 # Transfer function form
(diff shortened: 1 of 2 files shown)


simonster:
Any objections to this? On one hand, chaining filters is effectively multiplying their transfer functions, so overloading `*` seems reasonable. On the other hand, we can't overload `*` for FIR filters unless we wrap them in a type instead of representing them as bare vectors.

jfsantos:
I don't have any objections to this. Concerning your second point, I think wrapping FIR filters in a type makes sense and helps one's code to be cleaner. We have types for frequency-domain representations, TF-representations, IIR filters... having types for FIR filters as well will kind of make bare vectors in one's code be most likely raw signals.

JayKickliter:
I don't want to derail the the conversation too much, but I've been thinking about point (2) lately. I think we really need to make dealing with IIR and FIR filters more consistent. Example: I wanted to sanity check the response of some FIR taps yesterday, so I tried `freqz`, which didn't work of course since all the `freqz`  takes type `FilterCoefficients` as it's primary argument. We could make a new method that dispatches on an `AbstractVector`, but that would create even more inconstancy.

To argue the other side, for some things in `DSP.jl`, I find it typed to the point of being slightly difficult to use. `Periodogram` comes to mind. Every time I want to plot spectrum of a signal this happens:

```jlcon
julia> pg = periodogram(x)

julia> plot(pg)  # Oops, forgot that the result isn't just a vector
ERROR: `plot` has no method matching plot(::Periodogram{Float64,Frequencies})

julia> plot(z.freq,z.power)
Error showing value of type FramedPlot:
ERROR: `.*` has no method matching .*(::Float64, ::Frequencies)

julia> methodswith(Frequencies)
8-element Array{Method,1}:
 done(x::Frequencies,i::Int64) at /Users/jaykickliter/.julia/v0.3/DSP/src/util.jl:134                  
 size(x::Frequencies) at /Users/jaykickliter/.julia/v0.3/DSP/src/util.jl:135                           
 step(x::Frequencies) at /Users/jaykickliter/.julia/v0.3/DSP/src/util.jl:137                           
 fftshift(x::Frequencies) at /Users/jaykickliter/.julia/v0.3/DSP/src/util.jl:141                       
 getindex(x::Frequencies,i::Int64) at /Users/jaykickliter/.julia/v0.3/DSP/src/util.jl:129              
 next(x::Frequencies,i::Int64) at /Users/jaykickliter/.julia/v0.3/DSP/src/util.jl:133                  
 similar(x::Frequencies,T::Type{T<:Top},args...) at /Users/jaykickliter/.julia/v0.3/DSP/src/util.jl:136
 start(x::Frequencies) at /Users/jaykickliter/.julia/v0.3/DSP/src/util.jl:132                          
```

go read the documentation...

```jlcon
julia> plot(freq(pg), power(pg))
error during Tk callback: 
ERROR: `.*` has no method matching .*(::Float64, ::Frequencies)
```

Then give up and just plot the power. 

Another example is creating FIR taps.  `digitalfilter(Lowpass(ƒc), FIRWindow(hamming(hLen))`. The `FIRWindow` type seems to only exist to dispatch on. An alternative could be to have window functions return a `Window` type, but may just move the problem somewhere else. Since you guys know much more about Julia than me, would it make sense to have FIR taps/Windows be a subtype of AbstractVector?

I don't have any immediate answers, so I'm just another person on the internet with an opinion and no solution. But I can see how someone using the package for the first time could get frustrated by all the typing.

simonster:
@JayKickliter It's always good to have API feedback.

For better or for worse, the space of possible APIs for the periodogram function is restricted by practical constraints. We don't want to compute the frequencies when the user doesn't need them, we want the function to be type-stable, and Julia doesn't have optional output arguments. This leaves two possibilities. We can offer a separate function to compute the frequencies that needs to get the same set of arguments as `periodogram`. Or we can have our own type that represents the frequencies so that we can return them with the power but still not compute them until they're used. The latter seemed more Julian and probably more usable which is why we have it, but neither option seems totally ideal.

It is unfortunate that `.*` is not defined for the Frequencies type in your example above. This will be fixed for us in Julia 0.4, which defines a full set of arithmetic operators for generic AbstractArrays in Base. But we could also put the appropriate definitions in DSP.jl for now.

We do have a choice between returning a Periodogram object or a (power, frequency) tuple. The main benefit of the former is that it's an appropriate thing to dispatch on. At present, we don't really take advantage of this. (We have an `fftshift` method for Periodogram objects, but I think that's pretty much it.) But being able to use dispatch means that someday we can make `plot(periodogram(x))` just work out of the box with the common plotting packages (and get the frequency axis right). There's an [open issue](https://github.com/JuliaLang/julia/issues/2025) and [PR](https://github.com/JuliaLang/julia/pull/6884) on making this work in a non-hackish way, but I've also wondered whether we should just use one of the existing hacks since it's a pretty common thing to do.

For filter design, I'm not so sure our current API is ideal. It seemed to make sense for IIR filters with a fixed number of poles, which was the only thing I initially implemented, but now I'm not so sure it makes sense for FIR filters or when designing filters with automatic order selection. A somewhat more traditional approach like `butter(Lowpass(0.5), 2)` and `firwin(Highpass(0.5), hamming(hLen))` might be easier to work with and more flexible.

I'm not sure if windows need to be their own types if we change the design API. It sounds kind of like FIR taps should be, since then we can have `FIRTaps <: FilterCoefficients` and we can implement `*` for chaining.

PR #110, opened 2015-05-20
resample function

### LINKED ISSUE #104: resample function
Now that we have @JayKickliter's excellent multirate work, it would be great to have a high-performance resampling function. I think this works to design an appropriate filter:

```julia
function resample_filter(ratio::Rational{Int}, transitionwidth, attenuation)
    # Determine resampling filter order
    L, beta = kaiserord(transitionwidth/num(ratio), attenuation)

    # Set cutoff so that we will attain `attenuation` dB rolloff
    cutoff = 1/max(num(ratio), den(ratio))-2*transitionwidth/num(ratio)

    # Design filter
    h = digitalfilter(Lowpass(cutoff), FIRWindow(kaiser(L, beta)))
    scale!(h, num(ratio))
end
```

The tricky part appears to be applying the filter such that the first output sample after ramp up corresponds to the first input sample (and the center of the filter).

JayKickliter:
Criticism welcome. I still haven't found a third-party reference to test the arbitrary rate `resample` function against. 

Closes #104


test/resample.jl
```diff
@@ -0,0 +1,53 @@
...
+# sig(t) = [(1 + sin(2π*0.005*t)) * sin(2π*.05*t) for t in t]
+x_ml   = readdlm(joinpath(dirname(@__FILE__), "data", "resample_x.txt"),'\t')[:,1]
+
+#
+# [y1,b1] = resample(x, 1, 2)
+#
+rate   = 1//2
+h1_ml  = readdlm(joinpath(dirname(@__FILE__), "data", "resample_taps_1_2.txt"),'\t')[:,1]
+y1_ml  = readdlm(joinpath(dirname(@__FILE__), "data", "resample_y_1_2.txt"),'\t')[:,1]
+ty1_ml = [0:length(y1_ml)-1] / float64(rate)
```
simonster: This triggers a deprecation warning on Julia 0.4, where you need `[a;]` to convert an AbstractVector to a Vector and `float64(x)` has been changed to `Float64(x)` (actually `@compat Float64(x)` to maintain 0.3 compatibility). But it looks like these time vectors aren't used and could just be removed?

src/Filters/stream_filt.jl
```diff
@@ -151,6 +155,47 @@ function FIRFilter(h::Vector, rate::FloatingPoint, Nϕ::Integer=32)
...
+    FIRFilter(h, rate)
+end
+
+
+#
+# setphase! set's filter kernel phase index
+#
+function setphase!(kernel::FIRDecimator, ϕ::Real)
+    @assert zero(ϕ) <= ϕ
+    xThrowaway = int(ϕ)
```
simonster: On 0.4, this should be `round(Int, ϕ)` or `@compat Int(ϕ)`, depending on whether you want to round it or throw an error if it's not exactly convertible to an Int.

src/Filters/stream_filt.jl
```diff
@@ -562,18 +630,42 @@ end
...
+    # Use setphase! to
+    #   a) adjust the input samples to skip over before producing and output (integer part of τ)
+    #   b) set the ϕ index of the PFB (fractional part of τ)
+    setphase!(self, τ)
+
+    # Calculate the number of 0's required so that w
+    outLen      = ceil(Int, length(x)*rate)
+    reqInlen    = inputlength(self, outLen)
+    reqZerosLen = reqInlen - length(x)
+    xPadded     = [x, zeros(eltype(x), reqZerosLen)]
```
simonster: On 0.4 this should be `[x; zeros(eltype(x), reqZerosLen)]` to avoid a deprecation warning, but perhaps you could avoid copying `x` with something like:

```julia
out = Array(Base.promote_eltype(x, h), outLen)
n = filt!(out, self, x)
filt!(sub(out, n+1:outLen), self, zeros(eltype(x), reqZerosLen))
out
```

src/Filters/stream_filt.jl
```diff
@@ -151,6 +155,47 @@ function FIRFilter(h::Vector, rate::FloatingPoint, Nϕ::Integer=32)
...
+    @assert zero(ϕ) <= ϕ
+    xThrowaway = int(ϕ)
+    kernel.inputDeficit += xThrowaway
+    nothing
+end
+
+function setphase!(kernel::Union(FIRInterpolator, FIRRational), ϕ::Real)
+    @assert zero(ϕ) <= ϕ
+    (ϕ, xThrowaway) = modf(ϕ)
+    kernel.inputDeficit += int(xThrowaway)
```
simonster: Same as above here.

src/Filters/stream_filt.jl
```diff
@@ -151,6 +155,47 @@ function FIRFilter(h::Vector, rate::FloatingPoint, Nϕ::Integer=32)
...
+    kernel.ϕIdx = floor(ϕ*(kernel.Nϕ-1.0) + 1.0)
+    nothing
+end
+
+function setphase!(kernel::FIRArbitrary, ϕ::Real)
+    @assert zero(ϕ) <= ϕ
+    (ϕ, xThrowaway) = modf(ϕ)
+    kernel.inputDeficit += int(xThrowaway)
```
simonster: Same as above here.

simonster:
This looks great! It would be good to have some tests for the filter design methods, maybe just using `freqz` to make sure that the attenuation and bandwidth are similar to what we expect?

As far as arbitrary resampling goes, how close is the output to rational resampling? Maybe we could just use `@test_approx_eq_eps` to compare it with the rational results with an appropriately selected epsilon?

I'm not generally too picky about this, but It might also be good to rebase the PR so that there are only a few commits.

JayKickliter:
I was going to ask if I should squash the commit history, so that answers that question.

I'm not sure how similar the arbitrary and resample perform. I'll look into it.

Joseph Gaeddert re-released LiquidDSP under MIT license this week. I may be able to use some of his tests. The code, and particularly the [documentation](http://liquidsdr.org/doc/filter.html), could be a huge boon for us.

PR #120, opened 2016-01-31
IIR equalizer biquad filters

seebk:
This pull request implements a function for direct design of typical IIR equalizer filters. It is based on the formula and derivations from the DAFx book (Zölzer, 2011). Available filter types are lowpass, highpass, bandpass, bandstop, low shelving, high shelving and peak.

Please review and tell me if think this is interesting for inclusion into the main branch and if changes are required. If you consider merging, I will also add a few tests...


ssfrr:
I'd love to see these filters in DSP.jl.

One style comment - I think it might be nicer to have separate functions for the different filter types rather than a string argument. Perhaps `IIRHighPass`, `IIRLowPass`, `IIRHighShelf`, etc.?

Issue #116, opened 2015-10-12
Inline documentation using Julia 0.4 new doc system


jfsantos:
Julia 0.4 has a [new documentation system](https://github.com/JuliaLang/julia/pull/8791) which allows adding inline docs that can be accessed via the REPL. We should make sure all the functions exported by the DSP.jl module have inline docs. They do not need the same level of detail as our online docs, but should at least be slightly useful.

simonster:
Yes, this is a good idea. Ideally we'd be able to use the same docs for inline help and the website, but AFAIK there's no general way to do that yet. @shashi has a script [here](https://gist.github.com/shashi/5b2c8cd95d2f3834f9c7), but it doesn't support sections like our current sphinx/rst docs. We could also try to adapt [genstdlib.jl](https://github.com/JuliaLang/julia/blob/master/doc/genstdlib.jl) from Base, although that's clearly a bit of a hack.

jfsantos:
Do you think we should maybe port our current docs to inline docs by hand and then figure out a way of extracting it to rst/md? I could start working on that in a separate branch.

I've seen some people argue that the online docs should be an expanded version of the inline docs, but I don't have a strong opinion on it. I'm inclined to agree in cases where the documentation for each function is huge, includes several references and such, but ours is pretty succinct and should work well for both formats.

PR #128, opened 2016-07-13
Adding delay calculators (delay-utils) to utils.jl

CrocoDuckoDucks:
Hi there!

I wrote few simple methods similar to Matlab's [finddelay](http://uk.mathworks.com/help/signal/ref/finddelay.html?requestedDomain=www.mathworks.com) and [alignsignals](http://uk.mathworks.com/help/signal/ref/alignsignals.html). They are very simple, reflecting that I am Julia newbie. I make massive use of these methods and I though someone else might like them, so here my Pull Request. I have also other methods working with N-dimensional arrays, but [a bug in xcorr](https://github.com/JuliaLang/julia/issues/17351) prevents them to be fast and efficient. Hope you find the methods an interesting suggestion for your project.


jfsantos:
These are interesting, could you add some tests as well?

CrocoDuckoDucks:
Yep! I have been working on few tests, I just have to polish them up and make them a little more complete. I will commit soon.

CrocoDuckoDucks:
Hi there, I pushed the test I was working on. It is the first time I work with tests, so let me know whether there is something else I should do.

src/util.jl
```diff
@@ -288,4 +293,92 @@ function shiftin!{T}(a::Vector{T}, b::Vector{T})
...
 
+
+# Delay Utility Functions:
+
+# finddelay(): use peak of cross correlation from origin of time to calculate
+# time lag between two Arrays (signals)
+
+function finddelay{T <: Real}(x::AbstractArray{T, 1}, u::AbstractArray{T, 1})
+
+sₓᵤ = xcorr(x, u)
```
codles: The convention in the rest of the package is to indent the content of functions.

src/util.jl
```diff
@@ -288,4 +293,92 @@ function shiftin!{T}(a::Vector{T}, b::Vector{T})
...
+
+function shiftsignals!{T <: Real}(u::AbstractArray{T, 1}, δ::Int)
+
+lᵤ = length(u)
+
+if δ > 0
+
+    deleteat!(u, 1:δ)
+    
+    # append!() could be used, but this is faster and prevents allocation.
```
codles: Nice comments

codles:
Nice tests and comments. Would it be possible to add Inline documentation as discussed in issue #116?

CrocoDuckoDucks:
Yes, I never did it but looks straightforward. I will start working on it as soon as I am back from holidays early next week

CrocoDuckoDucks:
Sorry guys, I underestimated how much of a noob I am. At first I imagine that I was supposed to add documentation directly in src/util.jl following [this](http://docs.julialang.org/en/release-0.4/manual/documentation/). Not sure is that the case tho, as after skimming through the rest of the source code I couldn't see similar lines. I am afraid I have to ask what is the correct way to add inline documentation.

jfsantos:
We currently don't have inline docs for the library (this is the reason #116 is still open!). We currently use the documentation from the `docs` folder to automatically generate the documentation pages at readthedocs.

CrocoDuckoDucks:
Hi there cool people!

It has been a while since I had feedback on this PR, so I was wondering whether I completed it correctly on my side. Maybe I did not mention it, but it is my first PR so I might have missed some step...

galenlynch:
Hi @CrocoDuckoDucks do you have any interest in rebasing this and updating it for Julia 1.0?

CrocoDuckoDucks:
Hi @galenlynch, sorry for the very late reply, I missed all of my notifications for a couple of weeks...

I guess I could have a go. I'll see what I can do and revert back.

galenlynch:
Awesome, thank you! I'll review this soon.

CrocoDuckoDucks:
@galenlynch Good stuff, thanks. I will address the various changes  in the next few days.

CrocoDuckoDucks:
So, I changed a few things.

1. Fixed the typos.
2. Since we prefer to mutate the first argument in mutating functions, I changed all the function signatures so that mutating and not mutating versions are used in the same way.
3. Using `argmax` and `abs` to locate the absolute peak of the cross-correlation.
4. Improved documentation and added examples and references.
5. Removed unicode characters.
6. Removed unnecessary parametric types.
7. Redundancy removed as per suggestions.

It looks like tests are now failing for Julia 0.6. Is this a problem?

galenlynch:
Ok, this PR looks great to me now. Once the tests pass on 0.6, I will be more than happy to merge it.

galenlynch:
@CrocoDuckoDucks you should be able to fix the errors on 0.6 by using `Compat.mapreduce`. Sorry for suggesting something that didn't work!

CrocoDuckoDucks:
@galenlynch No problem, I went for a bit of trial and error. Sorry if that is making a bit of noise with notifications and the like. I am not entirely sure the behaviour of `finddelay` is what I want. I will do some experiments locally and report back.

galenlynch:
No problem! Now that I think about it, I don't really understand why you are taking the absolute value of the correlation in the first place, it seems like only positive lags would be desired. Also, there's another bug in my suggested implementation (finding the minimum lag, instead of the minimum absolute value of lag).

Here's my suggestion with all the bug fixes:
```julia
function finddelay(x::AbstractVector{<: Real}, y::AbstractVector{<: Real})
    s = xcorr(y, x)
    max_corr = maximum(s)
    max_idxs = Compat.findall(x -> x == max_corr, s)

    center_idx = cld(length(s), 2)
    # Delay is position of peak cross-correlation relative to center.
    # If the maximum cross-correlation is not unique, use the position
    # closest to the center.
    d_ind = Compat.argmin(abs.(center_idx .- max_idxs))
    d = center_idx - max_idxs[d_ind]
end

```

CrocoDuckoDucks:
> No problem! Now that I think about it, I don't really understand why you are taking the absolute value of the correlation in the first place, it seems like only positive lags would be desired.

Often one wants to find the delay of some kind of filter, which could have negative gain. If we pick the absolute peak of the cross-correlation we find the delay also in that case. For example:
```julia
julia> finddelay([0, 0, -1, -2, -3], [1, 2, 3])
2
```
So, the delay was correctly calculated also in case of phase inversion. Instead if we peak the maximum:
```julia
julia> finddelay([0, 0, -1, -2, -3], [1, 2, 3])
-1
```

CrocoDuckoDucks:
I committed a revision I am happy with. Let me know what you think.

galenlynch:
It seems you've changed `finddelay` to find the lag with maximum correlation, instead of maximum absolute correlation. From our last conversation, it sounded like you want it to be absolute correlation?

src/util.jl
```diff
@@ -336,4 +341,117 @@ function shiftin!(a::AbstractVector{T}, b::AbstractVector{T}) where T
...
+# Example
+```jldoctest
+julia> finddelay([0, 0, 1, 2, 3], [1, 2, 3])
+2
+
+julia> finddelay([1, 2, 3], [0, 0, 1, 2, 3])
+-2
+```
+"""
+function finddelay(x::AbstractVector{<: Real}, y::AbstractVector{<: Real})
```
CrocoDuckoDucks: @galenlynch the function below is searching for the peak of the absolute correlation.

```julia
s = xcorr(y, x)
max_corr = maximum(abs, s)
max_idxs = Compat.findall(x -> abs(x) == max_corr, s)
```

galenlynch:
Thanks for sticking with it!

PR #130, opened 2016-08-06
Ensure slicedim always returns an array on v0.5, replace sub -> view

ibadr:
On Julia v0.5-rc1+1, tests were failing due to `slicedim` returning a scalar instead of a vector of length 1 in cases of slicing at exactly one location. This is possibly due to the first breaking change mentioned here:
https://github.com/JuliaLang/julia/blob/release-0.5/NEWS.md#breaking-changes

Using the recommended syntax `i:i` fixed the problem and all tests passed. ~~Deprecation warnings are yet to be addressed, though.~~ UPDATE: deprecation warnings are now fixed, mainly by replacing `sub` -> `view` by using Compat v0.8.4. The warnings related to `symbol` -> `Symbol` are also already taken care of on the master of Polynomials.jl.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,4 +1,4 @@
 julia 0.4
 Polynomials
 Reexport
-Compat 0.7.15
+Compat 0.8.4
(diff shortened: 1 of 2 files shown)


simonster:
Thank you!

PR #147, opened 2017-01-31
0.6 update

sglyon:
FloatRange{T} is replaced by StepRangeLen{T,Base.TwicePrecision{T},Base.TwicePrecision{T}}
so I made a typealias back to FloatRange{T}

### DIFF
--- a/src/Filters/coefficients.jl
+++ b/src/Filters/coefficients.jl
@@ -285,7 +285,7 @@ function Base.convert{Z,P}(::Type{SecondOrderSections}, f::ZeroPoleGain{Z,P})
 
     # Allocate memory for biquads
     T = promote_type(realtype(Z), realtype(P))
-    biquads = Array(Biquad{T}, (n >> 1)+(n & 1))
+    biquads = Array{Biquad{T}}((n >> 1)+(n & 1))
 
     # Build second-order sections in reverse
     # First do complete pairs
--- a/src/Filters/design.jl
+++ b/src/Filters/design.jl
@@ -101,7 +101,7 @@ Chebyshev2(n::Integer, ripple::Real) = Chebyshev2(Float64, n, ripple)
 # Compute Landen sequence for evaluation of elliptic functions
 function landen(k::Real)
     niter = 7
-    kn = Array(typeof(k), niter)
+    kn = Array{typeof(k)}(niter)
     # Eq. (50)
     for i = 1:niter
         kn[i] = k = abs2(k/(1+sqrt(1-abs2(k))))
@@ -171,8 +171,8 @@ function Elliptic(T::Type, n::Integer, rp::Real, rs::Real)
     # Eq. (65)
     v0 = -im/n*asne(im/εp, k1)
 
-    z = Array(Complex{T}, 2*div(n, 2))
-    p = Array(Complex{T}, n)
+    z = Array{Complex{T}}(2*div(n, 2))
+    p = Array{Complex{T}}(n)
     gain = one(T)
     for i = 1:div(n, 2)
         # Eq. (43)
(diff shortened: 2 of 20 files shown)


sglyon:
Ok, so the one big issue here is that `@test_approx_eq_eps` is deprecated in 0.6 in favor of `@test ::Expr atol=eps`.

That doesn't work on 0.4 or 0.5. Does anyone have any recommendations for this? 

One approach would be to do `@test abs(::Expr) < eps` everywhere

andreasnoack:
I've been using https://github.com/JuliaLang/Compat.jl/blob/master/src/Compat.jl#L1769 directly instead of the two-argument macro.

sglyon:
Thanks for the tip! It seems to work well

src/Filters/coefficients.jl
```diff
@@ -201,7 +201,7 @@ function groupzp(z, p)
         closest_zero_idx = 0
         closest_zero_val = Inf
         for j = 1:length(z)
-            val = abs(z[j] - p[i])
+            val = @compat abs.(z[j] - p[i])
```
simonster: Shouldn't be necessary since `z[j]` and `p[i]` are scalar.

src/Filters/coefficients.jl
```diff
@@ -262,8 +262,8 @@ function Base.convert{Z,P}(::Type{SecondOrderSections}, f::ZeroPoleGain{Z,P})
     matched || throw(ArgumentError("complex poles could not be matched to their conjugates"))
 
     # Sort poles according to distance to unit circle (nearest first)
-    sort!(complexp, by=x->abs(abs(x) - 1))
-    sort!(realp, by=x->abs(abs(x) - 1))
+    @compat sort!(complexp, by=x->abs.(abs.(x) - 1))
+    @compat sort!(realp, by=x->abs.(abs.(x) - 1))
```
simonster: This also looks scalar

src/Filters/design.jl
```diff
@@ -282,8 +282,8 @@ function transform_prototype(ftype::Highpass, proto::ZeroPoleGain)
         newp[i] = ftype.w / p[i]
     end
 
-    abs(real(num) - 1) < np*eps(real(num)) && (num = 1)
-    abs(real(den) - 1) < np*eps(real(den)) && (den = 1)
+    @compat abs.(real(num) - 1) < np*eps(real(num)) && (num = 1)
+    @compat abs.(real(den) - 1) < np*eps(real(den)) && (den = 1)
```
simonster: This also looks scalar

test/util.jl
```diff
@@ -1,99 +1,98 @@
...
 
 #Sanity check with odd number of samples
-h2 = hilbert([ones(10); zeros(9)]) 
-@test_approx_eq real(h2) [ones(10); zeros(9)]
+h2 = hilbert([ones(10); zeros(9)])
+@test real(h2) ≈ [ones(10); zeros(9)]
 
 #Sanity check with integer arguments
-r = round(Int, rand(128)*20)
+r = @compat round.(Int, rand(128)*20)
```
simonster: Currently failing here on 0.4/0.5. I think these versions don't have the change to `broadcast` that makes types work, and I guess `Compat` is not handling this case? Fortunately, this could just be `rand(0:20, 128)`, which should work on all versions.

sglyon:
@simonster thanks for the feedback. I've removed the `@compat f.(...)` in those cases.

sglyon:
Hey @simonster  so things are passing on 0.5 now, but not 0.4.

There are two issues there

1. inference on 0.4 isn't great with stuff like `@compat sinpi.(x)`. It gives us back a `Vector{Any}` which prohibits us from calling routines like `hilbert` on the result. I've swapped that pattern out for `map(sinpi, x)` and it seems to work fine so **this is no longer an issue**
2. We've got some floating point stuff happening on 0.4. The tests in `filter_design.jl` no longer pass unless we set the tolerance to something like `5eps()`. 

What do you recommend doing for that second point?

simonster:
If it's not easy to make 0.4 work, I would be fine with dropping it. Just update `REQUIRE` and `.travis.yml`.

sglyon:
That'd be the easiest.

I can probably go back and remove all the `@compat f.(...)` stuff then.

simonster:
Sounds good. Thanks again for taking the lead on this!

sglyon:
OK @simonster sorry for the mess.

I've removed 0.4 support, which allowed me to remove a whole bunch of `@compat` stuff. I think we could go further and remove instances of `@compat Union{...}`, but I don't have the time to do that right now. 

I'm not sure why coverage is complaining, but I think this should be good to merge.

PR #156, opened 2017-05-10
Integrate LPC methods `lpc_burg` and `lpc_levinson`

staticfloat:
`lpc_burg` gratefully copied from https://github.com/oyd11/LPC.jl

See https://github.com/oyd11/LPC.jl/issues/1 for official permission to use the code in this repository.


src/lpc.jl
```diff
@@ -0,0 +1,75 @@
...
+NOT return the leading `1` present in the true autocorrelative estimate; it
+omits it as it is implicit in every LPC estimate, and must be manually
+reintroduced if the returned vector should be treated as a polynomial. This
+function implements the mathematics published in [1].
+
+[1] - Enhanced Partial Tracking Using Linear Prediction
+(DAFX 2003 article, Lagrange et al)
+http://www.sylvain-marchand.info/Publications/dafx03.pdf
+"""
+function lpc_burg{T <: Number}(x::Vector{T}, p::Int)
```
gummif: A better API would probably have a parameter of type maybe `LPCBurg` (similar for levinson) and rename the function to `lpc`. Then the default would be `lpc(x, p) = lpc(x,p, LPCBurg())`.

src/lpc.jl
```diff
@@ -0,0 +1,75 @@
...
+    ef = x                      # forward error
+    eb = x                      # backwards error
+    a = [1; zeros(T, p)...]     # prediction coefficients
```
simonster: splatting here is not necessary

src/lpc.jl
```diff
@@ -0,0 +1,75 @@
+"""
```
simonster: In DSP.jl, the convention is to wrap separate functionality in modules and `@reexport`.

src/lpc.jl
simonster: Should probably accept `AbstractVector` and not just `Vector`.

src/lpc.jl
```diff
@@ -0,0 +1,75 @@
...
+    eb = x                      # backwards error
+    a = [1; zeros(T, p)...]     # prediction coefficients
+
+    # Initialize prediction error wtih the variance of the signal
+    prediction_err = (x'*x ./ length(x))[1]
+
+    for m in 1:p
+        efp = ef[1:end-1]
+        ebp = eb[2:end]
+        k = (-2 .* ebp'*efp ./ ( ebp'*ebp + efp'*efp ))[1]
```
simonster: or `k = -2 * dot(ebp, efp) / (sum(abs2, ebp) + sum(abs2, efp))`

src/lpc.jl
```diff
@@ -0,0 +1,75 @@
...
+    ef = x                      # forward error
+    eb = x                      # backwards error
+    a = [1; zeros(T, p)...]     # prediction coefficients
+
+    # Initialize prediction error wtih the variance of the signal
+    prediction_err = (x'*x ./ length(x))[1]
+
+    for m in 1:p
+        efp = ef[1:end-1]
+        ebp = eb[2:end]
```
simonster: Not sure if performance is a concern here, but if I am reading the algorithm right, this could be done in-place by copying `ef` and `eb` into vectors at the start of the function, and then using `shift!` and `pop!` along with in-place scaling below.

src/lpc.jl
```diff
@@ -0,0 +1,75 @@
...
+(DAFX 2003 article, Lagrange et al)
+http://www.sylvain-marchand.info/Publications/dafx03.pdf
+"""
+function lpc_burg{T <: Number}(x::Vector{T}, p::Int)
+    ef = x                      # forward error
+    eb = x                      # backwards error
+    a = [1; zeros(T, p)...]     # prediction coefficients
+
+    # Initialize prediction error wtih the variance of the signal
+    prediction_err = (x'*x ./ length(x))[1]
```
simonster: `sum(abs2, x)/length(x)`

src/lpc.jl
```diff
@@ -0,0 +1,75 @@
...
+    a = zeros(p,p)
+    prediction_err = zeros(1,p)
+
+    # for m = 1
+    a[1,1] = -R_xx[2]/R_xx[1]
+    prediction_err[1] = R_xx[1]*(1-a[1,1]^2)
+
+    # for m = 2,3,4,..p
+    for m = 2:p
+        a[m,m] = (-(R_xx[m+1] + dot(vec(a[m-1,1:m-1]),R_xx[m:-1:2]))/prediction_err[m-1])
```
simonster: `vec` shouldn't be necessary. Also might be faster with views.

src/lpc.jl
staticfloat: So when you say separate functionality, do you mean that you want an `LPCBurg` module and an `LPCLevinson` module each containing a single function, then `@reexport` those functions into the main `DSP` module?

gummif: I think he means wrapping everything inside `module LPC`, see e.g. windows.jl.

simonster: Yes. The two algorithms can live in a single LPC module, but since the functionality is independent of the rest of DSP.jl, it should be a separate module.

staticfloat:
Gotcha.  I've done that, rebased,  and added a simple test so that the algorithms at least show up on coverage.

test/lpc.jl
```diff
@@ -0,0 +1,10 @@
+# Filter some noise, try to learn the coefficients
+coeffs = [1, .5, .3, .2]
+x = filt([1], coeffs, randn(50000))
+
+for method in [LPCBurg(), LPCLevinson()]
+    # Analyze the filtered noise, attempt to reconstruct the coefficients above
+    ar, e = lpc(x[1000:end], length(coeffs)-1)
```
gummif: Looks good. Just forgot to pass method to lpc.

staticfloat:
Oh wow that's embarrassing.  Here's a fixed version.

PR #160, opened 2017-06-13
Move DSP functions from Base into this package

ararslan:


ararslan:
Does this look alright, @simonster? The coverage drop here is unfortunately expected, since test/dsp.jl in Base has fairly poor coverage. We can always add more tests for those functions but in the meantime I'd like to merge and tag so we can get FFTW out of Base.

src/DSP.jl
```diff
@@ -2,14 +2,37 @@ __precompile__()
 
 module DSP
 
+# We want to be very sure we don't pull in Base names unless we're very sure we want them
+# This macro will be called in each submodule herein to do the appropriate imports
+macro importffts()
+    quote
+        using AbstractFFTs, FFTW
+        importall AbstractFFTs, FFTW
+        if VERSION >= v"0.7.0-DEV.585"
```
tkelman: this isn't right, use isdefined

ararslan: I know, it was a guess. I'm going to go back and change the explicit version checks to be exact immediately once the Base PR merges. The problem with `isdefined` is that I was pretty thorough in the deprecation to ensure that everything was still defined.

tkelman: this shouldn't be tagged with the wrong cutoff. do the base versions throw errors?

ararslan: I got the cutoff here by building the Base PR locally after a rebase and checking `VERSION` so it seems like it would at least be a decent approximation. Yes, all of the Base versions after deprecation throw errors.

tkelman: there have been commits to base since though, where this shouldn't apply yet.

could try running one of the base functions then instead of checking isdefined

ararslan: I can just set the version cutoff to something way off, e.g. `v"1.0.0"`, then go back and change it. That at least ensures that the code path doesn't get hit in the interim.

PR #161, opened 2017-06-13
Remove submodule structure

ararslan:
This is making adding FFT package dependencies very annoying


simonster:
Can you explain why this makes adding FFT package dependencies annoying? I'd rather not do this, since the modules are pretty well-separated in terms of what they actually do.

ararslan:
I keep getting "AbstractFFTs not defined," even after importing AbstractFFTs into every submodule. I'll keep at it if this is objectionable.

PR #162, opened 2017-06-15
Temporarily bump the FFTW importing check on VERSION to >=1.0.0

ararslan:
Does this seem reasonable, @tkelman?

### DIFF
--- a/src/DSP.jl
+++ b/src/DSP.jl
@@ -8,7 +8,7 @@ macro importffts()
     quote
         using AbstractFFTs, FFTW
         importall AbstractFFTs, FFTW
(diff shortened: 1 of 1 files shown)


tkelman:
Hm I guess this doesn't result in any less functionality in the package now, or any conflicts, just a quick fix necessary once the pr merges

ararslan:
Yeah, that's the idea. Seem like an okay solution?

ararslan:
Seems a better solution to me and you don't seem utterly opposed to it, so I'm going to go ahead and merge and retag to update the METADATA PR.

PR #164, opened 2017-06-17
Fix Base function overloading

ararslan:
This gets the MP3 and SampledSignals tests passing for me on master. cc @tkelman

### DIFF
--- a/src/DSP.jl
+++ b/src/DSP.jl
@@ -12,9 +12,6 @@ macro importffts()
             import AbstractFFTs: fftshift, ifftshift
             import FFTW: plan_fft, plan_fft!, plan_rfft, plan_brfft, plan_irfft, plan_bfft, plan_bfft!,
                          fft, fft!, ifft, ifft!, irfft, bfft, bfft!
-            if isdefined(Base, :DSP)
-                import Base: conv, conv2, deconv, filt, filt!, xcorr
-            end
         else
             import Base: plan_fft, plan_fft!, plan_rfft, plan_brfft, plan_irfft, plan_bfft, plan_bfft!,
                          fft, fft!, ifft, ifft!, irfft, bfft, bfft!, fftshift, ifftshift
@@ -24,7 +21,15 @@ end
 
 @importffts
 
-include("dspbase.jl")
+if VERSION >= v"0.7.0-DEV.602"
+    include("dspbase.jl")
+    if isdefined(Base, :DSP)
+        import Base: conv, conv2, deconv, filt, filt!, xcorr
+    else
+        export conv, conv2, deconv, filt, filt!, xcorr
+    end
+end
+
 include("util.jl")
 include("windows.jl")
 include("periodograms.jl")
--- a/src/Filters/Filters.jl
+++ b/src/Filters/Filters.jl
@@ -1,8 +1,13 @@
 module Filters
-import ..DSP: @importffts, filt, filt!
+import ..DSP: @importffts
 using Polynomials, ..Util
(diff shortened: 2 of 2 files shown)


ararslan:
Hmmmmmmmm uh oh. Looks like something might be wrong with FFTW on nightly now that the binary no longer has FFTW.

ararslan:
Ah, the tests are good on master but not on 0.6. Okay, this isn't the answer then.

tkelman:
Bump? Should we call in @ssfrr or anyone else for some help here? Add upper bounds for the time being?

ararslan:
Haven't been around today. Seems that BinDeps isn't installing FFTW properly so it isn't working on nightly. Trying to figure it out now.

src/Filters/Filters.jl
```diff
@@ -1,8 +1,9 @@
 module Filters
-import ..DSP: @importffts, filt, filt!
+import ..DSP: @importffts
 using Polynomials, ..Util
 import Base: *
 @importffts
+VERSION >= v"0.7.0-DEV.602" && import ..DSP: filt, filt!
```
ssfrr: On 0.6 we need to import these functions from `Base`. Changing this line to 

```julia
if VERSION >= v"0.7.0-DEV.602"
    import ..DSP: filt, filt!
else
    import Base: filt, filt!
end
```

seems to work, but I confess I don't have a full grasp of your FFTW/DSP migration master plan.

ararslan: Thanks, this is a case I forgot to update!

src/DSP.jl
```diff
@@ -24,6 +21,14 @@ end
...
 
+if VERSION >= v"0.7.0-DEV.602"
+    if isdefined(Base, :DSP)
+        import Base: conv, conv2, deconv, filt, filt!, xcorr
+    else
+        export conv, conv2, deconv, filt, filt!, xcorr
+    end
+end
+
 include("dspbase.jl")
```
ssfrr: The latest commit fixes things on SampledSignals and MP3, but it's failing its own tests. Looks like `dspbase.jl` also defines `filt` and `filt!` methods. Should that file even be included for Julia 0.6? AFAICT moving the `include("dspbase.jl")` to inside the `if VERSION >= v"0.7.0-DEV.602"` block above fixes things on 0.6 and 0.7 (nightly)

ssfrr: @ararslan Do you think this fix will get things working? Or at least not cause extra problems on 0.7 while getting things working on 0.6?

ararslan: Sorry for the delay, been pretty absent on GitHub due to JuliaCon. I'll try that fix now.

ssfrr: Totally understand. I'm guessing/hoping that there will be folks from my poster session that might want to try out the audio stuff now that our conference is wrapping up.

ararslan:
Nightly failing due to FFTW garbage. Can we just merge the fix we have for that figure out Windows later, @tkelman?

tkelman:
Nightly doesn't really matter yet. Fix 0.6.

ararslan:
0.6 is passing tests here

ssfrr:
did you fix the `dspbase.jl` issue?

ararslan:
Yeah I pushed that and squashed the commits

ssfrr:
Sounds good. Passes the latest tagged versions of SampledSignals and MP3 as well.

ararslan:
Awesome, thanks. I'll merge and tag then.

Issue #143, opened 2016-12-23
Numerical errors using freqz with high-order filters?


ma-laforge:
I used the following code to plot the response of a Chebyshev1 vs filter order (20, 25, 27, 30):
```
using DSP
using InspectDR
using Colors
InspectDR.defaults.rendersvg = false

npts = 500
fz = collect(linspace(0, pi, npts))
fmax = 500; fl = 150 #Hz
passrip = 40 #dB
filttype = Lowpass(fl/fmax)

plotobj = InspectDR.Plot2D()
offset = 0
for order in [20, 25, 27, 30]
    filtimpl = Chebyshev1(order, passrip)
    _filt = digitalfilter(filttype, filtimpl)
    tf_graph = freqz(_filt, fz)
    wfrm = add(plotobj, fz, 10*log10(abs2(tf_graph))+offset)
    wfrm.line = line(color=RGB(0,0,1), width=3, style=:solid)
    offset-=30
end
plotobj.ext = InspectDR.PExtents2D(NaN, NaN, -200, 10) #Focus on interesting bit
display(plotobj)
```
![dsp_freqz_bug_vsorder](https://cloud.githubusercontent.com/assets/11205411/21447111/a3ed93b4-c89d-11e6-8905-957ff926148a.png)

NOTE: The highest waveform is of order 20, with order 25 plotted 30dB lower, and order 27 plotted 30dB lower, and order 30 plotted 30dB lower - just for visibiltiy reasons.

As you can see, the graph generated for the 27th order filter is visibly starting to break down near the cutoff frequency - and the graph for the 30th order filter is dominated by noise under the cutoff frequency.

ma-laforge:
At first, I thought `Chebyshev1()` did not support high order filters, but my experiment using the notebook described in https://github.com/JuliaDSP/DSP.jl/issues/141 indicates otherwise.

By selecting the following parameters:
```
Lowpass / "ChebyshevPass" (Chebyshev1)
Filter order: 30
Pass-band ripple: 5dB
Low-frequency: 150 Hz
```

the notebook overlays the spectrum of filtered random data with the graph of the transfer function (generated with `freqz()`):
![dsp_freqz_bug_vsspectrum](https://cloud.githubusercontent.com/assets/11205411/21447157/7f31fd3e-c89e-11e6-94d2-4f630deb5581.png)

It would appear that the filtered data spectrum behaves as expected, whereas the graph generated with `freqz()` is visibly corrupt.

[...]

PR #169, opened 2017-07-26
Specialized freqz/freqs methods

martinholters:
Just bumped into #143 when I was trying to numerically verify some custom filter design methods and thought I might give fixing it a try. While at it, I did the same for `freqs`, although I didn't construct test cases that previously would suffer badly from the rounding.

I've left
```julia
function freq[sz](filter::FilterCoefficients, w::Number)
    filter = convert(PolynomialRatio, filter)
    # ...
end
```
in place as a fallback for potential additional `FilterCoefficients` subtypes.


martinholters:
Travis failure on nightly looks unrelated.

martinholters:
Bump.

PR #171, opened 2017-08-01
 LPC burg: ar_burg() interface - returns reflection coeffs - same as …

oyd11:
…matlab call as well, lpc() - call - returns 2 params - same as Levinson version

### DIFF
--- a/src/lpc.jl
+++ b/src/lpc.jl
@@ -1,12 +1,13 @@
 module LPC
 export lpc, LPCBurg, LPCLevinson
+export ar_burg
 
 # Dispatch types for lpc()
 mutable struct LPCBurg; end
 mutable struct LPCLevinson; end
 
 """
-`lpc_burg(x::AbstractVector, p::Int)`
+`ar_burg(x::AbstractVector, p::Int)`
 
 LPC (Linear-Predictive-Code) estimation, using the Burg method.
 
@@ -21,13 +22,13 @@ function implements the mathematics published in [1].
 (DAFX 2003 article, Lagrange et al)
 http://www.sylvain-marchand.info/Publications/dafx03.pdf
 """
-function lpc{T <: Number}(x::AbstractVector{T}, p::Int, ::LPCBurg)
+function ar_burg{T <: Number}(x::AbstractVector{T}, p::Int)
     ef = x                      # forward error
     eb = x                      # backwards error
     a = [1; zeros(T, p)]        # prediction coefficients
-
     # Initialize prediction error wtih the variance of the signal
-    prediction_err = (sum(abs2, x) ./ length(x))[1]
+    prediction_err :: T = sum(abs2, x) ./ length(x)
+    reflection_coeffs = zeros(T, p)
 
     for m in 1:p
         efp = ef[1:end-1]
@@ -37,8 +38,14 @@ function lpc{T <: Number}(x::AbstractVector{T}, p::Int, ::LPCBurg)
         eb = ebp + k .* efp
         a[1:m+1] = [a[1:m]; 0] + k .* [0; a[m:-1:1]]
         prediction_err *= (1 - k*k)
+        reflection_coeffs[m] = k
     end
 
+    return a, prediction_err, reflection_coeffs
+end
+
+function lpc{T <: Number}(x::AbstractVector{T}, p::Int, ::LPCBurg)
+    a, prediction_err = ar_burg(x, p)
     return a[2:end], prediction_err
 end
 


oyd11:
Hi, I wrote the function @staticfloat integrated here before (see: https://github.com/JuliaDSP/DSP.jl/pull/156 ) - this is great, but to be able to use this version in my own code - I also need the Reflection-Coefficients as a return value (they can be re-evaluated - but there's a numerical stability cost) 
I've wrapped the lpc() call in a unified interface - just like Static did, but provided a 'ar_burg()' function call with the specific Burg algorithm which includes the Reflection-Coefficients as a side effect. This is the same as the Matlab "arburg()" call [ https://www.mathworks.com/help/signal/ref/arburg.html ], feel free to remove the extra dash (now that I see that Matlab doesn't have that.

oyd11:
note: the non passing CI test seems to be due to some non-related incompatibility issue with Julia nightly (0.7 line)  - something about some functions to have moved out of base into DSP

src/lpc.jl
```diff
@@ -1,12 +1,13 @@
...
 export lpc, LPCBurg, LPCLevinson
+export arburg
 
 # Dispatch types for lpc()
 mutable struct LPCBurg; end
 mutable struct LPCLevinson; end
 
 """
-`lpc_burg(x::AbstractVector, p::Int)`
+`ar_burg(x::AbstractVector, p::Int)`
```
staticfloat: this should be `arburg()` not `ar_burg()`

src/lpc.jl
```diff
@@ -59,7 +66,7 @@ function implements the mathematics described in [1].
 """
 function lpc{T <: Number}(x::AbstractVector{T}, p::Int, ::LPCLevinson)
     R_xx = xcorr(x,x)[length(x):end]
-    a = zeros(p,p)
+    a = zeros(T,p,p)
```
staticfloat: Good catch

oyd11: I've tested the versions for type stability - the Burg versions works with Float(16/32/64)/BigFloat - where as the Levinson version works only with Float(32/64) - because of conv() doing an FFT (which is alright for now I guess).
Anyhow - the Levinson version can also have the reflection coeffs returns eventually.
Test is cool, I'll try to add a Ladder(Lattice) filter - where the reflection coeffs are used, and that can also be added to the test

src/lpc.jl
```diff
@@ -21,13 +22,13 @@ function implements the mathematics published in [1].
...
 """
-function lpc{T <: Number}(x::AbstractVector{T}, p::Int, ::LPCBurg)
+function arburg{T <: Number}(x::AbstractVector{T}, p::Int)
     ef = x                      # forward error
     eb = x                      # backwards error
     a = [1; zeros(T, p)]        # prediction coefficients
-
     # Initialize prediction error wtih the variance of the signal
-    prediction_err = (sum(abs2, x) ./ length(x))[1]
+    prediction_err :: T = sum(abs2, x) ./ length(x)
```
martinholters: `/` instead of `./`? Or is there some reason for the broadcasting I'm not aware of?

oyd11: no, change it to / if you see fit. 
This was not changed in the current commit, I probably wrote it as to be explicit about doing a scalar operation (rather than a broadcast)

src/lpc.jl
galenlynch: Is this accurate for integer `T`?

galenlynch: Ah I wasn't even thinking about fixed point arithmetic. That would be cool.

My point with these comments is that the current PR will error if `x` is an abstract vector with `Integer` elements, because of the type assertions and container types used in the code. I think you should either change the type assertions and container types to reflect the actual types, or restrict the type bounds in the method definitions to types that will work as you expect (e.g. `AbstractFloat`).

Because restricting the type bounds will change the API for this function, I think it would be best to be more careful about the container types. If `T = eltype(x)` and `T <: AbstractFloat` or `T <: FixedPoint` then `prediciton_error`, `reflection_coeffs`, and `a` will be type `T` (or have eltype `T`). If `T <: Integer` then `prediction_error` and `reflection_coeffs` will be type `Float64`, and the Levinson algorithm will error because `xcorr` errors with integer inputs.

oyd11: Fixed-Point is quite common for LPC implementations, it's an import use-case as LPC is used in some analysis and compression algorithms in some embedded systems with Integer-Only hardware (still I believe); the few integer LPC's I've seen had the Fixed-Point arithmetics inlined by hand (for example in Speex), but this could actually be an interesting use case to see whether we can produce efficient machine code with an abstract type here (might change the abstraction for that). Anyhow, I think I've even tested this with Rational before, it's quite cool that it works.

PR #177, opened 2017-09-27
Fix a bunch of deprecations on Julia 0.7

martinholters:
* `Range` -> `AbstractRange` (requires Compat 0.32 on Julia 0.6)
* use `where` syntax for parametric functions
* `2.^` -> `2 .^`
* `-` -> `.-` (where necessary)

The tests show some `Loop variable "x" overwrites a variable in an enclosing scope. In the future the variable will be local to the loop instead.` warnings, but they're all harmless, the last value of `x` isn't actually used for anything.

### DIFF
--- a/src/periodograms.jl
+++ b/src/periodograms.jl
@@ -5,6 +5,7 @@
 module Periodograms
 using ..DSP: @importffts
 using ..Util, ..Windows
+using Compat: AbstractRange
 export arraysplit, nextfastfft, periodogram, welch_pgram, mt_pgram,
        spectrogram, power, freq, stft
 @importffts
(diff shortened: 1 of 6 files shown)


src/periodograms.jl
```diff
@@ -357,7 +358,7 @@ function spectrogram(s::AbstractVector{T}, n::Int=length(s)>>3, noverlap::Int=n>
 
     out = stft(s, n, noverlap, PSDOnly(); onesided=onesided, nfft=nfft, fs=fs, window=window)
     Spectrogram(out, onesided ? rfftfreq(nfft, fs) : fftfreq(nfft, fs),
-                ((0:size(out,2)-1)*(n-noverlap)+n/2)/fs)
+                (n/2 : n-noverlap : (size(out,2)-1)*(n-noverlap)+n/2) / fs)
```
martinholters: `range + scalar` is deprecated in 0.7, `range .+ scalar` produces a `Vector` in 0.6, so this rewrite seemed most straight-forward for cross-version support.

ararslan:
Looks great, thanks!

PR #178, opened 2017-09-27
Add function to calculate the mean frequency of a signal

halleysfifthinc:
The mean frequency of a signal is a useful analysis and is also present in MATLAB.

I'm open to suggestions to improve the code or re-work the pull request to make it more convenient to accept.

### DIFF
--- a/doc/util.rst
+++ b/doc/util.rst
@@ -72,11 +72,11 @@
     Return the root mean square of signal ``s`` given the FFT transform
     ``f = fft(s)``. Equivalent to ``rms(ifft(f))``.
 
-.. function:: meanfreq(f, fs)
+.. function:: meanfreq(x, fs)
 
-    Calculate the mean power frequency of `f` with a sampling frequency of `fs`, defined as:
+    Calculate the mean power frequency of `x` with a sampling frequency of `fs`, defined as:
 
-    .. math:: \\textup{MPF} = \\frac{\\sum_{i=1}^{F} f_i X_i^2 }{\\sum_{i=0}^{F} X_i^2 } \\textup{Hz}
+    .. math:: MPF = \frac{\sum_{i=1}^{F} f_i X_i^2 }{\sum_{i=0}^{F} X_i^2 } Hz
 
     where `F` is the Nyquist frequency, and `X` is the power spectral density.
 
--- a/doc/util.rst
+++ b/doc/util.rst
@@ -72,9 +72,9 @@
     Return the root mean square of signal ``s`` given the FFT transform
     ``f = fft(s)``. Equivalent to ``rms(ifft(f))``.
 
-.. function:: meanfreq(f)
+.. function:: meanfreq(f, fs)
 
-    Calculate the mean power frequency of `x`, defined as:
+    Calculate the mean power frequency of `f` with a sampling frequency of `fs`, defined as:
 
     .. math:: \\textup{MPF} = \\frac{\\sum_{i=1}^{F} f_i X_i^2 }{\\sum_{i=0}^{F} X_i^2 } \\textup{Hz}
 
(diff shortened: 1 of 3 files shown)


src/util.jl
```diff
@@ -172,7 +173,16 @@ rms(s::AbstractArray{T}) where {T<:Number} = sqrt(sum(abs2, s)/length(s))
 # root mean square of fft of signal
 rmsfft(f::AbstractArray{T}) where {T<:Complex} = sqrt(sum(abs2, f))/length(f)
 
+function meanfreq{T}(f::AbstractVector{T},ΔT)
```
martinholters: Should use `where T` syntax (or not introduce `T` a all). Also probably `where T<:Real` would be in order, no?
Also, if we want to match Matlab, the second argument should be the sampling rate rather than the sampling interval. And a default of 2π would be nice to get normalized frequency by default.

doc/util.rst
```diff
@@ -72,3 +72,11 @@
     Return the root mean square of signal ``s`` given the FFT transform
     ``f = fft(s)``. Equivalent to ``rms(ifft(f))``.
 
+.. function:: meanfreq(f)
```
martinholters: Function argument should be `x` here to match the description. Second function argument is missing.

src/util.jl
```diff
@@ -172,7 +173,16 @@ rms(s::AbstractArray{T}) where {T<:Number} = sqrt(sum(abs2, s)/length(s))
...
+    pxxX = abs2.(rfft(f))
```
martinholters: Should this apply a window function to `f` first?

halleysfifthinc:
I changed `fs` to default to 2π (that was what you meant, correct?). As far as applying a window function, no; the MATLAB meanfreq effectively does not use a window function: it applies a Kaiser window with an α of 0, which should leave the input signal untouched.

src/util.jl
```diff
@@ -172,7 +173,16 @@ rms(s::AbstractArray{T}) where {T<:Number} = sqrt(sum(abs2, s)/length(s))
 # root mean square of fft of signal
 rmsfft(f::AbstractArray{T}) where {T<:Complex} = sqrt(sum(abs2, f))/length(f)
 
+function meanfreq{T<:Real}(f::AbstractVector{T},fs=2*π)
```
martinholters: Still uses the old form for parametric functions which is deprecated in Julia 0.7. As `T` isn't used in the function, this can just be
```julia
function meanfreq(f::AbstractVector{<:Real}, fs=2*π)
```
(While at it, the space before the `,` would be nice, too.)

test/util.jl
```diff
@@ -95,6 +95,8 @@ for n = 1:7
     @test fftshift(fftfreq(n)) ≈ fftshift([fftfreq(n);])
 end
 
+@test isapprox(meanfreq(sin.(2*pi*10*(0:1e-3:10*pi)),1e3),10.0,rtol=0.1)
```
martinholters: Some spaces for readability would be appreciated. This could also be
```julia
@test meanfreq(sin.(2*pi*10*(0:1e-3:10*pi)), 1e3) ≈ 10.0 rtol=0.1
```
But that might be a matter of taste.
Also `rtol` could be lowered to `1e-3`.

martinholters:
Regarding the added documentation: Did you try to build it? I noticed that in other places no double backslashes are used in the LaTeX code inside `math`, but I'm rather clueless about RST/Sphinx.

Also related to the docs: Am I right that this is _not_ the spectral centroid? Otherwise, we should mention that term somewhere to make the function discoverable. But IIUC, spectral centroid would use `abs.(rfft(f))` (instead of `abs2`), is that correct?

halleysfifthinc:
That is correct, this is not the spectral centroid. The LaTeX equations should render correctly now.

martinholters:
I don't think I've ever used `meanfreq` in MATLAB or missed that functionality in Julia, so if anyone else has comments, that would be appreciated. Otherwise, I'm going to merge this tomorrow.

martinholters:
Thanks for the contribution, @halleysfifthinc !

PR #180, opened 2017-09-28
Apply complex conjugatation to second argument of xcorr

### LINKED ISSUE #173: xcorr working with complex vector
It seems there is a problem in the the xcorr when dealing with complex vector .

I am expecting to see a peak when I do the following:

val = xcorr(x,x)
figure(1)
plot(abs.(val))

however it is not as expected. I have to manually conjugate the second vector then the results is correct.
val = xcorr(x,conj(x))
figure(1)
plot(abs.(val))

conclusion would be in the current xcorr function, there is no conjugation implemented for the 2nd input vector.

martinholters:
Fixes #173.

This is still WIP because I'd like to include
```julia
@test size(@inferred(xcorr(x, su))) == (19,)
```
after JuliaLang/julia#23900 is merged (or the issue is otherwise resolved).

### DIFF
--- a/src/dspbase.jl
+++ b/src/dspbase.jl
@@ -197,5 +197,5 @@ function xcorr(u, v)
     elseif sv < su
         v = [v;zeros(eltype(v),su-sv)]
     end
-    flipdim(conv(flipdim(u, 1), v), 1)
+    conv(u, flipdim(conj(v), 1))
 end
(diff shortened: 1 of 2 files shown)


martinholters:
Ok, I've added the extra test. Unfortunately, it's not run on CI yet due to the necessary condition on Julia `VERSION`, but I've confirmed it works locally. So good to go from my side. Any comments before I merge?

martinholters:
Ah, I thought being elevated to "Member" allowed me to merge. Apparently, that is not the case.

ararslan:
Looks like permissions are set on a team level rather than at an org level, so I've added you to the appropriate team. You should have write access now.

martinholters:
Thanks!

PR #187, opened 2017-12-12
Fix deprecations on Julia 0.7

martinholters:
This should fix all current deprecations except for a bunch of implicit assignments to global variables during the tests. These are probably best handled by introducing testsets to completely avoid globals, but that's for another PR...

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,5 +1,5 @@
 julia 0.6-rc1
-Compat 0.32
+Compat 0.38
 Polynomials 0.1.0
 Reexport
 SpecialFunctions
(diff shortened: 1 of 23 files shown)


martinholters:
Test on nightly fails due to JuliaMath/SpecialFunctions.jl#64 and JuliaMath/FFTW.jl#48 which are both cause by a BinDeps issue being worked on at https://github.com/JuliaLang/BinDeps.jl/pull/344. Should wait until that is resolved and rerun CI here.

martinholters:
Ok, BinDeps is fixed, waiting for the version tag to be merged into METADATA: https://github.com/JuliaLang/METADATA.jl/pull/12481

Issue #188, opened 2017-12-21
deconv(b,a) modifies b 


hessammehr:
On Julia 0.6.1 (`Base.deconv`) and nightly (`DSP.conv`):

```julia
a = [1.0, 2.0, 3.0]
c = deconv(a, [1000.0, 0.0])
a
```
gives
```
3-element Array{Float64,1}:
 0.001
 0.002
 0.003
```
My understanding is this function is not supposed to modify its arguments.

hessammehr:
The culprit seems to be `dspbase.jl:53`
```julia
if a[1] != 1
    norml = a[1]
    a ./= norml
    b ./= norml
end
```

martinholters:
Yes, that `./=` should probably just be a `/=`. Do you want to open a PR with that change?

PR #189, opened 2018-01-16
Fix for #188

hessammehr:
Test case included.


test/dsp.jl
```diff
@@ -58,3 +58,13 @@ a = [1., 2., 1., 2.]
 b = [1., 2., 3.]
 @test conv(a, b) ≈ [1., 4., 8., 10., 7., 6.]
 @test conv(complex.(a, ones(4)), complex(b)) ≈ complex.([1., 4., 8., 10., 7., 6.], [1., 3., 6., 6., 5., 3.])
+
+# Test for issue #188: deconv mutates inputs
+let b = [4.0, 2.0, 1.0]; a = [2.0, 1.0]
```
martinholters: Can you guard the test with `if VERSION >= v"0.7.0-DEV.602"`? Prior to that, the functions from `Base` are used which contain this bug, too, so the test fails.

martinholters:
I'm actually working on making the package usable on Julia 0.7 again, and I'd prefer to wait with merging this until it's testable.

martinholters:
Um, I was hoping resolving conflicts would re-trigger CI. Apparently it doesn't, so I'll close and re-open.

martinholters:
Thanks for the patience here!

hessammehr:
No worries, glad to see this through. Thanks for all the good work.

P.S. I can try to lend a hand with future PRs if given commit rights.

PR #191, opened 2018-01-22
Make collect(::ArraySplit) do the expected thing; cosmetic changes

davidavdav:
Tiny PR, makes collect(::ArraySplit) produce the correct collection of arrays, plus some small cosmetic changes. 

I intend to add options in the future to ArraySplit to deal with the last lost samples, and to do this symmetrically.  I don't know how you guys think about that.  It would be better for MFCC.jl, which depends on DSP for `spectogram`, in the end.


galenlynch:
Thanks!

PR #190, opened 2018-01-18
Grabbag of fixes for Julia 0.7 and switch to testsets

martinholters:
This is WIP because it cannot actually be tested on Julia 0.7 at the moment because FFTW.jl is broken there. Once that is resolved, I'll try to fix any additional problems that may have been caused in the meantime by further breaking changes or deprecations in Julia.

PR #193, opened 2018-03-19
Add esprit function to util

sirtom67:
esprit finds the frequencies of sinusoids. It can also be used for direction finding in array processing and time-delay estimation in the frequency domain.

It should maybe go in a different package, although it didn't seem to fit any of the existing packages.


sirtom67:
1) I'll add a test for esprit ASAP.
2) Not sure what to do about the Travis build failure. Any ideas?
3) Shall we add a new file for "Estimation"? Or perhaps "Parametric" or "Modeling" (although LPC could go in there) to complement the "Periodograms" file? And ideas?
PS. I'm really enjoying Julia. I worked on the Signal Processing Toolbox at MathWorks way back when and I've been writing python for 10 years. I find Julia an interesting mashup of python and matlab with some of the best features of both.

ssfrr:
Agreed there should be some tests before this is merged.

I don't think `util.jl` is the right place for this, I think either a new `estimation.jl` or even a dedicated `esprit.jl` would be reasonable. There are others with more experience in this codebase who might have different/better-informed opinions though.

You can ignore the failure on 0.7 - DSP.jl isn't generally working on 0.7 yet, but hopefully will once #190 is merged.

src/util.jl
```diff
@@ -370,5 +371,34 @@ function shiftin!(a::AbstractVector{T}, b::AbstractVector{T}) where T
...
+- `Fs::Float64`: sampling frequency, in Hz.
+    
+# Returns
+length p real array of frequencies in units of Hz.
+"""
+function esprit(x::Array, M::Int, p::Int, Fs::Float64=1.0)
+    N = length(x)
+    X = x[ (1:M) .+ (0:N-M)' ];  # in julia, likely faster to loop than to allocate huge index array
+    U,s,V = svd(X);
+    D,_ = eig( U[1:end-1,1:p] \ U[2:end,1:p] );
```
ssfrr: these lines don't need semicolons

src/util.jl
```diff
@@ -370,5 +371,34 @@ function shiftin!(a::AbstractVector{T}, b::AbstractVector{T}) where T
...
+      For faster execution (due to smaller SVD), use small M or small N-M
+- `p::Int`: number of sinusoids to estimate.
+- `Fs::Float64`: sampling frequency, in Hz.
+    
+# Returns
+length p real array of frequencies in units of Hz.
+"""
+function esprit(x::Array, M::Int, p::Int, Fs::Float64=1.0)
+    N = length(x)
+    X = x[ (1:M) .+ (0:N-M)' ];  # in julia, likely faster to loop than to allocate huge index array
```
ssfrr: yeah, it's worth benchmarking (performance might be dominated by the operations below so it might not matter much). You can allocate `X` using `similar` to get the same element types as `x` and then iterate through, using `@views` so you don't need to allocate each of the column vectors as you're copying.

src/util.jl
```diff
@@ -370,5 +371,34 @@ function shiftin!(a::AbstractVector{T}, b::AbstractVector{T}) where T
...
+    U,s,V = svd(X);
+    D,_ = eig( U[1:end-1,1:p] \ U[2:end,1:p] );
+    angle.(D)*Fs/(2*π)
```
ssfrr: If you want to be fancy this could be `angle.(D)*Fs/2π`, because `2π` multiplication is higher precedence than `2*π`

ssfrr:
looking at it again a separate file seems overkill for a pretty short function.

sirtom67:
Thanks for the comments Spencer, I'll take a look and respond when I can (gotta work my day job :-)

sirtom67:
Checks pass! No conflicts! Yay! Ready to merge?

sirtom67:
@ssfrr @martinholters any issues pulling this one in? My first contrib - working on porting remez over from scipy now, has anyone done this already?

martinholters:
There are tests in `util.jl` and `estimation.jl`. One of them should go. Considering that we might add more estimation stuff in the future, keeping `estimation.jl` would be fine with me. But the test should be wrapped in a `@testset`.

Also the docs should be updated to include at least the docstring.

Concerning the algorithm/implementation itself, I can't really judge, but it seems to do the right thing.

sirtom67:
I don't think what I committed is the cause of the failures.

martinholters:
> I don't think what I committed is the cause of the failures.

Bo, they certainly look unrelated.

> How do we get the checks to pass?

I'm looking into it...

martinholters:
Checks fail due to an inference problem in FFTW.jl, in turn caused by https://github.com/JuliaLang/julia/issues/26827.

If there is any urgency, I think we could merge this as is. Otherwise, I'd prefer waiting a few days to see whether we get the tests fixed.

martinholters:
I've rerun CI now that FFTW is fixed and ... green. So I guess this is ready to go, or do you have any further remarks, @ssfrr?

sirtom67:
@martinholters: thank you for rerunning the build. I'm happy to see green. :+1:

sirtom67:
So - can we get a merge and close this request??? Thanks!
 - Tom

martinholters:
Thanks for the contribution, @sirtom67!

Issue #196, opened 2018-04-08
Adding Unwrap.jl to JuliaDSP


platawiec:
I've written arbitrary-dimensional phase unwrapping in the [Unwrap.jl](https://github.com/platawiec/Unwrap.jl) package. See the README for more info.

@bjarthur had asked me to combine the code into the DSP org, either as part of DSP.jl or a separate package in the org. The code is at the point where I'm comfortable doing either - I'd like to get the community's perspective on the best way forward. Once it's settled I can make a PR.

ssfrr:
Looks cool. I'd support merging this into DSP.jl rather than a separate package.

The current `unwrap` function in DSP.jl allows you to specify the wrap overflow amount. I generally use it for phase unwrapping at 2pi so I don't need it to be configurable, but I assume some folks must want that. How hard would it be to make it configurable in your implementations?

platawiec:
Thanks for the feedback - that shouldn't be hard to implement. I'll submit a PR to DSP.jl when I'm done.

PR #198, opened 2018-04-15
add n-dimensional unwrap

platawiec:
Ref #196 

I've added the n-dimensional `unwrap` from `Unwrap.jl` [here](https://github.com/platawiec/Unwrap.jl). The doc-string for `unwrap` goes over how to use it. The new functionality allows the user to take arbitrary arrays that are "wrapped" (or modulo some arbitrary number) and recover the original array. This is primarily useful when looking at the recovered phases for 2D/3D scenes. Also included is the ability to specify which dimensions are periodic.

There are two other changes to the old `unwrap`:
* There is no longer a `dim` keyword, a user who wants to recover that functionality will have to pass the appropriate `view`s.
* The old `unwrap` would perform the modulo function on the input array. In other words, an array like `[0.0, 6pi, 0.0]` evaluated to `[0.0, 0.0, 0.0]`. If the user is worried that the array they are passing into `unwrap` has values which lie outside the range, then they should do `.%` before calling `unwrap`. Typically, the array passed in is already module some number, because that is why `unwrap` is being called in the first place.

I have not tested against v0.7. Let me know if there are any other tweaks to do.

### DIFF
--- a/src/unwrap.jl
+++ b/src/unwrap.jl
@@ -68,16 +68,16 @@ measurement of a scene, such as when retrieving the phase information of
 of an image, as each pixel is wrapped to stay within (-pi, pi].
 
 # Arguments
-- `m::AbstractArray{T, N}`: Array to unwrap
+- `m::AbstractArray{T, N}`: Array to unwrap.
 - `dims=nothing`: Dimensions along which to unwrap. If `dims` is an integer, then
-    `unwrap` is called that dimension. If `dims=1:ndims(m)`, then `m` is unwrapped
+    `unwrap` is called on that dimension. If `dims=1:ndims(m)`, then `m` is unwrapped
     across all dimensions.
-- `range=2pi`: Range of wrapped array
+- `range=2pi`: Range of wrapped array.
 - `circular_dims=(false, ...)`:  When an element of this tuple is `true`, the
     unwrapping process will consider the edges along the corresponding axis
-    of the array to be connected
+    of the array to be connected.
 - `rng=GLOBAL_RNG`: Unwrapping of arrays with dimension > 1 uses a random
-    initialization. A user can be pass their own RNG through this argument.
+    initialization. A user can pass their own RNG through this argument.
 """
 function unwrap(m::AbstractArray{T,N}; dims=nothing, kwargs...) where {T, N}
     if dims === nothing && N != 1
--- a/src/unwrap.jl
+++ b/src/unwrap.jl
@@ -10,12 +10,12 @@ export unwrap, unwrap!
 
 In-place version of [`unwrap`](@ref).
 """
-function unwrap!(m::AbstractArray{T,N}; dims=nothing, range=2T(pi), kwargs...) where {T, N}
+function unwrap!(m::AbstractArray{T,N}; dims=nothing, kwargs...) where {T, N}
     if dims === nothing && N != 1
         Base.depwarn("`unwrap!(m::AbstractArray)` is deprecated, use `unwrap!(m, dims=ndims(m))` instead", :unwrap!)
         dims = N
     end
-    unwrap!(m, m; dims=dims, range=range, kwargs...)
+    unwrap!(m, m; dims=dims, kwargs...)
 end
 
 """
@@ -23,7 +23,7 @@ end
 
 Unwrap `m` storing the result in `y`, see [`unwrap`](@ref).
 """
-function unwrap!(y::AbstractArray{T,N}, m::AbstractArray{T,N}; dims=nothing, range::Number=2T(pi), kwargs...) where {T, N}
+function unwrap!(y::AbstractArray{T,N}, m::AbstractArray{T,N}; dims=nothing, range=2T(pi), kwargs...) where {T, N}
     if dims === nothing
         if N != 1
             throw(ArgumentError("`unwrap!`: required keyword parameter dims missing"))
@@ -40,7 +40,7 @@ function unwrap!(y::AbstractArray{T,N}, m::AbstractArray{T,N}; dims=nothing, ran
     return y
 end
 
-unwrap_kernel(range=2π) = (x, y) -> y - round((y - x) / range) * range
+unwrap_kernel(range) = (x, y) -> y - round((y - x) / range) * range
 
 """
     unwrap(m; dims=nothing, range=2pi)
@@ -64,6 +64,9 @@ of an image, as each pixel is wrapped to stay within (-pi, pi].
 
 # Arguments
 - `m::AbstractArray{T, N}`: Array to unwrap
+- `dims=nothing`: Dimensions along which to unwrap. If `dims` is an integer, then
+    `unwrap` is called that dimension. If `dims=1:ndims(m)`, then `m` is unwrapped
+    across all dimensions.
 - `range=2pi`: Range of wrapped array
 - `circular_dims=(false, ...)`:  When an element of this tuple is `true`, the
     unwrapping process will consider the edges along the corresponding axis
--- a/src/unwrap.jl
+++ b/src/unwrap.jl
@@ -11,7 +11,7 @@ import Compat.Random: GLOBAL_RNG
 export unwrap, unwrap!
 
 """
-    unwrap!(m; dims=nothing, range=2pi)
+    unwrap!(m; kwargs...)
 
 In-place version of [`unwrap`](@ref).
 """
@@ -24,7 +24,7 @@ function unwrap!(m::AbstractArray{T,N}; dims=nothing, kwargs...) where {T, N}
 end
 
 """
-    unwrap!(y, m; dims=nothing, range=2pi)
+    unwrap!(y, m; kwargs...)
 
 Unwrap `m` storing the result in `y`, see [`unwrap`](@ref).
 """
@@ -48,7 +48,7 @@ end
 unwrap_kernel(range) = (x, y) -> y - round((y - x) / range) * range
 
 """
-    unwrap(m; dims=nothing, range=2pi)
+    unwrap(m; kwargs...)
 
 
 Assumes `m` to be a sequence of values that has been wrapped to be inside the
--- a/src/unwrap.jl
+++ b/src/unwrap.jl
@@ -293,11 +293,11 @@ function calculate_reliability(pixel_image::AbstractArray{T, N}, circular_dims, 
             end
             border_begin          = fill(2, N)
             border_begin[idx_dim] = size_img[idx_dim]
-            border_begin          = CartesianIndex{N}(border_begin...)
+            border_begin_cart     = CartesianIndex{N}(border_begin...)
             border_end            = collect(size_img)-1
             border_end[idx_dim]   = size_img[idx_dim]
-            border_end = CartesianIndex{N}(border_end...)
-            for i in CartesianRange(border_begin, border_end)
+            border_end_cart       = CartesianIndex{N}(border_end...)
+            for i in CartesianRange(border_begin_cart, border_end_cart)
                 @inbounds pixel_image[i].reliability = calculate_pixel_reliability(pixel_image, i, pixel_shifts_border, range)
             end
             # second border
@@ -312,27 +312,27 @@ function calculate_reliability(pixel_image::AbstractArray{T, N}, circular_dims, 
             end
             border_begin          = fill(2, N)
             border_begin[idx_dim] = 1
-            border_begin          = CartesianIndex{N}(border_begin...)
+            border_begin_cart     = CartesianIndex{N}(border_begin...)
             border_end            = collect(size_img)-1
             border_end[idx_dim]   = 1
-            border_end = CartesianIndex{N}(border_end...)
-            for i in CartesianRange(border_begin, border_end)
+            border_end_cart       = CartesianIndex{N}(border_end...)
+            for i in CartesianRange(border_begin_cart, border_end_cart)
                 @inbounds pixel_image[i].reliability = calculate_pixel_reliability(pixel_image, i, pixel_shifts_border, range)
             end
         end
     end
 end
 
-function calculate_pixel_reliability(pixel_image::AbstractArray{T, N}, pixel_index, pixel_shifts, range) where {T, N}
-    sum_val = zero(fieldtype(T, :val))
+function calculate_pixel_reliability(pixel_image::AbstractArray{Pixel{T}, N}, pixel_index, pixel_shifts, range) where {T, N}
+    sum_val = zero(T)
     for pixel_shift in pixel_shifts
-        @inbounds sum_val += (wrap_val(pixel_image[pixel_index+pixel_shift].val - pixel_image[pixel_index].val), range)^2
+        @inbounds sum_val += wrap_val(pixel_image[pixel_index+pixel_shift].val - pixel_image[pixel_index].val, range)^2
     end
     return sum_val
 end
 
 # specialized pixel reliability calculations for different N
-function calculate_pixel_reliability(pixel_image::AbstractArray{T, 2}, pixel_index, pixel_shifts, range) where T
+function calculate_pixel_reliability(pixel_image::AbstractArray{Pixel{T}, 2}, pixel_index, pixel_shifts, range) where T
     @inbounds D1 = wrap_val(pixel_image[pixel_index+pixel_shifts[2]].val - pixel_image[pixel_index].val, range)
     @inbounds D2 = wrap_val(pixel_image[pixel_index+pixel_shifts[4]].val - pixel_image[pixel_index].val, range)
     @inbounds H  = wrap_val(pixel_image[pixel_index+pixel_shifts[6]].val - pixel_image[pixel_index].val, range)
@@ -340,8 +340,8 @@ function calculate_pixel_reliability(pixel_image::AbstractArray{T, 2}, pixel_ind
     return H*H + V*V + D1*D1 + D2*D2
 end
 
-function calculate_pixel_reliability(pixel_image::AbstractArray{T, 3}, pixel_index, pixel_shifts, range) where T
-    sum_val = zero(fieldtype(T, :val))
+function calculate_pixel_reliability(pixel_image::AbstractArray{Pixel{T}, 3}, pixel_index, pixel_shifts, range) where T
+    sum_val = zero(T)
     @inbounds sum_val += (wrap_val(pixel_image[pixel_index+pixel_shifts[1]].val - pixel_image[pixel_index].val, range))^2
     @inbounds sum_val += (wrap_val(pixel_image[pixel_index+pixel_shifts[2]].val - pixel_image[pixel_index].val, range))^2
     @inbounds sum_val += (wrap_val(pixel_image[pixel_index+pixel_shifts[3]].val - pixel_image[pixel_index].val, range))^2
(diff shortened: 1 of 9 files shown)


martinholters:
> There is no longer a `dim` keyword, a user who wants to recover that functionality will have to pass the appropriate `view`s.

AFAICT, it wasn't a keyword, but a positional argument. Anyway, we should (at least) add a deprecation, to avoid breaking anyones code. (I'm also not quite sure what you mean with the "appropriate `view`s", so a deprecation will clarify things for me.)

platawiec:
Sure, will add the deprecation warning.

The previous algorithm would give errors as seen in Fig. 5 (g, h) here: [pdf](https://www.ljmu.ac.uk/~/media/files/ljmu/about-us/faculties-and-schools/tae/geri/two_dimensional_phase_unwrapping_finalpdf.pdf?la=en)

For multi-dimensional unwrapping this algorithm is far more robust, so if someone was interested in doing a true n-d unwrap they shouldn't be using the old algorithm. However, if they have some array `A::Matrix{T}`, where adjacent columns are independent, then they could instead perform unwrapping on each column:

```
for i in 1:size(A, 2)
    unwrap!(view(A, :, i))
end
```

platawiec:
Added a deprecation warning by preserving the old code. Existing code will be unaffected, but users will be notified that there is now a better option available if their code is > one-dimensional.

src/unwrap.jl
```diff
@@ -0,0 +1,370 @@
...
+
+function unwrap(m::AbstractArray{T, N}, args...; kwargs...) where {T<:AbstractFloat, N}
+    unwrap!(copy(m), args...; kwargs...)
+end
+
+### Deprecated unwrap function
+function unwrap!(m::Array{T}, dim::Integer; range::Number=2pi) where T<:AbstractFloat
+    Base.depwarn(string("`unwrap!(m::Array{T}, dim::Integer; range::Number=2pi) ",
+        " where T<:AbstractFloat` is deprecated, ",
+        "use `unwrap!(m; range)` instead if your data has more than one dimension."),
```
martinholters: This doesn't tell me how I get the old behavior that I'd need if I have e.g. multiple one-dimensional signals collected in a matrix that I'd like to individually unwrap.

platawiec:
Added some more detail to the depwarn. I didn't add a code example for brevity, and because passing a n-dimensional view of n-dimensional data embedded in an n<m-dimensional array is generally applicable advice. A code example would lose generality.

martinholters:
What I had been imagining was more like
```julia
@deprecate(unwrap!(m::Array{T}, dim::Integer; range::Number=2pi) where T<:AbstractFloat,
    whatever_code_construct_is_needed_here)
```
If `whatever_code_construct_is_needed_here` turns out to be too long, that might be an indication that  we actually don't want to deprecate the current method. OTOH, I agree that writing this for the N-dimensional case is both non-trivial and probably rarely needed (if at all). 

Another thing to note is that before this PR, the `dim` argument defaulted to `ndims(m)`, so the meaning of `unwrap(X)` for `X` having more than one dimension would silently change. If anyone was relying on the present behavior, that would come as a nasty surprise.

What we could do is deprecate the multi-dimensional case now, and add the new version later (or even now but with a different name). However, I'm all ears for better ideas. Anyone else willing to chime in? @ssfrr perhaps?

platawiec:
I had tried something with the `@deprecate` macro, but because the `unwrap(m, args...; kwargs...)` function signature is generic, it ended up "deprecating" function calls which should not have been. It's possible I mis-implemented something, but this is the behavior I observed going the `@deprecate` route. This is also why I dropped the `dim` argument, that way only people who explicitly used the `dim` functionality would be warned. Unfortunately, as you pointed out, this means the behavior is silently changed for people simple calling `unwrap!(m)`.

> Another thing to note is that before this PR, the dim argument defaulted to ndims(m), so the meaning of unwrap(X) for X having more than one dimension would silently change. If anyone was relying on the present behavior, that would come as a nasty surprise.

This is the main issue I see as well, and I think it's important to get it right. At the moment, I feel we may be prioritizing a valid, but "side-effect" use case (unwrap 1D data stored in N-D array) over the intended use case (unwrap N-D data). The previously implemented algorithm does the intended use case (poorly), via the side-effect. A complicating factor is that the previous doc-string seems to sanction the side-effect use case - I suppose this is coming from people who are more used to Audio/time-series DSP vs. image processing.

I'm fine with doing a staged deprecation, with a differently-named function in the interim, as long as the deprecation time frames are set beforehand and the interface is unified into a common `unwrap!` function in the end.

martinholters:
I'd say what you call the "intended use case" is not handled by the current `unwrap`, it's only aiming at the "side-effect use case". And I whole-heartedly agree that we should add the proper multi-dimensional `unwrap` and that it should eventually be the default thing to do for `unwrap(A)`. I'm not entirely convinced we should deprecated the case with `dim` specified, though. If we look at the direction base has taken (e.g. `cumsum` is somewhat comparable), the following API might make sense (optional `range` keyword argument omitted for brevity):
* `unwrap(A)`: multi-dimensional unwrap as added by this PR; coinciding with current behavior for one-dimensional `A`
* `unwrap(A, dims=n)`: one-dimensional unwrap along dimension `n` (currently `unwrap(A, n)`)

The mutating version should probably take input and output arguments, which may be equal for true in-place operation.

This suggests the following route:
1a. Deprecate `unwrap(A, dim)` to `unwrap(A; dims=dim)` and `unwrap(A)` where `A` has more than one dimension to `unwrap(A; dims=ndims(A))`. Deprecate `unwrap!(A, dim)` to `unwrap!(A, A, dims=dim)` and `unwrap!(A)` to `unwrap!(A, A)` (for one-dimensional `A`) or `unwrap!(A, A, dims=ndims(A))` (for multi-dimensional `A`).
1b. Add the multi-dimensional unwrap with a distinct name.
2. Tag a release, let some time pass for users to adjust their code.
3. Rename the multi-dimensional unwrap to `unwrap` (with a deprecation).
4. Tag a release, let some time pass for users to adjust their code.
5. Remove the old name for the multi-dimensional unwrap.

If we can agree on something like this, I could take care of 1a while you think of a nice name for 1b :smile:

platawiec:
Sounds like a plan, very clear thanks.

So the multi-dimensional unwrap should look something like `unwrapimage!(dest, src)`?

ssfrr:
I haven't looked into the code deeply yet, but from an API standpoint I support a `dim` argument vs. relying on `view`, though I think that [`sum`](https://docs.julialang.org/en/latest/base/collections/#Base.sum) might be a better model than `cumsum`. `cumsum` is an inherently 1D operation that could be embedded inside a higher-dimensional `Array`. `sum` is a multi-dimensional operation like the new `unwrap`, (but that could also be embedded inside a higher-dimensional `Array`) so rather than a `dim` argument there's `dims` that describes along which dimensions you want to unwrap.

That said, I think it's OK to currently only support the 1D case where `dims::Integer`, and then in the future we could extend to support iterable `dims` without breaking anything.

So my vote would be for an interface like:
* `unwrap(A)` - multidimensional unwrap
* `unwrap(A; dims=2)` - 1D unwrap along dimension 2
* `unwrap(A; dims=(1,2))` - 2D unwrap inside a higher-dimensional array (like a 3D Array representing frames of a video) (maybe to be implemented in the future)

I'm on-board with a 2-step deprecation where `unwrap(A)` deprecated for multidimensional arrays. I don't have a good sense for the right timeframe - maybe a few months?

One alternative to a separate function `unwrapimage` would be `unwrap(A, dims=1:ndims(A))`, basically we just force that people specify what they want explicitly during the deprecation while we change the default.

as a side note: [`fft`](https://juliamath.github.io/AbstractFFTs.jl/latest/api.html#Public-Interface-1) seems like it would be an appropriate model, but it uses `dims` as a positional argument rather than keyword. I [filed an issue](https://github.com/JuliaMath/AbstractFFTs.jl/issues/15) to ask whether it makes sense to unify with the Base convention.

platawiec:
Great discussion points, thanks @ssfrr . I think the analogy to `fft` is good, where multi-dimensional is the default.

Just wanted to point out, for educational reason:
> `unwrap(A; dims=(1,2))` - 2D unwrap inside a higher-dimensional array (like a 3D Array representing frames of a video) (maybe to be implemented in the future)

Frames of a video should be unwrapped via a 3D unwrap - presumably the frames of a video are looking at an object in time, and the phase information is not entirely scrambled between frames. Because there is a dimensional relationship, 3D unwrap would be the appropriate function to call. If, on the other hand, a 3D array was storing a bunch of 2D images, then the signature you quote would be correct.

ssfrr:
> Frames of a video should be unwrapped via a 3D unwrap - presumably the frames of a video are looking at an object in time, and the phase information is not entirely scrambled between frames. Because there is a dimensional relationship, 3D unwrap would be the appropriate function to call. If, on the other hand, a 3D array was storing a bunch of 2D images, then the signature you quote would be correct.

Ah yes, excellent point. Assume I meant a collection of unrelated images then. ;)

martinholters:
> One alternative to a separate function `unwrapimage` would be `unwrap(A, dims=1:ndims(A))`

Yes, that's much better. So the revised plan:
1a. Deprecate `unwrap(A, dim)` to `unwrap(A; dims=dim)` and `unwrap(A)` where `A` has more than one dimension to `unwrap(A; dims=ndims(A))`. Deprecate `unwrap!(A, dim)` to `unwrap!(A, A, dims=dim)` and `unwrap!(A)` to `unwrap!(A, A)` (for one-dimensional `A`) or `unwrap!(A, A, dims=ndims(A))` (for multi-dimensional `A`).
1b. Add the multi-dimensional unwrap for `unwrap(A, dims=1:ndims(A))`.
2. Tag a release, let some time pass for users to adjust their code.
3. Make `dims=1:ndims(A)` the default.

martinholters:
Can you rebase this on top of #198?

platawiec:
Yup, I'll have a look, thanks everyone for getting #201 over the finish line.

platawiec:
Alright, I think I've implemented the functionality as @martinholters laid out above. Most of the changes are in my code, but I had to make a few changes in #201 where tests were failing because the `range` keyword was set to `2pi`, causing it to lose precision when dealing with non-Float64 numbers.

I haven't tested against 0.7 yet.

src/unwrap.jl
```diff
@@ -0,0 +1,373 @@
...
+
+Assumes `m` (along the single dimension `dims`) to be sequences of values that
+have been wrapped to be inside the given `range` (centered around
+zero), and undoes the wrapping by identifying discontinuities.
+
+A common usage is for a phase measurement over time, such as when
+comparing successive frames of a short-time-fourier-transform, as
+each frame is wrapped to stay within (-pi, pi].
+
+If `dims` is a `Range`, then the data is unwrapped according to
```
martinholters: I think this whole docstring needs to be rewritten to make sense.  The first two paragraphs and this one just don't fit together.

src/unwrap.jl
```diff
@@ -0,0 +1,373 @@
...
+the number of dimensions passed to `dims`.
+
+# Arguments
+- `m::AbstractArray{T, N}`: Array to unwrap
+- `range=2pi`: Range of wrapped array
+- `wrap_around=`:  When an element of this tuple is `true`, the
```
martinholters: I find `wrap_around` a bit unfortunate due to the overloaded meaning of "wrap" here. How about `circular_dims`?

src/unwrap.jl
```diff
@@ -0,0 +1,373 @@
...
+    unwrapping process will consider the edges along the corresponding axis
+    of the image to be connected
```
martinholters: "image" is unfortunate. "array"?

src/unwrap.jl
```diff
@@ -0,0 +1,373 @@
...
+- `seed=-1`: Unwrapping of arrays with dimension > 1 uses a random initialization. This
```
martinholters: Why pass the seed, if you can just manually seed the RNG before calling `unwrap`? Alternatively, one might wish to pass the RNG.

src/unwrap.jl
```diff
@@ -0,0 +1,373 @@
...
+    sets the seed of the RNG.
+"""
+function unwrap(m::AbstractArray{T,N}; dims=nothing, range=2T(pi), kwargs...) where {T, N}
```
martinholters: No need for the explicit `range` here, can just be passed on as part of `kwargs...`

test/unwrap.jl
```diff
@@ -0,0 +1,140 @@
...
+        f_wr = f_uw .% (2convert(T, π))
+        uw_test = unwrap(f_wr, dims=1:3, wrap_around=(true, true, true))
+        offset = first(f_uw) - first(uw_test)
+        @test (uw_test+offset) ≈ f_uw #oop, 3wrap
+        # test in-place version
+        unwrap!(f_wr, dims=1:3, wrap_around=(true, true, true))
+        offset = first(f_uw) - first(f_wr)
+        @test (f_wr+offset) ≈ f_uw #oop, 3wrap
+    end
+end
```
martinholters: IIUC, `calculate_pixel_reliability` is special-cased for 2d and 3d. So maybe even test for even more dimensions?

src/unwrap.jl
```diff
@@ -0,0 +1,375 @@
...
+    @inbounds sum_val += (wrap_val(pixel_image[pixel_index+pixel_shifts[7]].val - pixel_image[pixel_index].val, range))^2
+    @inbounds sum_val += (wrap_val(pixel_image[pixel_index+pixel_shifts[8]].val - pixel_image[pixel_index].val, range))^2
+    @inbounds sum_val += (wrap_val(pixel_image[pixel_index+pixel_shifts[9]].val - pixel_image[pixel_index].val, range))^2
+    @inbounds sum_val += (wrap_val(pixel_image[pixel_index+pixel_shifts[10]].val - pixel_image[pixel_index].val, range))^2
+    @inbounds sum_val += (wrap_val(pixel_image[pixel_index+pixel_shifts[11]].val - pixel_image[pixel_index].val, range))^2
+    @inbounds sum_val += (wrap_val(pixel_image[pixel_index+pixel_shifts[12]].val - pixel_image[pixel_index].val, range))^2
+    @inbounds sum_val += (wrap_val(pixel_image[pixel_index+pixel_shifts[13]].val - pixel_image[pixel_index].val, range))^2
+    @inbounds sum_val += (wrap_val(pixel_image[pixel_index+pixel_shifts[15]].val - pixel_image[pixel_index].val, range))^2
+    @inbounds sum_val += (wrap_val(pixel_image[pixel_index+pixel_shifts[17]].val - pixel_image[pixel_index].val, range))^2
```
martinholters: Omission of 16 here is intentional?

src/unwrap.jl
```diff
@@ -0,0 +1,375 @@
...
+    @inbounds sum_val += (wrap_val(pixel_image[pixel_index+pixel_shifts[23]].val - pixel_image[pixel_index].val, range))^2
+    @inbounds sum_val += (wrap_val(pixel_image[pixel_index+pixel_shifts[24]].val - pixel_image[pixel_index].val, range))^2
+    @inbounds sum_val += (wrap_val(pixel_image[pixel_index+pixel_shifts[25]].val - pixel_image[pixel_index].val, range))^2
+    @inbounds sum_val += (wrap_val(pixel_image[pixel_index+pixel_shifts[26]].val - pixel_image[pixel_index].val, range))^2
+    @inbounds sum_val += (wrap_val(pixel_image[pixel_index+pixel_shifts[27]].val - pixel_image[pixel_index].val, range))^2
+    return sum_val
+end
+
+A = rand(10)
+unwrap(A; range=10)
```
martinholters: Forgot to delete these?

src/unwrap.jl
platawiec: Good eye... not intentional

martinholters: Have benchmarked whether the manual unrolling is any faster than a for loop?

platawiec: Yup, the hand-unrolled `calculate_reliability` function is modestly (~15%) faster, with 2 fewer allocations. This isn't where most of the time spent is, though. When I profiled, most of the time was spent in `sort!` for moderately-sized arrays (1000x1000) and `gather_pixels!` for even larger arrays.

When I re-checked my benchmarks, I was able to track down a small type-instability that snuck through.

platawiec:
Quick ping, is there anything else you'd like to see?

src/unwrap.jl
```diff
@@ -0,0 +1,375 @@
+module Unwrap
+
+import Compat: selectdim
+import Compat
+
+export unwrap, unwrap!
+
+"""
+    unwrap!(m; dims=nothing, range=2pi)
```
martinholters: Does not mention all keywords args. Maybe just `m; kwargs...`?

src/unwrap.jl
```diff
@@ -0,0 +1,375 @@
...
+function unwrap!(m::AbstractArray{T,N}; dims=nothing, range=2T(pi), kwargs...) where {T, N}
+    if dims === nothing && N != 1
+        Base.depwarn("`unwrap!(m::AbstractArray)` is deprecated, use `unwrap!(m, dims=ndims(m))` instead", :unwrap!)
+        dims = N
+    end
+    unwrap!(m, m; dims=dims, range=range, kwargs...)
+end
+
+"""
+    unwrap!(y, m; dims=nothing, range=2pi)
```
martinholters: Same here. Just `y, m; kwargs...`?

src/unwrap.jl
```diff
@@ -0,0 +1,375 @@
...
+"""
+function unwrap!(y::AbstractArray{T,N}, m::AbstractArray{T,N}; dims=nothing, range::Number=2T(pi), kwargs...) where {T, N}
+    if dims === nothing
+        if N != 1
+            throw(ArgumentError("`unwrap!`: required keyword parameter dims missing"))
+        end
+        dims = 1
+    end
+    if dims isa Integer
+        Compat.accumulate!(unwrap_kernel(range), y, m, dims=dims)
```
martinholters: Should probably check that `kwargs` is empty and throw an error otherwise.

src/unwrap.jl
```diff
@@ -0,0 +1,375 @@
...
+    else
+        throw(ArgumentError("`unwrap!`: Invalid dims specified: $dims"))
+    end
+    return y
+end
+
+unwrap_kernel(range=2π) = (x, y) -> y - round((y - x) / range) * range
+
+"""
+    unwrap(m; dims=nothing, range=2pi)
```
martinholters: Either list all keyword args here or again just `kwargs...`.

src/unwrap.jl
```diff
@@ -0,0 +1,375 @@
...
+comparing successive frames of a short-time-fourier-transform, as
+each frame is wrapped to stay within (-pi, pi].
+
+A common usage for unwrapping across multiple dimensions is for a phase
+measurement of a scene, such as when retrieving the phase information of
+of an image, as each pixel is wrapped to stay within (-pi, pi].
+
+# Arguments
+- `m::AbstractArray{T, N}`: Array to unwrap
+- `range=2pi`: Range of wrapped array
```
martinholters: Should add `dims` to this list, too.

martinholters:
Also, if you could try to eliminate the deprecation warnings on 0.7, that would be appreciated. Otherwise, I can have a go over it after merging.

platawiec:
It's done, 0.7 dep warns are taken care of too.

Thanks for keeping your eye on this, really appreciate the help and fine-tuning!

src/unwrap.jl
```diff
@@ -0,0 +1,378 @@
...
+- `m::AbstractArray{T, N}`: Array to unwrap
+- `dims=nothing`: Dimensions along which to unwrap. If `dims` is an integer, then
+    `unwrap` is called that dimension. If `dims=1:ndims(m)`, then `m` is unwrapped
+    across all dimensions.
+- `range=2pi`: Range of wrapped array
+- `circular_dims=(false, ...)`:  When an element of this tuple is `true`, the
+    unwrapping process will consider the edges along the corresponding axis
+    of the array to be connected
+- `rng=GLOBAL_RNG`: Unwrapping of arrays with dimension > 1 uses a random
+    initialization. A user can be pass their own RNG through this argument.
```
martinholters: "can ~be~ pass"

martinholters:
Thanks for not giving up here! Save for the one type, LGTM now.

martinholters:
@ssfrr, do you want to take another look? Otherwise, I'll merge soonish.

ssfrr:
I've got a pretty long queue ATM and don't want to hold this up any longer, so I say if you're happy with it go for it!

Thanks both of you for bringing it home.

PR #201, opened 2018-05-16
Modernize unwrap API and prepare for #198

martinholters:
Make the API of `unwrap`/`unwrap!` similar to the one of `accumulate`/`accumulate!` and friends and deprecate the case of unspecified `dims` to make room for proper multi-dimensional unwrapping
(#198), CC @platawiec.

Also switch the implementation to use `accumulate!` for simpler and more efficient code.

I've realized that `unwrap!(y, x)` could actually use default `dims=1:ndims(x)` right now (well, once the functionality of #198 is in), but that would be very inconsistent with `unwrap(x)` defaulting to `dims=ndims(x)` (with a deprecation warning). So I suggest we provide no default `dims` for `unwrap!` until we remove that deprecation and then use the same default in both.

### DIFF
--- a/src/util.jl
+++ b/src/util.jl
@@ -1,6 +1,7 @@
 module Util
 using ..DSP: @importffts, mul!
 import Base: *
+import Compat
 using Compat: copyto!, ComplexF64, selectdim, undef
 import Compat.LinearAlgebra.BLAS
 @importffts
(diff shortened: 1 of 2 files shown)


martinholters:
Can you give this a look, @platawiec and @ssfrr?

src/util.jl
```diff
@@ -29,46 +30,58 @@ export  unwrap!,
...
+        throw(ArgumentError("`unwrap!`: Invalid dims specified: $dims"))
     end
-    return m
+    return y
 end
 
+unwrap_kernel(range=2π) = (x, y) -> y - round((y - x) / range) * range
+
+@deprecate(unwrap!(m::Vector{T}; range::Number=2pi) where T<:AbstractFloat,
+    unwrap!(m, m, range=range))
```
ssfrr: remove this deprecation if you end up going with the in-place fallback suggestion above.

src/util.jl
```diff
@@ -29,46 +30,58 @@ export  unwrap!,
...
-        d = selectdim(m, dim, i:i) - selectdim(m, dim, i-1:i-1)
-        slice_tuple = ntuple(n->(n==dim ? (i:i) : (1:size(m,n))), ndims(m))
-        offset = floor.((d.+thresh) / (range)) * range
-#        println("offset: ", offset)
-#        println("typeof(offset): ", typeof(offset))
-#        println("typeof(m[slice_tuple...]): ", typeof(m[slice_tuple...]))
-#        println("slice_tuple: ", slice_tuple)
-#        println("m[slice_tuple...]: ", m[slice_tuple...])
-        m[slice_tuple...] = m[slice_tuple...] - offset
+    if dims isa Integer
```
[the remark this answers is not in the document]
martinholters: That was my thinking. Might be reconsidered as part of #198.

ssfrr:
Sounds good. I'm :+1: to merge with the tweaks above

martinholters:
Ok, I think I have addressed all your comments. I've widened the type signatures to just `AbstractArray` everywhere. Leaving out the `T <: AbstractFloat` allows e.g.:
```julia
julia> unwrap([1,2,3,14,15,6], range=10)
6-element Array{Int64,1}:
 1
 2
 3
 4
 5
 6
```
Not that this seems like a particularly useful feature, but why disallow it?
On the down side, the error here is maybe a little unintuitive now:
```
julia> unwrap([1,2,3,14,15,6])
ERROR: InexactError: Int64(Int64, 1.4336293856408275)
```

We could make `unwrap` call `accumulate` directly, as this works:
```julia
accumulate(DSP.Util.unwrap_kernel(), [1,2,3,14,15,6], dims=1)
```
It returns an `Array{Float64}`. However, just like that, it would also return an `Array{Float64}` even if `range` is an `Int`. But we don't have to sort that out in this PR, I'd say.

src/util.jl
```diff
@@ -29,46 +30,67 @@ export  unwrap!,
...
 A common usage is for a phase measurement over time, such as when
 comparing successive frames of a short-time-fourier-transform, as
 each frame is wrapped to stay within (-pi, pi].
-
 """
-function unwrap(m::Array{T}, args...; kwargs...) where T<:AbstractFloat
-    unwrap!(copy(m), args...; kwargs...)
+function unwrap(m::AbstractArray{<:Any,N}; dims=nothing, range=2pi) where {N}
+    if dims === nothing && N != 1
+        Base.depwarn("`unwrap(m):AbstractArray` is deprecated, use `unwrap(m, dims=ndims(m))` instead", :unwrap)
```
ssfrr: whoops, I missed this last time - should this be `unwrap(m::AbstractArray)` rather than `unwrap(m):AbstractArray`?

Other than that I think this is good to merge.

martinholters: Good catch! I probably had too little coffee before reworking the deprecation messages for the widened type signatures. Update incoming...

martinholters:
Ok, one (hopefully) last look?

ssfrr:
Something weird is going on with coverage - most of the lines it says are uncovered seem to be unrelated to this PR. I don't have time to look into it now but hopefully it'll resolve itself.

martinholters:
Thanks, having a second pair of eyes here definitely was worth it.

I think there is something fishy with to coverage data on 0.7, but I haven't looked at the details, yet.

PR #205, opened 2018-06-04
Add missing window tests and fix `gaussian` window function

### LINKED ISSUE #204: Misplaced parens in gaussian window definition
The current [gaussian window](https://github.com/JuliaDSP/DSP.jl/blob/b358239fdbaeb9061a0337964b652d7ddced61a8/src/windows.jl#L131-L136) seems similar to the [wikipedia description](https://en.wikipedia.org/wiki/Window_function#Gaussian_window) except that the grouping in the denominator is wrong (`n-1/2` vs `(n-1)/2`).

There's no reference in the docs for what is intended, but this doesn't seem right.

Also if you add the parens to get:
```julia
[exp(-0.5*((k-(n-1)/2)/(sigma*(n-1)/2))^2) for k=0:(n-1)]
```

You get the same answer as
```julia
g(x) = exp(-0.5*(x/sigma)^2)
g.(linspace(-1.0, 1.0, n)
```

Which seems like a pretty reasonable definition (defining `sigma` as the std. dev as a fraction of the total window length, so `sigma=0.5` corresponds to the right and left edges of the window being 1 stddev away).

The other issue with the current implementation is that it breaks the principle that the discrete windows are just sampled versions of the underlying continuous function (`triang` also has this property, but I think that's expected). With the current implementation the shape of the window changes slightly as you vary `n`.

If this is a bug and we want to fix it, deprecation is a bit of a challenge. Perhaps during the deprecation period we define the function:

```julia
function gaussian(n, sigma; legacy=true)
    legacy && depwarn(
        """
        As of DSP v0.5.0 `sigma` is interpreted as the std dev as a fraction of the window
        length. use `gaussian(n, sigma; legacy=false)` to opt into the new definition
        during the deprecation period. After the deprecation period this will be the default.
        """, :legacy_gaussian)
    if legacy
        # old definition...
    else
        # new definition
end
```

I do worry a little about this deprecation strategy, because after the deprecation window is over any old code that didn't get updated will be silently broken. I'm not sure if there's a sensible interpretation for the current meaning of `sigma`, but I wouldn't be surprised if there's code out there that just eyeballed a reasonable value for it, and those windows will be messed up.

On the other hand, the package is not yet at 1.0, and it would be nice to have a `sigma` value that meant something.

I can do the PR, I just wanted to get some feedback on this strategy.

ssfrr:
This PR fixes #204 by fixing the gaussian window. It also:

1. adds some tests for other windows that were missing
2. adds a more detailed doc string to `gaussian`.
3. switches to using a `σ` argument rather than spelling out `sigma` - it's a positional arg so it doesn't get typed by the user and this way the code matches the LaTeX more closely.
4. removes the limitation that `σ <= 0.5` - it seems like an arbitrary cutoff. It still enforces positive σ, though.

A comment I added explaining the tests assumes this will get folded into a v0.5.0 release.


src/windows.jl
```diff
@@ -123,16 +123,16 @@ function bartlett(n::Integer)
...
+correspond to 1 standard deviation away from the center.
 """
-function gaussian(n::Integer, sigma::Real)
-    if !(0 < sigma <= 0.5)
-        error("sigma must be greater than 0 and less than or equal to 0.5.")
-    end
-    [exp(-0.5*((k-(n-1)/2)/(sigma*(n-1/2)))^2) for k=0:(n-1)]
+function gaussian(n::Integer, σ::Real)
+    σ > 0.0 || error("σ must be positive")
+    [exp(-0.5*((k-(n-1)/2)/(σ*(n-1)))^2) for k=0:(n-1)]
```
ararslan: I realize this is preexisting, but note that 0.5 is a `Float64` literal, so this will unconditionally promote everything to `Float64`. It would be safer (in some sense) to simply divide by 2.

ssfrr: we'll already be in `Float64` land because of the `(k-(n-1)/2)`

martinholters: If we use `[exp(-((2*k-(n-1))/(2*σ*(n-1)))^2 / 2) for k=0:(n-1)]`, one could get a `Float32`-valued window by specifying σ as a `Float32`. But the other window functions also just return `Float64`s, so I doubt it's worth it.

test/windows.jl
```diff
@@ -40,6 +49,22 @@ end
     kaiser_jl = kaiser(128, 0.4/π)
     kaiser_ml = readdlm(joinpath(dirname(@__FILE__), "data", "kaiser128,0.4.txt"), '\t')
     @test kaiser_jl ≈ kaiser_ml
+
+    gaussian_jl = gaussian(128, 0.2)
+    gaussian_ref = readdlm(joinpath(dirname(@__FILE__), "data", "gaussian128,0.2.txt"), '\t')
+    @test gaussian_jl ≈ gaussian_ref
```
martinholters: Could also test against `w(x) = exp(-0.5*(x/sigma)^2); w.(linspace(-0.5, 0.5, n)` to verify it matches the docstring.

PR #207, opened 2018-07-05
Fix a few deprecations on Julia 0.7

martinholters:
Remaining deprecation warnings at the moment are all due to the moval of `mean` and `middle` to `Statistics`. I'll fix those once https://github.com/JuliaLang/Compat.jl/pull/583 has landed. Whether as part of this PR or in another one is a matter of how fast we get the Compat. I'd like to tag a version with the new `unwrap` soon, but wanted to fix those deprecations where it was easy. So if the Statistics thing takes its time, I'll tag before, despite the remaining deprecations.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,5 +1,5 @@
 julia 0.6
-Compat 0.63
+Compat 0.69
 Polynomials 0.1.0
 Reexport
 SpecialFunctions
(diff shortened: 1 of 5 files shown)


martinholters:
```
esprit: Test Failed at /home/travis/build/JuliaDSP/DSP.jl/test/estimation.jl:20
  Expression: isapprox(frequencies', frequencies_estimated; atol=0.01)
   Evaluated: isapprox([1000.0; 1250.0], [999.958, 1249.95]; atol=0.01)
```
in one of the CI runs. We should probably loosen the bound here.

PR #208, opened 2018-07-09
Add @simd macro to amenable loops

galenlynch:
A number of loops that satisfy the constraints to use the @simd macro did not. I
have added them to the loops that I could find, and found that performance and
memory use for `mt_pgram` increased by a few percent. While I expect other
performance gains to be similarly modest, it's a 'free' boost.

### DIFF
--- a/src/Filters/filt.jl
+++ b/src/Filters/filt.jl
@@ -542,7 +542,7 @@ function fftfilt(b::AbstractVector{T}, x::AbstractArray{T},
             mul!(tmp1, p2, tmp2)
 
             # Copy to output
-            for j = 0:min(L - 1, nx - off)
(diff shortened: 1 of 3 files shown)


galenlynch:
The failed test seems unrelated to my changes. In fact that test seems like it should randomly fail, as it used random data for the test and (apparently) does not always estimate the frequency with the specified precision.

galenlynch:
I am closing this for now, since there doesn't seem to be any interest.

ssfrr:
sorry for the delay - have you benchmarked these to see how much of a difference it makes?

galenlynch:
I did when I first made the PR, but the situation may have changed in julia 1.0. I primarily wanted better performance out of `fft2pow!`, but now the simd annotation makes exactly no difference.

PR #209, opened 2018-07-09
Make dpss type stable

galenlynch:
I was using dpss in one of my functions and noticed that the output of dpss is
not type stable. I have made it type stable, while maintaining the algorithm and
output of the function.

### DIFF
--- a/src/windows.jl
+++ b/src/windows.jl
@@ -185,37 +185,43 @@ follow the convention that the first element of the skew-symmetric
 (odd) tapers is positive. The time-bandwidth product is given by
 `nw`.
 """
-function dpss(n::Int, nw::Real, ntapers::Int=ceil(Int, 2*nw)-1)
+function dpss(n::Integer, nw::Real, ntapers::Integer=ceil(Int, 2*nw)-1)
     0 < ntapers <= n || error("ntapers must be in interval (0, n]")
     0 <= nw < n/2 || error("nw must be in interval [0, n/2)")
 
     # Construct symmetric tridiagonal matrix
     v = cospi(2*nw/n)
-    mat = SymTridiagonal([v*abs2((n - 1)/2 - i) for i=0:(n-1)],
-                         [0.5.*(i*n - abs2(i)) for i=1:(n-1)])
+    dv = Vector{Float64}(n)
+    ev = Vector{Float64}(n - 1)
+    @inbounds dv[1] = v * abs2((n - 1) / 2)
+    @inbounds @simd for i = 1:(n-1)
+        dv[i + 1] = v * abs2((n - 1) / 2 - i)
+        ev[i] = 0.5 * (i * n - i^2)
+    end
+    mat = SymTridiagonal(dv, ev)
 
     # Get tapers
     @static if VERSION < v"0.7.0-DEV.3159"
-        eigvec = eigfact!(mat, n-ntapers+1:n)[:vectors]
+        eigvec = eigfact!(mat, n-ntapers+1:n)[:vectors]::Array{Float64, 2}
     elseif VERSION < v"0.7.0-DEV.5211"
-        eigvec = eigfact!(mat, n-ntapers+1:n).vectors
+        eigvec = eigfact!(mat, n-ntapers+1:n).vectors::Array{Float64, 2}
     else
-        eigvec = eigen!(mat, n-ntapers+1:n).vectors
+        eigvec = eigen!(mat, n-ntapers+1:n).vectors::Array{Float64, 2}
     end
-    v = Compat.reverse(eigvec::Matrix{Float64}, dims=2)
+    rv = Compat.reverse(eigvec, dims=2)::Array{Float64, 2}
 
     # Slepian's convention; taper starts with a positive element
-    sgn = ones(size(v, 2))
-    for i = 2:2:size(v, 2)
-        s = 0
+    sgn = ones(size(rv, 2))
+    for i = 2:2:size(rv, 2)
+        s = zero(Float64)
         for j = 1:n
-            s = sign(v[j, i])
+            s = sign(rv[j, i])
             s != 0 && break
         end
         @assert s != 0
         sgn[i] = s
     end
-    rmul!(v, Diagonal(sgn))
+    rmul!(rv, Diagonal(sgn))
 end
 
 # Eigenvalues of DPSS, following Percival & Walden p. 390, exercise 8.1


src/windows.jl
```diff
@@ -185,37 +185,43 @@ follow the convention that the first element of the skew-symmetric
...
     v = cospi(2*nw/n)
-    mat = SymTridiagonal([v*abs2((n - 1)/2 - i) for i=0:(n-1)],
-                         [0.5.*(i*n - abs2(i)) for i=1:(n-1)])
+    dv = Vector{Float64}(n)
+    ev = Vector{Float64}(n - 1)
+    @inbounds dv[1] = v * abs2((n - 1) / 2)
+    @inbounds @simd for i = 1:(n-1)
+        dv[i + 1] = v * abs2((n - 1) / 2 - i)
+        ev[i] = 0.5 * (i * n - i^2)
+    end
```
martinholters: Is the performance gain by by using this `@inbounds @simd` loop compared to the (IMO more readable) comprehension significant/relevant?

src/windows.jl
```diff
@@ -185,37 +185,43 @@ follow the convention that the first element of the skew-symmetric
...
     # Get tapers
     @static if VERSION < v"0.7.0-DEV.3159"
-        eigvec = eigfact!(mat, n-ntapers+1:n)[:vectors]
+        eigvec = eigfact!(mat, n-ntapers+1:n)[:vectors]::Array{Float64, 2}
     elseif VERSION < v"0.7.0-DEV.5211"
-        eigvec = eigfact!(mat, n-ntapers+1:n).vectors
+        eigvec = eigfact!(mat, n-ntapers+1:n).vectors::Array{Float64, 2}
     else
-        eigvec = eigen!(mat, n-ntapers+1:n).vectors
+        eigvec = eigen!(mat, n-ntapers+1:n).vectors::Array{Float64, 2}
```
martinholters: I'd this can be inferred, like the one before. Only the `[:vectors]` version should need the type assertion.

src/windows.jl
```diff
@@ -185,37 +185,43 @@ follow the convention that the first element of the skew-symmetric
...
     end
-    v = Compat.reverse(eigvec::Matrix{Float64}, dims=2)
+    rv = Compat.reverse(eigvec, dims=2)::Array{Float64, 2}
```
martinholters: Personal taste: I'd still prefer `Matrix{Float64}` over `Array{Float64, 2}`. (Same applies above.)

src/windows.jl
```diff
@@ -185,37 +185,43 @@ follow the convention that the first element of the skew-symmetric
...
 
     # Slepian's convention; taper starts with a positive element
-    sgn = ones(size(v, 2))
-    for i = 2:2:size(v, 2)
-        s = 0
+    sgn = ones(size(rv, 2))
+    for i = 2:2:size(rv, 2)
+        s = zero(Float64)
```
martinholters: `0.0`?

davidavdav:
```julia
s = zero(eltype(rv))
```

src/windows.jl
galenlynch: I'm not really sure how to answer this question. There is a performance gain to using this `@simd` loop. My real use case, which motivatred this PR, is analyzing the spectral content of a bunch of recordings that have around `23756800` samples in them. In this case, with `nw` set to 8, the `@simd` versions takes `230.26` seconds while the `let` block version takes `229.60` seconds:

`@simd` version:
```julia
julia> @time dpss(23756800, 8)
230.260550 seconds (89 allocations: 10.089 GiB, 0.53% gc time)
```

`let` block version:
```julia
julia>@time dpss(23756800, 8)
229.606108 seconds (105 allocations: 10.089 GiB, 0.47% gc time)
```
This (very small) performance difference seems pretty reliable.

I didn't know, until your comment above, that a let block could be used to get rid of Julia's boxing of local variables used in comprehensions. I personally do not find list comprehensions inside a function call more readable than a regular for loop, but it's a questions of personal preference. I'm happy to replace it with list comprehensions inside of a let block, if you think that would be more consistent with the DSP.jl style.

src/windows.jl
galenlynch: I'm running on 6.3 and put it there out of caution since I couldn't check and it couldn't hurt. I'll remove the type assertions on the 0.7 branches.

src/windows.jl
galenlynch: Sure! Seems like an improvement. I didn't realize until looking around in DSP.jl that people use the `Matrix` alias. I'll change `Array{N, 2}` to `Matrix`.

src/windows.jl
galenlynch: definitely seems better.

src/windows.jl
galenlynch: My real question is this: why is this function so slow? After profiling it I see that the vast majority of its time is spent in lapack/blas, but it seems to be stuck in single-threaded eigen decomposition.

If I run `x=randn(5000,5000); eig(x);` on my laptop then it will use all of my cores to factor the matrix, while the eigenvalue problem in `dpss` pins a single core at 100% for ~4 minutes.

When I first started to poke at this I thought there could be some performance gains, from solving all the type instability. After making it type stable and seeing it was still slow I just decided to submit the PR as is.

martinholters: Actually, I'm not sure whether there needs to be a `let` here or just different variables for `v` and `rv` as you're doing already. Anyway, the loop version is also fine with me. I'm not 100% sold on the `@inbounds` and `@simd` annotations, because they will put a certain mental burden ("are they still valid?") on anyone making modifications here later on.

galenlynch: I just tried
```julia
    dv = [v*abs2((n - 1)/2 - i) for i=0:(n-1)]
    ev = [0.5.*(i*n - abs2(i)) for i=1:(n-1)]
    mat = SymTridiagonal(dv, ev)
```
And it loooks like julia can't determine the element type of dv and ev, and also boxes v to use in the comprehension.

Here's the perf. difference with and without @inbounds and @simd on a smaller test case:

with `@inbounds @simd`:
```julia
@benchmark dpss(200, 4, 7)
BenchmarkTools.Trial: 
  memory estimate:  67.31 KiB
  allocs estimate:  58
  --------------
  minimum time:     359.785 μs (0.00% GC)
  median time:      366.155 μs (0.00% GC)
  mean time:        389.776 μs (0.54% GC)
  maximum time:     1.684 ms (66.00% GC)
  --------------
  samples:          10000
  evals/sample:     1
```

Without them:
```
@benchmark dpss(200, 4, 7)
BenchmarkTools.Trial: 
  memory estimate:  67.31 KiB
  allocs estimate:  58
  --------------
  minimum time:     379.487 μs (0.00% GC)
  median time:      383.245 μs (0.00% GC)
  mean time:        404.403 μs (0.43% GC)
  maximum time:     2.219 ms (68.23% GC)
  --------------
  samples:          10000
  evals/sample:     1
```

So only ~5% speedup. Happy to remove it if you still want me to.

martinholters: Thanks for benchmarking that! I would have expected the eigen factorization to dominate the time even more. Fair to keep the annotations then.

galenlynch: No problem!

galenlynch:
My pleasure! Thanks for being a maintainer of such a useful package, and for the thoughtful code review!

PR #210, opened 2018-07-11
Remez

sirtom67:
Finally got a pure Julia implementation of the "remez" function. Speed is actually faster than C for the few long filters that I tried!
Unittests are added that test one or more of each type of filter (even and odd length, symmetric and asymmetric coefficients).


martinholters:
Thanks for doing this! I'll try to find some time to review; feel free to bump in case I forget. One thing I've noticed from a cursory glance is that some comments from the C code might better be reformatted as proper julia comments.

Oh, and there is a segfault on Julia nightly. Probably worth reducing and reporting in julia. I don't think you're doing anything unsafe here, and it's happening during codegen, not when running the code. Wild guess: The new optimizer fails to handle all those `@goto`s correctly.

martinholters:
Opened https://github.com/JuliaLang/julia/issues/28077 regarding the failure on 0.7.

src/Filters/remez_fir.jl
```diff
@@ -0,0 +1,845 @@
...
+f = linspace(0, 0.5, 1000)
+plot(f, 20*log10.(abs.(freqz(b,f,1.0))))
+plot(f, 20*log10.(abs.(freqz(b2,f,1.0))))
+grid()
+```
+"""
+function remez(numtaps::Integer, bands::Array, desired::Array; 
+               weight::Array=[], 
+               Hz::Real=1.0, 
+               filter_type::String="bandpass",
```
martinholters: `filter_type` should a `Symbol`, or an `@enum FilterType bandpass=1 differentiator=2 hilbert=3`. Then comparison is fast so that we can eliminate the `jtype` variable.

src/Filters/remez_fir.jl
```diff
@@ -0,0 +1,845 @@
...
+using PyPlot
+b = DSP.Filters.PolynomialRatio(bpass, [1.0]);
+b2 = DSP.Filters.PolynomialRatio(bpass2, [1.0]);
+f = linspace(0, 0.5, 1000)
+plot(f, 20*log10.(abs.(freqz(b,f,1.0))))
+plot(f, 20*log10.(abs.(freqz(b2,f,1.0))))
+grid()
+```
+"""
+function remez(numtaps::Integer, bands::Array, desired::Array;
```
martinholters: I think we should restrict these to `Vector`s. Or is there a valid reason to pass matrices or higher-dimensional arrays here?

src/Filters/remez_fir.jl
```diff
@@ -0,0 +1,845 @@
...
+    end
+
+    if (length(weight)==0)
+        weight = ones(desired)
+    end
+    bands = copy(bands)/Hz
+
+    bands = vec(bands)   # in C, known as "edge"
+    desired = Array{Float64,1}(vec(desired))
+    weight = Array{Float64,1}(vec(weight))
```
martinholters: I don't think we ever write to these, so we can avoid making a copy if the input already is of the desired type by calling `convert(Vector{Float64}, ...)`.

src/Filters/remez_fir.jl
```diff
@@ -0,0 +1,845 @@
...
+    # types based on starting letter of variable name LOL.
+    # jtype input:
+    #    Type I and II symmetric linear phase: neg==0   (jtype==1)
+    #    Type III and IV negative symmetric linear phase: neg==1   (jtype==2 or 3)
+    if (jtype == 1)
+      neg = 0
+    else
+      neg = 1     # "neg" means negative symmetry.
+    end
+    nodd = numtaps % 2   # nodd == 1: length is odd; nodd == 0: length is even
```
martinholters: Again, these could be booleans.

src/Filters/remez_fir.jl
```diff
@@ -0,0 +1,845 @@
...
+    nfcns = numtaps ÷ 2  # integer divide
+    if (nodd == 1 && neg == 0)
+        nfcns = nfcns + 1
+    end
+    temp = Float64(ngrid-1) / nfcns
```
martinholters: No need for the explicit `Float64` here.

src/Filters/remez_fir.jl
```diff
@@ -0,0 +1,845 @@
...
+    """/*
+     * SET UP A NEW APPROXIMATION PROBLEM WHICH IS EQUIVALENT
+     * TO THE ORIGINAL PROBLEM
+     */"""
+    if (neg <= 0)
+        if (nodd != 1)
+            for j = 1 : ngrid
+                change = cos(π*grid[j]);
+                des[j] = des[j] / change;
+                wt[j]  = wt[j] * change;
```
martinholters: Trailing `;` not needed. (There are a few throughout.

src/Filters/remez_fir.jl
```diff
@@ -0,0 +1,845 @@
...
+                change = sin(2π * grid[j]);
+                des[j] = des[j] / change;
+                wt[j]  = wt[j]  * change;
+            end
+        end
+    end
+        
+    iext = vec(zeros(Int64, dimsize))   # indices of extremals
+    x = vec(zeros(Float64, dimsize))
+    y = vec(zeros(Float64, dimsize))
```
martinholters: `vec` here is unnecessary.

src/Filters/remez_fir.jl
```diff
@@ -0,0 +1,845 @@
...
+    #    CALCULATION OF THE COEFFICIENTS OF THE BEST APPROXIMATION
+    #    USING THE INVERSE DISCRETE FOURIER TRANSFORM
+    #
+
+    a = vec(zeros(Float64, dimsize))   # frequency response on evenly spaced grid
+    p = vec(zeros(Float64, dimsize))
+    q = vec(zeros(Float64, dimsize)) 
+    alpha = vec(zeros(Float64, dimsize))   # return vector
+    
+    PI = 3.14159265358979323846
```
martinholters: Just use `pi`

src/Filters/remez_fir.jl
```diff
@@ -0,0 +1,845 @@
...
+    nm1 = nfcns - 1;   # nm1 => "nfcns minus 1"
+    fsh = 1.0e-06;
+    gtemp = grid[1];   # grid[1] is temporarily used in freq_eval in this loop
+    x[nzz] = -2.0;
+    cn  = 2*nfcns - 1;
+    delf = 1.0/cn;
+    l = 1;
+    kkk = 0;
+
+    if (bands[1] == 0.0 && bands[end] == 0.5) kkk = 1; end
```
martinholters: `kkk` also looks booleanesque to me. Any idea what it means? Can we get a more descriptive name?

src/Filters/remez_fir.jl
```diff
@@ -0,0 +1,845 @@
...
+
+
+"""
+build_grid(numtaps, bands, desired, weight, grid_density)
+return "grid" and "des" and "wt" arrays
+"""
+function build_grid(numtaps, bands, desired, weight, grid_density, jtype)
+    # translate from scipy remez argument names
+    L = numtaps
+    M = Int(floor(L / 2))
```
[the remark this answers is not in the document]
martinholters: Actually, as `L` is an integer, just `L ÷ 2` here.

src/Filters/remez_fir.jl
```diff
@@ -0,0 +1,845 @@
...
+ *  AN ARBITRARY FUNCTION OF FREQUENCY CAN BE
+ *  APPROXIMATED IF THE USER REPLACES THIS FUNCTION
+ *  WITH THE APPROPRIATE CODE TO EVALUATE THE IDEAL
+ *  MAGNITUDE.  NOTE THAT THE PARAMETER FREQ IS THE
+ *  VALUE OF NORMALIZED FREQUENCY NEEDED FOR EVALUATION.
+ *-----------------------------------------------------------------------
+ */
+"""
+function eff(freq::Float64, fx::AbstractVector, lband::Integer, jtype::Integer)
+    if (jtype != 2)
```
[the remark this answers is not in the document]
sirtom67: Done, with exceptions.
There are many single-line "if" statements, e.g.
```        if (l > 1) l = l-1; end```
I've left these in place - we can expand these out but it will expand the code quite a bit.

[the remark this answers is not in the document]
Comment (ssfrr) [reply src/Filters/remez_fir.jl]: those could be replaced with short-circuiting, e.g. in this case `l > 1 && (l = l-1)` IMO `cond && expr` is more idiomatic julia for one-line conditionals than `if cond expr; end`

**Comment by sirtom67** [comment]
@martinholters : I addressed your review comments - thanks for the review. Let me know any comments for the next pass.
@ssfrr : thank you for the comment, the short circuit syntax is interesting - should I switch over all these statements? Example:
```if (j == nzz) ynz = comp; end   # current```
```j == nzz && (ynz = comp)   # proposed```
The "if" syntax is more readable to me and more C-like. As you can tell, the Julia is pretty much a line-for-line translation of the C code. I was re-writing it but then came to the extremal update portion with all the gotos, and decided a faster path would be to just translate directly. That has the benefit of producing the SAME filter for the same inputs - I still use python and scipy and this way I can be sure julia gives the same result.
Anyway, that is an aside. I'm willing to defer to you for switching over to short-circuit, since I am new to this project and julia. I like it as is, but please advise.

File: src/Filters/remez_fir.jl [hunk 12]
```diff
@@ -0,0 +1,825 @@
...
+        end
+    end
+
+    if nfcns <= 3
+        alpha[nfcns+1] = alpha[nfcns+2] = 0.0
+    end
+
+    #for j = 1 : nfcns 
+    #    printfmtln("  j={}: alpha[j]={}", j, alpha[j])
+    #end
```
martinholters: Remove these.

src/Filters/remez_fir.jl
```diff
@@ -0,0 +1,825 @@
...
+    #for j = 1 : nfcns 
+    #    printfmtln("  j={}: a[j]={}", j, a[j])
+    #end
+
+    dden = 2π / cn
+    for j = 1 : nfcns
+        dtemp = 0.0
+        dnum = (j-1) * dden
+        if nm1 >= 1
+            for k = 1 : nm1
```
martinholters: The `if` here isn't necessary. If `nm < 1`, the loop won't be executed.

src/Filters/remez_fir.jl
```diff
@@ -0,0 +1,825 @@
...
+        if ((xt-xe) < fsh) @goto L415; end
+        grid[1] = ft
+        a[j] = freq_eval(1,nz,grid,x,y,ad)
+      @label L425
+        if (l > 1) l = l-1; end
+    end
+    grid[1] = gtemp  # restore grid[1]
+    #for j = 1 : nfcns 
+    #    printfmtln("  j={}: a[j]={}", j, a[j])
+    #end
```
martinholters: Remove this.

src/Filters/remez_fir.jl
```diff
@@ -0,0 +1,825 @@
...
+    x[nzz] = -2.0
+    cn  = 2*nfcns - 1
+    delf = 1.0/cn
+    l = 1
+
+    # Boolean for "kkk" in C code.
+    full_grid = (bands[1] == 0.0 && bands[end] == 0.5)
+    if (nfcns <= 3)
+        full_grid = true
+    end
```
martinholters: Just `full_grid = (bands[1] == 0.0 && bands[end] == 0.5) || nfcns <= 3` ?

src/Filters/remez_fir.jl
```diff
@@ -0,0 +1,825 @@
...
+        end
+        dev = dnum / dden
+        # printfmtln("DEVIATION = {}", dev)
+
+        y = 0*x
+        nu, dev = initialize_y(dev, nz, iext, des, wt, y)
+        
+        if dev <= devl
+            # finished 
+            return -1  # error - deviation should always increase
```
martinholters: I think we should either throw an exception or return `nothing`. Where I'd prefer the former.

src/Filters/remez_fir.jl
```diff
@@ -0,0 +1,825 @@
...
+        iext[nzz] = ngrid + 1
+        niter += 1
+        if niter > maxiter
+            break
+        end
+        
+        x[1:nz] = cospi.(2grid[iext[1:nz]])
+        #for j = 1 : nz
+        #    printfmtln("  j={}: iext[j]={} grid[iext[j]]={} x[j]={}", j, iext[j], grid[iext[j]], x[j])
+        #end
```
martinholters: Remove this.

src/Filters/remez_fir.jl
```diff
@@ -0,0 +1,825 @@
...
+        for j = 1 : nz
+            l = iext[j]
+            dnum += ad[j] * des[l]
+            dden += k * ad[j] / wt[l]
+            k = -k
+        end
+        dev = dnum / dden
+        # printfmtln("DEVIATION = {}", dev)
+
+        y = 0*x
```
martinholters: This allocates a new vector. Is that intended? Do the vector sizes change during the iteration? Otherwise just `fill!(y, 0.0)`.

src/Filters/remez_fir.jl
```diff
@@ -0,0 +1,825 @@
...
+        nfcns = nfcns + 1
+    end
+    temp = (ngrid-1) / nfcns
+    dimsize = Int64(ceil(numtaps/2.0 + 2))
+    #printfmtln("  dimsize={}   nfcns={}", dimsize, nfcns)
+
+    """/*
+     * SET UP A NEW APPROXIMATION PROBLEM WHICH IS EQUIVALENT
+     * TO THE ORIGINAL PROBLEM
+     */"""
```
martinholters: Julia comment.

src/Filters/remez_fir.jl
```diff
@@ -0,0 +1,825 @@
...
+    #end
+    
+    neg = filter_type != filter_type_bandpass      # boolean: "neg" means negative symmetry.
+    nodd = isodd(nfilt)   # boolean: "nodd" means filter length is odd
+    nfcns = numtaps ÷ 2   # integer divide
+    if nodd && !neg
+        nfcns = nfcns + 1
+    end
+    temp = (ngrid-1) / nfcns
+    dimsize = Int64(ceil(numtaps/2.0 + 2))
```
martinholters: `ceil(Int64, numtaps/2 + 2)`

src/Filters/remez_fir.jl
```diff
@@ -0,0 +1,825 @@
...
+    nodd = isodd(nfilt)
+    nfcns = nfilt ÷ 2
+    if nodd && !neg
+        nfcns = nfcns + 1
+    end
+
+    """/*
+     * SET UP THE DENSE GRID. THE NUMBER OF POINTS IN THE GRID
+     * IS (FILTER LENGTH + 1)*GRID DENSITY/2
+     */"""
```
martinholters: Julia  comment.

martinholters:
Concerning the short-circuit syntax: I definitely think it's more Julian than squeezing a whole `if` block in one line. In general, I personally like an `if` across multiple lines even better; but for the present code, that would be an unduly increase in lines of code.

martinholters:
I've spent some time poking at all those `@label`s and `@goto`s. For the big loop, I could eliminate some of them, but it's still a mess. But further down, I found this equivalent:
```diff
     # an already computed response on one of the extremals "l" -
     # if the extremal is equal to the frequency ft. If no y[l]
     # matches, a[j] is computed using freq_eval.
    for j = 1 : nfcns
         ft = (j - 1) * delf
         xt = cospi(2ft)
         if !full_grid
             xt = (xt-bb)/aa
             ft = acos(xt)/(2π)
         end
-      @label L410
-        xe = x[l]
-        if (xt > xe) @goto L420; end
-        if ((xe-xt) < fsh) @goto L415; end
-        l += 1
-        @goto L410
-      @label L415
-        a[j] = y[l]
-        @goto L425
-      @label L420
-        if ((xt-xe) < fsh) @goto L415; end
-        grid[1] = ft
-        a[j] = freq_eval(1,nz,grid,x,y,ad)
-      @label L425
+        while x[l]-xt >= fsh
+            l += 1
+        end
+        if xt-x[l] < fsh
+            a[j] = y[l]
+        else
+            grid[1] = ft
+            a[j] = freq_eval(1,nz,grid,x,y,ad)
+        end
         if (l > 1) l = l-1; end
     end
```
I definitely think the code is much more readable. It's also interesting that the comparison `xt > xe` (`xt > x[l]`) turned out to be redundant (after the transformation).

sirtom67:
Martin: thanks for round 2. I've incorporated all of your comments. Nice job re-writing that simple loop with no "gotos" - much better!

sirtom67:
I switched over all the single-line "if" statements to either short-circuit or 3 lines. I think keeping the line-for-line translation of the extremal index update loop is good, so those are all short-circuit.

Anything else?

martinholters:
These comments got me thinking (explaining `eff` and `wate`):
```
 *  AN ARBITRARY FUNCTION OF FREQUENCY CAN BE
 *  APPROXIMATED IF THE USER REPLACES THIS FUNCTION
 *  WITH THE APPROPRIATE CODE TO EVALUATE THE IDEAL
 *  MAGNITUDE.
```
```
 *  SIMILAR TO THE FUNCTION eff, THIS FUNCTION CAN
 *  BE REPLACED BY A USER-WRITTEN ROUTINE TO CALCULATE ANY
 *  DESIRED WEIGHTING FUNCTION.
```
Clearly, going into the source and editing these functions is not what we want users to do. Therefore, I wonder whether we should rework the interface to have the following:
* A "basic" interface which corresponds to the `filter_type=filter_type_bandpass` case.
* An "advanced" interface which takes functions `eff` and `wate` and a flag for the symmetry.

I haven't thought this through completely, just wanted to toss out the idea.

sirtom67:
Good idea. Is there a way to include a couple functions in a structure? We could have a structure called a "RemezSpecification" that would include all the existing specs, plus a couple functions that would default to eff and wate. Then the current "remez" function would be like it is interface-wise, but it would just be a wrapper: it would create a default spec based on the input arguments, and then pass that down to a new "remez_core" function. A common type of filter that needs a custom desired function is one that is a LPF but has a 1/sinc compensation - we can add one to the test set as an example.

Martin what do you think of merging this pull request in its current form, since with my plan a custom filter API will be backwards compatible with the existing remez function? I like the existing API since it is very similar to the API in scipy - I think that has value now and will continue to have value if we come up with something better.

src/Filters/remez_fir.jl
```diff
@@ -0,0 +1,829 @@
...
+#    Type III and IV negative symmetric linear phase: neg==1   (filter_type==hilbert or differentiator)
+@enum RemezFilterType filter_type_bandpass=1 filter_type_differentiator=2 filter_type_hilbert=3
+
+
+
+"""/*
+ *-----------------------------------------------------------------------
+ * FUNCTION: lagrange_interp (d)
+ *  FUNCTION TO CALCULATE THE LAGRANGE INTERPOLATION
+ *  COEFFICIENTS FOR USE IN THE FUNCTION gee.
```
martinholters: While this docstring may be of historical interest (?), it's pretty pointless in its current form. It doesn't explain the parameters, and it refers to a non-existing function `gee`. Also, the embedded C-style comment marks don't make too much sense when interpreted as markdown. Not sure what to do here though. Come up with a proper docstring? Convert into a Juila style comment?

src/Filters/remez_fir.jl
```diff
@@ -0,0 +1,829 @@
...
+function eff(freq::Float64, fx::AbstractVector, lband::Integer, filter_type::RemezFilterType)
+    if filter_type == filter_type_differentiator
+        return fx[lband] * freq
+    else 
+        return fx[lband]
+    end
+end
+
+"""
+wate(freq::Float64, fx::AbstractVector, wtx::AbstractVector, lband::Integer, filter_type::RemezFilterType)
```
martinholters: Not that it matters much, but technically, the signature here is missing the `wtx` parameter.

src/Filters/remez_fir.jl
```diff
@@ -0,0 +1,829 @@
...
+
+"""
+build_grid(numtaps, bands, desired, weight, grid_density)
+return "grid" and "des" and "wt" arrays
+"""
+function build_grid(numtaps, bands, desired, weight, grid_density, filter_type::RemezFilterType)
+    # translate from scipy remez argument names
+    L = numtaps
+    M = L ÷ 2    # integer divide (truncated)
+    grid_spacing = 0.5 / (grid_density * M)  # "delf" or "delta f"
```
martinholters: Unused? Then `L` and `M` could also be removed.

src/Filters/remez_fir.jl
```diff
@@ -0,0 +1,829 @@
...
+
+    lgrid = grid_density
+    dimsize = Int( ceil(numtaps/2.0 + 2) )
```
martinholters: `ceil(Int, ...)` as in `remez`

src/Filters/remez_fir.jl
```diff
@@ -0,0 +1,829 @@
...
+    while true
+        fup = edge[l + 1]
+        while true
+            temp = grid[j]
+            des[j] = eff(temp,fx,lband,filter_type)
+            wt[j] = wate(temp,fx,wtx,lband,filter_type)
+            j += 1
+            if j > wrksize
+                # too many points, or too dense grid
+                return -1
```
martinholters: If this is ever hit, it will give a rather obscure error message because this case is not covered in `remez`. Should probably throw an error from here, though I'm not sure under what circumstances this can happen.

src/Filters/remez_fir.jl
```diff
@@ -0,0 +1,829 @@
...
+using PyPlot
+b = DSP.Filters.PolynomialRatio(bpass, [1.0])
+b2 = DSP.Filters.PolynomialRatio(bpass2, [1.0])
+f = linspace(0, 0.5, 1000)
+plot(f, 20*log10.(abs.(freqz(b,f,1.0))))
+plot(f, 20*log10.(abs.(freqz(b2,f,1.0))))
+grid()
+```
+"""
+
```
martinholters: There should be no blank lines between docstring and function in Juila 0.7.

src/Filters/remez_fir.jl
```diff
@@ -0,0 +1,829 @@
...
+=========#
+
+function remez(numtaps::Integer, bands::Vector, desired::Vector; 
+               weight::Vector=[], 
+               Hz::Real=1.0, 
+               filter_type::RemezFilterType=filter_type_bandpass,
+               maxiter::Integer=25, 
+               grid_density::Integer=16)
+    if length(weight)==0
+        weight = ones(desired)
```
martinholters: `fill(1.0, length(desired))` (`ones` is deprecated in Julia 0.7).

src/Filters/remez_fir.jl
```diff
@@ -0,0 +1,829 @@
...
+    end
+    bands = copy(bands)/Hz
```
martinholters: No need for the `copy` here.

src/Filters/remez_fir.jl
```diff
@@ -0,0 +1,829 @@
...
+    while true
+    
+        #
+        # Start next iteration
+        #        
+      @label L100
+        iext[nzz] = ngrid + 1
+        niter += 1
+        if niter > maxiter
+            warn("remez() iteration count exceeds maxiter = $maxiter, filter is not converged; try increasing maxiter")
```
martinholters: Use `Compat.@warn`

src/Filters/remez_fir.jl
```diff
@@ -0,0 +1,829 @@
...
+        dden = 0.0
+        k = 1
+        for j = 1 : nz
+            l = iext[j]
+            dnum += ad[j] * des[l]
+            dden += k * ad[j] / wt[l]
+            k = -k
+        end
+        dev = dnum / dden
+        #printfmtln("  DEVIATION = {} at niter {}", dev, niter)
```
martinholters: Remove.

src/Filters/remez_fir.jl
```diff
@@ -0,0 +1,829 @@
...
+
+        fill!(y, 0.0)
+        nu, dev = initialize_y(dev, nz, iext, des, wt, y)
+        
+        if dev <= devl
+            # finished
+            #printfmtln("dev {} <= devl {}, throwing", dev, devl)
```
martinholters: Remove.

src/Filters/remez_fir.jl
```diff
@@ -0,0 +1,829 @@
...
+        j == nzz && (ynz = comp)   # equivalent to "if (j == nzz) ynz = comp; end"
+        j >= nzz && @goto L300
+        kup = iext[j+1]
+        l = iext[j]+1
+        nut = -nut
+        j == 2 && (y1 = comp)
+        comp = dev
+        l >= kup && @goto L220
+        err = (freq_eval(l,nz,grid,x,y,ad)-des[l]) * wt[l]
+        (nut*err-comp) <= 0.0 && @goto L220
```
martinholters: I'd prefer `nut*err <= comp` for the condition here (and for the following occurrences).

src/Filters/remez_fir.jl
```diff
@@ -0,0 +1,829 @@
...
+        @goto L100
+      @label L370
+        
+        
+        if jchnge <= 0  # we are done if none of the extremal indices changed
+            break
+        end
+    end  # while
+
+    # printfmtln("Iterations complete, niter = {}", niter)
```
martinholters: Remove.

src/Filters/remez_fir.jl
```diff
@@ -0,0 +1,829 @@
...
+ *  Email:
+ *      egil@kvaleberg.no
+ *  Web:
+ *      http://www.kvaleberg.com/
+ *
+ *********************************************************/
+
+=============================================#
+
+#using Formatting
```
martinholters: Remove.

src/Filters/remez_fir.jl
martinholters: Could instead of this `if`-block  just use `fill(1.0, length(desired))` as the default value in function signature.

martinholters:
> Ive started cleaning it up and may likely propose the changed version in the next couple of days

Opened a PR into this PRs branch at https://github.com/sirtom67/DSP.jl/pull/1

src/Filters/remez_fir.jl
```diff
@@ -0,0 +1,797 @@
...
+  ARE A DENSE GRID WHICH REPLACES THE FREQUENCY AXIS, THE
+  DESIRED FUNCTION ON THIS GRID, THE WEIGHT FUNCTION ON THE
+  GRID, THE NUMBER OF COSINES, AND AN INITIAL GUESS OF THE
+  EXTREMAL FREQUENCIES.  THE PROGRAM MINIMIZES THE CHEBYSHEV
+  ERROR BY DETERMINING THE BSMINEST LOCATION OF THE EXTREMAL
+  FREQUENCIES (POINTS OF MAXIMUM ERROR) AND THEN CALCULATES
+  THE COEFFICIENTS OF THE BEST APPROXIMATION.
+-----------------------------------------------------------------------
+=========#
+function remez(numtaps::Integer, bands::Vector, desired::Vector;
```
martinholters: Julia 0.7 still complains:
```
┌ Warning: Deprecated syntax `multiple line breaks between doc string and object` at /home/travis/build/JuliaDSP/DSP.jl/src/Filters/remez_fir.jl:373.
│ Use `at most one line break` instead.
└ @ ~/build/JuliaDSP/DSP.jl/src/Filters/remez_fir.jl:373
```
Can you move the comment block with the C code banner above the docstring?

src/Filters/remez_fir.jl
```diff
@@ -0,0 +1,797 @@
...
+               Hz::Real=1.0, 
+               filter_type::RemezFilterType=filter_type_bandpass,
+               maxiter::Integer=25, 
+               grid_density::Integer=16)
+    bands = bands/Hz
+
+    # Sanity checks on arguments
+    @assert issorted(bands)
+    @assert length(bands) == 2length(desired) == 2length(weight)
+    @assert (0 <= bands[1]) && (bands[end]/2 <= 0.5)
```
martinholters: These should throw `ArgumentError`s if the conditions are violated. E.g.
```julia
issorted(bands) || throw(ArgumentError("`bands` is not monotonically increasing"))
```

src/Filters/remez_fir.jl
```diff
@@ -0,0 +1,797 @@
...
+        klow = l - 1
+        jchnge += 1
+        @goto L200
+
+      @label L220
+        l -= 1
+      @label L225
+        l -= 1; l <= klow && @goto L250
+        err = (freq_eval(l,nz,grid,x,y,ad)-des[l]) * wt[l]
+        (nut*err-comp) > 0.0 && @goto L230
```
martinholters: Also `nut*err > comp`

sirtom67:
Martin - thanks for all your contributions on this new tool - it is so much improved! 

Did you know I am giving a talk about remez and our efforts at porting it to Julia? The talk is at JuliaCon in London exactly two weeks from today! One of the things I will show is a comparison in run-time between Julia and the C version - here's a preview. C is about 10x faster for extremely short (length 11) filters, but for reasonable filter lengths (length 101) it is only 2x and for very long filters, C is no faster.
remez_time_julia_c_compare

test/remez_fir.jl
```diff
@@ -0,0 +1,140 @@
...
+#
+@testset "remez_180_bpf" begin
+    h = remez(180, [0, 0.375, 0.4, 0.5, 0.525, 1.0], [0.0, 1.0, 0.0]; Hz=2.0, maxiter=30);
+    h_scipy = readdlm(joinpath(dirname(@__FILE__), "data", "remez_180_bpf.txt"),'\t')
+    @test h ≈ h_scipy
+end
+
+@testset "remez_warn_no_converge_after_maxiter_iterations" begin
+    warn_check(msg) = contains(msg, "filter is not converged")
+    @test_warn warn_check remez(180, [0, 0.375, 0.4, 0.5, 0.525, 1.0], [0.0, 1.0, 0.0]; Hz=2.0)
```
martinholters: Good idea to test this, but doesn't work on 0.7 as is. This is quite tricky, I think it should be something like
```julia
@static if VERSION ≥ v"0.7.0-DEV.2988"
    @test_logs (:warn, r"filter is not converged") remez(180, [0, 0.375, 0.4, 0.5, 0.525, 1.0], [0.0, 1.0, 0.0]; Hz=2.0)
else
    warn_check(msg) = contains(msg, "filter is not converged")
    @test_warn warn_check remez(180, [0, 0.375, 0.4, 0.5, 0.525, 1.0], [0.0, 1.0, 0.0]; Hz=2.0)
end
```

martinholters:
> Did you know I am giving a talk about remez and our efforts at porting it to Julia? The talk is at JuliaCon in London exactly two weeks from today!

I didn't. That's great! I won't make it to JuliaCon, but I'll surely watch the video once it's online.

I'll give this a final (hopefully!) look tomorrow and merge then (unless anyone objects). It's a bit unfortunate it crashes on Julia 0.7-beta2, but fingers crossed for an -rc1 ready for JuliaCon :smile: 

Apropos performance: I've noticed that the current code evaluates `cospi(2grid[k])` for different `k` over and over again (in `freq_eval`). Might be worth pre-computing that once. But that would be "cheating" in the comparison with C. But that may well be a future optimization.

martinholters:
Thanks again!

Issue #213, opened 2018-07-25
FIR filtering fails with bounds error when using large vectors


galenlynch:
Attempting to filter large vectors produces a `BoundsError`:

```julia
julia> nx = 350088492;
julia> x = rand(nx);
julia> nb = 301;
julia> b  = rand(nb);
julia> filt(b, x)
ERROR: BoundsError: attempt to access 28-element Array{Int64,1} at index [29]
Stacktrace:
 [1] optimalfftfiltlength(::Int64, ::Int64) at /home/glynch/.julia/v0.6/DSP/src/Filters/filt.jl:485
 [2] filt(::Array{Float64,1}, ::Array{Float64,1}) at /home/glynch/.julia/v0.6/DSP/src/Filters/filt.jl:572
```

This happens whenever the array is larger than 2^28, which is the last pre-computed power of two used by the heuristic lookup table `optimalfftfiltlength`:
```julia
julia> DSP.Filters.optimalfftfiltlength(nb, nx)
ERROR: BoundsError: attempt to access 28-element Array{Int64,1} at index [29]
Stacktrace:
 [1] optimalfftfiltlength(::Int64, ::Int64) at /home/glynch/.julia/v0.6/DSP/src/Filters/filt.jl:485
```

I would like to fix this problem, but I'm struggling to understand the logic of `optimalfftfiltlength`.

ssfrr:
fixed in #215

PR #214, opened 2018-07-25
Fix esprit tests so they don't randomly fail

galenlynch:
After esprit was added to DSP.jl, testing the package randomly fails. This can
be fixed by seeding the random number generator prior to testing esprit.

PR #215, opened 2018-07-25
Prevent bounds error when filtering large arrays

galenlynch:
As described in #213, FIR filtering will fail if the input vector is larger than
`2^28`, which is the last saved benchmark used by `optimalfftfiltlength` to
heuristically choose a chunk size `L` for the overlap-save algorithm used in `fftfilt`.

Instead of relying on saved benchmarks to pick `L`, it is possible to calculate
the per-sample complexity of the overlap-save algorithm, as a function of `nfft`
and by extension `L` [0] (the result is the same for overlap-save as the overlap-add calculation shown here). I have replaced looking up benchmarks with calls to a
new function, `os_fft_complexity`, which calculates the per-sample complexity for
a given `nfft` that is a power of 2 and the filter size `nb`. I then choose the
minimum complexity for a range of `nfft`s, and use that as the best `nfft`.

[0] e.g. https://www.allaboutcircuits.com/technical-articles/overlap-add-method-linear-filtering-based-on-the-discrete-fourier-transform/

### DIFF
--- a/src/Filters/Filters.jl
+++ b/src/Filters/Filters.jl
@@ -3,7 +3,7 @@ using ..DSP: @importffts, mul!, rmul!
 using ..Unwrap
 using Polynomials, ..Util
 import Base: *
-using Compat: copyto!, undef
+using Compat: copyto!, undef, argmin
 import Compat
 using Compat.LinearAlgebra: I
 using Compat.Statistics: middle
--- a/src/Filters/filt.jl
+++ b/src/Filters/filt.jl
@@ -484,40 +484,28 @@ filt(h::AbstractArray, x::AbstractArray) =
 # fftfilt and filt
 #
 
-const FFT_LENGTHS = 2 .^ (1:28)
-# FFT times computed on a Core i7-3930K @4.4GHz
-# The real time doesn't matter, just the relative difference
-const FFT_TIMES = [6.36383e-7, 6.3779e-7 , 6.52212e-7, 6.65282e-7, 7.12794e-7, 7.63172e-7,
-                   7.91914e-7, 1.02289e-6, 1.37939e-6, 2.10868e-6, 4.04436e-6, 9.12889e-6,
-                   2.32142e-5, 4.95576e-5, 0.000124927, 0.000247771, 0.000608867, 0.00153119,
-                   0.00359037, 0.0110568, 0.0310893, 0.065813, 0.143516, 0.465745, 0.978072,
-                   2.04371, 4.06017, 8.77769]
+# Number of real operations required for overlap-save with nfft = 2^pow2 and filter
+# length nb
+os_fft_complexity(pow2, nb) = 4 * (2 ^ pow2 * (pow2 + 1)) / (2 ^ pow2 - nb + 1)
 
 # Determine optimal length of the FFT for fftfilt
 function optimalfftfiltlength(nb, nx)
-    nfft = 0
-    if nb > FFT_LENGTHS[end] || nb >= nx
-        nfft = nextfastfft(nx+nb-1)
-    else
-        fastestestimate = Inf
-        firsti = max(1, searchsortedfirst(FFT_LENGTHS, nb))
-        lasti = max(1, searchsortedfirst(FFT_LENGTHS, nx+nb-1))
-        L = 0
-        for i = firsti:lasti
-            curL = FFT_LENGTHS[i] - (nb - 1)
-            estimate = ceil(Int, nx/curL)*FFT_TIMES[i]
-            if estimate < fastestestimate
-                nfft = FFT_LENGTHS[i]
-                fastestestimate = estimate
-                L = curL
-            end
-        end
+    first_pow2 = ceil(Int, log2(nb))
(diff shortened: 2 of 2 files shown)


galenlynch:
I don't understand why this fails on 0.7, it works on my machine with 0.7. Whatever the reason, the failure seem unrelated to my changes.

galenlynch:
Looking at the code in `test/lpc.jl`, it seems like another case (c.f. #214) of testing with uncontrolled random inputs leading to sporadic failures. 

I think it might be a good idea to seed the random number generator for all tests, instead of getting spurious failures when testing.

galenlynch:
Now that I look at it, I don't think I agree with the heuristics in `filt` to decide between fft filtering and direct form filtering... shouldn't the primary consideration be the size of b?

Some testing shows that the current heuristics may be off. The following code is a little confusing since I'm using functions in PR #206 to eliminate any timing differences due to output allocation. However, even for problems that are normally direct-form filtered due to the first heuristic of `filtops` being smaller than 500000, DSP.jl is actually making the wrong choice by not doing fft filtering:

```julia
julia> nx = 10000;
julia> x = rand(nx);
julia> nb = 301;
julia> b = rand(nb);
julia> out = similar(x);
julia> @benchmark DSP.Filters._fftfilt!(out, b, x, 4096)
BenchmarkTools.Trial: 
  memory estimate:  103.36 KiB
  allocs estimate:  128
  --------------
  minimum time:     145.195 μs (0.00% GC)
  median time:      152.081 μs (0.00% GC)
  mean time:        163.406 μs (2.21% GC)
  maximum time:     1.975 ms (44.39% GC)
  --------------
  samples:          10000
  evals/sample:     1

julia> @benchmark DSP.Filters._tdfilt!(out, b, x)
BenchmarkTools.Trial: 
  memory estimate:  2.50 KiB
  allocs estimate:  1
  --------------
  minimum time:     466.888 μs (0.00% GC)
  median time:      469.177 μs (0.00% GC)
  mean time:        487.608 μs (0.00% GC)
  maximum time:     797.892 μs (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     1

```

Even when the number of ops saved per sample by using fft filtering is much smaller (e.g. when nb = 51 and nx = 10000, which has a margin of 6.7 ops per sample), the overhead of fft planning seems to be outweighed by the ops savings:

```julia
julia> nx = 10000;
julia> x = rand(nx);
julia> nb = 51;
julia> b = rand(nb);
julia> out = similar(x);
julia> @benchmark DSP.Filters._tdfilt!(out, b, x)
BenchmarkTools.Trial: 
  memory estimate:  544 bytes
  allocs estimate:  1
  --------------
  minimum time:     149.406 μs (0.00% GC)
  median time:      151.479 μs (0.00% GC)
  mean time:        157.488 μs (0.00% GC)
  maximum time:     375.683 μs (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     1

julia> @benchmark DSP.Filters._fftfilt!(out, b, x, 512)
BenchmarkTools.Trial: 
  memory estimate:  19.47 KiB
  allocs estimate:  123
  --------------
  minimum time:     124.505 μs (0.00% GC)
  median time:      129.221 μs (0.00% GC)
  mean time:        138.933 μs (0.65% GC)
  maximum time:     8.926 ms (26.63% GC)
  --------------
  samples:          10000
  evals/sample:     1
```

I don't really know how to calculate the true ops overhead of planning the `rfft` and `brfft`, but from these tests it seems like the overhead may not even be worth considering. Maybe it would be better to just compare the number of ops per sample using direct form filtering and fft filtering, and choose the algorithm solely based on that? At the very least, the current heuristics seem wrong.

galenlynch:
Should I split the commits pertaining to the original problem away from the commits changing when fft filtering is chosen over time domain filtering off in this PR? 

The original issue is quite troublesome: filtering simply does not work for large arrays. I think this should be fixed somehow, even if it's not the way that I suggest in this PR.

ssfrr:
IMO these things are coupled enough that it's not worth the trouble of splitting into separate PRs.

I haven't looked in detail yet but it looks like the new implementation of `optimalfftfiltlength` should be pretty fast, and given that it's more general than the previous table lookup approach it seems like a good idea.

Regarding whether we can ignore the planning overhead when choosing between time-domain and fft filtering, it might be worth benchmarking the planning itself to get a sense for how much of the total filtering time it makes up (particularly for small filter operations). Perhaps the right approach is to have a test that goes through different filter lengths and makes sure that the heuristic lines up with which method is actually faster (by explicitly running in the time domain and fft).

galenlynch:
@ssfrr Thanks for the comments! In my testing (see above) there doesn't seem to be any noticeable planning overhead, so for now I think it makes sense to just choose based on algorithmic complexity.

Just as a note, the tests seem to have failed due to CI stalling.

galenlynch:
Gentle bump, would love to get these changes in master.

galenlynch:
Any thoughts? Happy to close the PR if needed.

galenlynch:
bump.

PR #216, opened 2018-07-26
Workaround JuliaLang/julia#28077

martinholters:
Place a no-op (`nothing`) immediately after labels which would otherwise be followed by a phi-node. Looks silly but prevents Julia 0.7-beta2 from crashing on the remez tests due to JuliaLang/julia#28077 (and should generally make `remez` work correctly on beta2).

CC @sirtom67

### DIFF
--- a/src/Filters/remez_fir.jl
+++ b/src/Filters/remez_fir.jl
@@ -84,6 +84,12 @@ C CODE BANNER
  *********************************************************/
 
 =============================================#
+if v"0.7.0-DEV.5103" ≤ VERSION < v"0.7.0-beta2.22"
+    # work around JuliaLang/julia#28077 by inserting a no-op (`nothing`) after each label
+    macro label(sym)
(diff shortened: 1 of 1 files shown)


sirtom67:
Nice job finding a work-around. 
I agree it looks silly. Would the same fix work with the "nothing" on a single line, e.g.
@label L310; nothing
I find that slightly less silly looking. :-)

martinholters:
Changed to define a custom `@label` macro on affected Julia versions which inserts the `nothing`.

martinholters:
Barring objections (or CI failures), I'll merge this later today to get tests passing on 0.7 again.

martinholters:
Got one
```
LPCBurg(): Test Failed at /home/travis/build/JuliaDSP/DSP.jl/test/lpc.jl:9
  Expression: all(abs.([1; ar] .- coeffs) .<= 0.01)
```
which happens occasionally. (I definitely do see this one more often than the esprit test, @galenlynch).

PR #217, opened 2018-07-26
Fix #211

### LINKED ISSUE #211: resampling/filtering don't accept integer resampling rates
The resampling rate arguments for `FIRFilter` and `resample` accept `Rational` and `AbstractFloat`, but not integers.

```julia
julia> x = rand(8)
8-element Array{Float64,1}:
 0.179779
...

julia> resample(x, 2)
ERROR: MethodError: no method matching resample_filter(::Int64)
Closest candidates are:
  resample_filter(::AbstractFloat) at C:\Users\sfr\.julia\v0.6\DSP\src\Filters\design.jl:594
  resample_filter(::AbstractFloat, ::Any) at C:\Users\sfr\.julia\v0.6\DSP\src\Filters\design.jl:594
  resample_filter(::AbstractFloat, ::Any, ::Any) at C:\Users\sfr\.julia\v0.6\DSP\src\Filters\design.
jl:594
  ...
Stacktrace:
 [1] resample(::Array{Float64,1}, ::Int64) at C:\Users\sfr\.julia\v0.6\DSP\src\Filters\stream_filt.j
l:706
 [2] macro expansion at C:\Users\sfr\.julia\v0.6\Atom\src\repl.jl:118 [inlined]
 [3] anonymous at .\<missing>:?

julia> resample(x, 2//1)
15-element Array{Float64,1}:
 0.179762
...
```

jordancluts:
Intended to fix #211 by simply changing the accepted type signature where needed.

### DIFF
--- a/src/Filters/stream_filt.jl
+++ b/src/Filters/stream_filt.jl
@@ -138,7 +138,7 @@ Construct a stateful FIRFilter object from the vector of filter taps `h`.
 the input to output sample rate relationship (e.g. `147//160` for
 converting recorded audio from 48 KHz to 44.1 KHz).
 """
-function FIRFilter(h::Vector, resampleRatio::Union{Integer,Rational} = 1//1)
+function FIRFilter(h::Vector, resampleRatio::Union{Integer,Rational} = 1)
     interpolation = numerator(resampleRatio)
     decimation    = denominator(resampleRatio)
     historyLen    = 0
@@ -310,7 +310,7 @@ function outputlength(kernel::FIRStandard, inputlength::Integer)
 end
 
 function outputlength(kernel::FIRInterpolator, inputlength::Integer)
-    outputlength(inputlength-kernel.inputDeficit+1, kernel.interpolation//1, kernel.ϕIdx)
+    outputlength(inputlength-kernel.inputDeficit+1, kernel.interpolation, kernel.ϕIdx)
 end
 
 function outputlength(kernel::FIRDecimator, inputlength::Integer)
@@ -347,7 +347,7 @@ function inputlength(kernel::FIRStandard, outputlength::Integer)
 end
 
 function inputlength(kernel::FIRInterpolator, outputlength::Integer)
-    inLen = inputlength(outputlength, kernel.interpolation//1, kernel.ϕIdx)
+    inLen = inputlength(outputlength, kernel.interpolation, kernel.ϕIdx)
     inLen += kernel.inputDeficit - 1
 end
 
--- a/src/Filters/stream_filt.jl
+++ b/src/Filters/stream_filt.jl
@@ -297,7 +297,7 @@ end
 # ( It's hard to explain how this works without a diagram )
 #
 
-function outputlength(inputlength::Integer, ratio::Rational, initialϕ::Integer)
+function outputlength(inputlength::Integer, ratio::Union{Integer,Rational}, initialϕ::Integer)
     interpolation = numerator(ratio)
     decimation    = denominator(ratio)
     outLen        = ((inputlength * interpolation) - initialϕ + 1) / decimation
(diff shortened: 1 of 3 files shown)


jordancluts:
I believe the failed test will be fixed by #216

src/Filters/stream_filt.jl
```diff
@@ -58,7 +58,7 @@ end
 # Rational resampler FIR kernel
 mutable struct FIRRational{T}  <: FIRKernel{T}
     pfb::PFB{T}
-    ratio::Rational{Int}
```
martinholters: Moving away from a concrete type here might have a performance impact, though I haven't checked where `ratio` is used/how large the impact is. Alternatively, we could leave this as `Rational{Int}` and convert upon construction, no?

martinholters:
Can you add tests exercising the integer case?

jordancluts:
I'll give it a try, I'm new and haven't tried doing anything with testing before but it should be easy enough to crib off of the test_rational in filt_stream.jl

jordancluts:
Could someone take a look at the latest rendition here?

test/filt_stream.jl
```diff
@@ -354,4 +408,7 @@ end
             test_arbitrary(Th, x, convert(Float64, ratio)+rand(), 32)
         end
     end
+    if numerator(ratio) == interpolation && denominator(ratio) == decimation && numerator(ratio) != 1 && denominator(ratio) == 1
```
ssfrr: Could this just be `if decimation == 1` ?

test/filt_stream.jl
```diff
@@ -275,6 +275,60 @@ function test_rational(h, x, ratio)
 end
 
 
+#
+# Integer resampling
+#
+
+function test_integer(h, x, ratio)
```
ssfrr: It seems like the `test_xxx` functions are almost identical and could be combined into a single `test_resampling` function. Having a bunch of these almost-identical functions makes it hard to see what's actually different about the tests. This isn't strictly your problem because there's already quite a bit of duplication before this PR, but perhaps this is a good opportunity to try to refactor and clean up this testing. Would you be willing to take a crack at it?

I'd be happy to drop the big ASCII-art headings if that makes it easier to combine the functions. I don't know if anyone else is attached to them - maybe @simonster?

test/filt_stream.jl
jordancluts: I believe you are right here. I couldn't figure out what the `numerator(ratio)==interpolation && denominator(ratio == decimation` was for in the test_rational if statement so I left them in case I was simply not understanding the purpose.

test/filt_stream.jl
jordancluts: It'd be very easy to combine the rational resampling and integer resampling (the rational resampling needs I think 1 change to run the integer tests). I had separated them because that seemed to be how the rest of the test functions were structured. I will edit this PR to reduce the additions. Refactoring all the tests is a bit beyond me. I'm trying to get my feet wet with my first PR here and I don't really understand all the testing machinery. I spent awhile looking at the tests and they have so much similarity there has to be was to reduce code duplication but I'm not sure how to go about it without obfuscating the process further or reducing code coverage. I also have a hard time prototyping a change as I'm not even sure how tests are run locally so I've been sort of manually testing the tests as it were.

ssfrr: Yeah, that's reasonable. It's always a tough judgment call whether to duplicate for consistency, but the new version seems like a good balance - at least it's not making the duplication worse. :smile: 

I'll take a look tonight at the latest changes. I'll also see if I can figure out what's going on with all this `@timeifinteractive` stuff (it predates me working on this package). I usually just run the tests with `julia test/runtests.jl`.

Thanks for the contribution!

ssfrr:
Looks good, I was going to make some minor tweaks (just removing some now-unnecessary `x//1`s that were converting ints to rationals), but AFAICT I don't have permission to push to this branch. Would you mind clicking the "allow maintainers to make changes to this PR" checkbox?

jordancluts:
The "Allow edits from maintainers" box is already checked here. Is there somewhere else I need to grant permission?

ssfrr:
Figured it out. It was trying to push to the wrong branch name on your fork. Once CI passes I'll merge.

ssfrr:
merged! Thanks for the contribution.

PR #218, opened 2018-07-28
Normalize filter before loop in fftfilt

galenlynch:
This normalization only needs to be done once, but is instead done every
iteration.

### DIFF
--- a/src/Filters/filt.jl
+++ b/src/Filters/filt.jl
@@ -561,6 +561,7 @@ function _fftfilt!(out::AbstractArray{T}, b::AbstractVector{T}, x::AbstractArray
     filterft = similar(tmp2)
     copyto!(tmp1, b)
     tmp1[nb+1:end] .= zero(T)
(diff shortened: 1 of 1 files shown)


src/Filters/filt.jl
```diff
@@ -561,6 +561,7 @@ function _fftfilt!(out::AbstractArray{T}, b::AbstractVector{T}, x::AbstractArray
     filterft = similar(tmp2)
     copyto!(tmp1, b)
     tmp1[nb+1:end] .= zero(T)
+    broadcast!(*, tmp1, tmp1, normfactor)
```
martinholters: Couldn't this be combined with the `copyto!` above to `tmp1[1:nb] .= b .* normfactor`?

galenlynch: Just made that change and pushed to my fork, though for some reason it's not showing up here.

galenlynch:
bump

PR #221, opened 2018-08-12
fix for 1.0 julia, as prior, changed to still work on 0.6 version

wherrera10:


dlfivefifty:
It looks like the tests are only failing on nightly, so maybe nightly should be an allowed failure? 

Looking for to this being merged and a new release tagged ☺️

ssfrr:
Looks like a networking issue so I restarted the CI run. I think we want to support 0.6, 0.7, and 1.0, but nightly can be an allowed failure. I'm not sure how much current nightly is diverging from the 1.0 release.

dlfivefifty:
Currently nightly is 1.1- which should support all 1.0 code, so if the build only fails on nightly it’s a bug in nightly, not in the PR.

ssfrr:
It's currently not being tested on 1.0 though, just 0.7. I think I'll be able to look into this later today and get it reviewed/merged/tagged.

ssfrr:
all good, passing on 0.6, 0.7, 1.0, and nightly. thanks!

PR #223, opened 2018-08-24
Some clean-up of remez

martinholters:
Start of an attempt to make this more readable. I found it quite hard to keep track of what all those variables mean, so I've started eliminating some of them.

Also, there is one algorithmic change, basically doing `grid .= cospi.(2.*grid)`. Before, the cosine might have potentially been applied only to a subset of grid points, but OTOH potentially multiple times. Net performance impact seems small (but an improvement), but this allows to remove some `cospi` calls, making for slightly less code.

What does have a significant performance impact is fixing a type-instabilty in `build_grid`:
```julia
# master
julia> @btime remez(20, $[0, 0.475, 0.5, 1.0], $[1.0, 0.0]; Hz=2.0);
  283.074 μs (8114 allocations: 142.48 KiB)

julia> @btime remez(151, $[0, 0.475, 0.5, 1.0], $[1.0, 0.0]; Hz=2.0);
  4.846 ms (87515 allocations: 1.41 MiB)
```
```julia
# this PR
julia> @btime remez(20, $[0, 0.475, 0.5, 1.0], $[1.0, 0.0]; Hz=2.0);
  53.988 μs (88 allocations: 11.80 KiB)

julia> @btime remez(151, $[0, 0.475, 0.5, 1.0], $[1.0, 0.0]; Hz=2.0);
  2.803 ms (104 allocations: 60.39 KiB)
```
@sirtom67 have you done the benchmarks shown at JuliaCon with the version from master? That might explain (part of) the overhead for small filter sizes.

### DIFF
--- a/src/Filters/remez_fir.jl
+++ b/src/Filters/remez_fir.jl
@@ -408,8 +408,7 @@ function remez(numtaps::Integer, bands::Vector, desired::Vector;
         nfcns = nfcns + 1
     end
     temp = (ngrid-1) / nfcns
-    dimsize = ceil(Int64, numtaps/2 + 2)
-    
+
     #
     # SET UP A NEW APPROXIMATION PROBLEM WHICH IS EQUIVALENT
     # TO THE ORIGINAL PROBLEM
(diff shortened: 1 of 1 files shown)


sirtom67:
Martin: Nice work! I will regenerate the timing plot with your fix. A
couple Julia experts commented at the conference that the slow speed for
small filter sizes was probably due to type instability. I had gotten
code_typewarn to run and and it found a number of things, but i have not
fixed or measured anything.

sirtom67:
I know this is merged already but I finally got around to pulling the latest on master and re-running my timing comparison script. Here's the plot:
remez_time_julia_c_compare
As you can see the new version is much faster for small filters, and even faster than C by a bit for most orders.

sirtom67:
Here's the old timing, for reference.
remez_time_julia_c_compare-180729

PR #229, opened 2018-09-11
Update examples in remez

ClemensFMN:
In my Julia 1.0.0 installation, a 1d column vector is required in the input vector arguments. In addition, linspace is no longer available...


sirtom67:
This also addresses the issue #233 that I created yesterday. Looks good to me, thanks for the contribution!

PR #230, opened 2018-09-14
RFC/WIP: Use functions in remez API

martinholters:
Ref. https://github.com/JuliaDSP/DSP.jl/pull/210#issuecomment-406966472

I actually needed this functionality just the other day. WIP as I'm not quite sure about the API, yet. (Hence, also no tests or docs, yet.) Presently, this allows `desired` and `weight` to contain functions from frequency to desired gain or weight, respectively. I wonder whether we should instead allow an API that looked somewhat like this:
```julia
remez(152, [0..0.475 => 1,
            0.5..1.0 => (desired=0.0, weight=2.0)],
      Hz=2.0)
```
instead of
```julia
remez(152, [0, 0.475, 0.5, 1.0], [1.0, 0.0]; weight=[1,2], Hz=2.0)
```
where the RHS of `=>` could be a single number for desired gain (with weight defaulting to 1) or a (named) tuple, or a function/(named) tuple of functions (or a function returning a tuple?). Opinions?

### DIFF
--- a/src/Filters/Filters.jl
+++ b/src/Filters/Filters.jl
@@ -3,7 +3,7 @@ using ..DSP: @importffts, mul!, rmul!
 using ..Unwrap
 using Polynomials, ..Util
 import Base: *
-using Compat: copyto!, undef
+using Compat: copyto!, Nothing, undef
 import Compat
 using Compat.LinearAlgebra: I
 using Compat.Statistics: middle
--- a/src/Filters/remez_fir.jl
+++ b/src/Filters/remez_fir.jl
@@ -118,67 +118,16 @@ end
 
 
 """
-eff(freq::Float64, fx::AbstractVector, lband::Integer, filter_type::RemezFilterType)
-
-/*
- *-----------------------------------------------------------------------
- * FUNCTION: eff
- *  FUNCTION TO CALCULATE THE DESIRED MAGNITUDE RESPONSE
- *  AS A FUNCTION OF FREQUENCY.
- *  AN ARBITRARY FUNCTION OF FREQUENCY CAN BE
- *  APPROXIMATED IF THE USER REPLACES THIS FUNCTION
- *  WITH THE APPROPRIATE CODE TO EVALUATE THE IDEAL
- *  MAGNITUDE.  NOTE THAT THE PARAMETER FREQ IS THE
- *  VALUE OF NORMALIZED FREQUENCY NEEDED FOR EVALUATION.
- *-----------------------------------------------------------------------
- */
-"""
-function eff(freq::Float64, fx::AbstractVector, lband::Integer, filter_type::RemezFilterType)
-    if filter_type == filter_type_differentiator
-        return fx[lband] * freq
-    else 
-        return fx[lband]
-    end
-end
-
-"""
-wate(freq::Float64, fx::AbstractVector, wtx::AbstractVector, lband::Integer, filter_type::RemezFilterType)
-/*
- *-----------------------------------------------------------------------
- * FUNCTION: wate
- *  FUNCTION TO CALCULATE THE WEIGHT FUNCTION AS A FUNCTION
- *  OF FREQUENCY.  SIMILAR TO THE FUNCTION eff, THIS FUNCTION CAN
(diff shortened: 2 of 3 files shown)


martinholters:
CC @sirtom67 
Any opinion, anyone?

sirtom67:
Will review later today.

martinholters:
First, to be clear about this: I absolutely mean to keep the current API, and provide another one in addition to it. (At least for the foreseeable future.)

IMO the new API should
* _not_ take the filter type as an argument
* take the symmetry (odd/even) as a flag 
* take a single list (probably a `Vector` or possibly a `Dict`) with a mapping from band frequencies to desired gain and weight, where the latter defaults to 1. Both desired gain and weight may be given as a single number or as function of frequency.

The last point leaves some possibilities to explore. The band frequencies could reasonably be given as a two-tuple (`(lower, upper)`) or as an interval using [IntervalSets.jl](https://github.com/JuliaMath/IntervalSets.jl) (`lower..upper`) as used in the example above. The latter looks nicer, but introduces a new dependency. The mapping part is best expressed using the `=>` operator. The desired gain/weight can be stuffed in a two-tuple or (on Julia 0.7 and up) in a named tuple. If the weight is not specified, the wrapping (named) tuple could be omitted.

sirtom67:
OK, cool, keeping the current API is good.

Holy cow, Julia is so cool. Thanks for pointing out the IntervalSets.jl package - that is what I was missing. That would be a new dependency of DSP.jl, right?

Also, => is the operator for creating a pair.

So your new API makes sense to me:
It is a list of pairs.
Each pair specifies a closed interval (one of two-tuple of numbers, or an interval from IntervalSets), and the desired function / weight designator. The desired function / weight designator must be one of:
- a number, in which case the weight is 1
- a tuple of numbers
- a function, in which case the weight is 1
- a tuple of functions
- a tuple of (function, number) or (number, function)

This sounds good to me.

Are you proposing a 2nd function that then calls remez_fir, or would the remez_fir function arguments be modified to handle both the new and old APIs?

sirtom67:
Martin: Guess what! I had to design an inverse sinc filter yesterday for work.
I'd like to add this to our tests and doc. But we need to finalize the API first.
inverse_sinc_response
```
using DSP
using PyPlot

L = 64

Fs = 4800*L
f = range(0, stop=0.5, length=10000)

P = (π*f*Fs/4800) ./ sin.(π*f*Fs/4800)
Pdb = 20*log10.(abs.(P))

g_vec = remez(201, [0.0;2880;10000;Fs/2], 
                   [f-> (f==0) ? 1.0 : abs.((π*f*Fs/4800) ./ sin.(π*f*Fs/4800)), 0.0]; 
                   Hz=Fs, weight=[1;100])
g = DSP.Filters.PolynomialRatio(g_vec, [1.0])
Gdb = 20*log10.(abs.(freqz(g,f,1.0)))

clf()
plot( f*Fs/1e3, -Pdb )
plot( f*Fs/1e3, Pdb )
plot( f*Fs/1e3, Gdb )
plot( f*Fs/1e3, Gdb - Pdb )
grid()

setp(gca(), ylim=(-120,40), xlim=(0,30))
xlabel("frequency (KHz)")
ylabel("magnitude (dB)")
legend(("Sinc","Inverse sinc","Filter response","Sinc + Filter response"), loc="best")
fig_filename = "inverse_sinc_response.png"
println("Saving figure to '$fig_filename'")
savefig(fig_filename)
```

martinholters:
Ok, I finally got around to implementing a draft for the new API. Check the tests to get a feel for what it looks like. E.g. the differentiator now becomes
```julia
h = remez(201, [(0.05, 0.95) => (f -> f/2, f -> 1/f)]; neg=true, Hz=2.0);
```
I'm still not sure whether I like this tuple of functions better than a function returning a tuple, which would read
```julia
h = remez(201, [(0.05, 0.95) => (f -> (f/2, 1/f))]; neg=true, Hz=2.0);
```
(Note that I've changed that the callback to the desired and weight functions to undo the scaling by `Hz`, i.e. `f` now ranged between `0` and `Hz/2`, not `0.5` as before. Keep that in mind when trying the current state.)

The docs still need to be updated.

martinholters:
Merged as part of #242.

Issue #228, opened 2018-09-09
Convolution of AbstractVector{<:Number}


RuiRojo:
Right now, things like `conv([1, 0], 1:5)` fail.

Is there a reason why the main definition is limited to `StridedVectors` instead of any `AbstractVector{<:Number}`?

ssfrr:
I've had on my back-burner a project to include some kind of dummy `AbstractArray` type in the tests and make sure that all the DSP functionality works as broadly as possible. I think there are definitely some places in there where the function types are just tighter than necessary, though there are also probably memory layout considerations when arrays get passed to C libraries like FFTW. That said we have a naive convolution implementation, so it seems like we should be able to have an `AbstractArray` fallback that does the naive convolution. There's probably also an FFT implementation that doesn't depend on a particular memory layout.

So in general yeah, it would be great to get DSP.jl working on `AbstractArray` as much as possible, someone just needs to write the tests and work through any edge cases.

RuiRojo:
The fallback could always just `collect` the vectors if the naive fallback it's too slow in comparison. I don't think it's required that the output vector be of the same type, but one can add special cases in time to make it the case as much as possible.

Can we add either the naive or the collect fallback and work through the edge cases, optimizations, etc, in time?

I do get that this may introduce future backward incompatibilities if we then want to create custom behaviour for a particular type of `AbstractArray` (say, we want `OffsetArray`s to convolve adding the offsets).

RuiRojo:
I mean something along these lines

```
function conv(u::AbstractVector{U}, v::AbstractVector{V}) where {U<:Union{Real, Complex}, V<:Union{Real, Complex}}
    T = promote_type(U, V)

    return conv(collect(T, u), collect(T, v))
end
function conv(u::AbstractVector{U}, v::AbstractVector{U}) where U<:Real
    T = Float64
    
    return conv(convert(Vector{T}, u), convert(Vector{T}, v))
end
function conv(u::AbstractVector{U}, v::AbstractVector{U}) where U<:Complex
    T = Complex{Float64}
    
    return conv(convert(Vector{T}, u), convert(Vector{T}, v))
end
function conv(u::AbstractVector{<:Number}, v::AbstractVector{<:Number})
    # fallback implementation in Julia that should work with any Number (units, etc)
end
```

martinholters:
> say, we want `OffsetArrays` to convolve adding the offsets

[...]

PR #231, opened 2018-09-17
WIP: Fix for #228

RuiRojo:
This is a first attempt, feedback would be nice.

I used a generated function in an attempt to aviod creating multiple copies while trying to dispatch, and also avoid introducing a type instability.

Tests still missing

Naive implementation for all <:Numbers is also missing (or at least not
plugged in, there should be one somewhere)


martinholters:
I'd really prefer using multiple dispatch instead of a `@generated` function with `hasmethod` calls here. Though that may mean disentangling the existing methods. Alternatively, one could just implement the totally general `conv` and leave anything else for future optimization.

galenlynch:
Superseded by #279

PR #240, opened 2018-11-15
window functions: Adds `zerophase` and `padding` options and expands documentation

ssfrr:
This PR makes a few changes to the window functions:

1. Refactors the window functions such that each one is defined in terms of a continuous function in [-0.5, 0.5], which is sampled to the appropriate size by a `makewindow` function.
2. Adds a `zerophase` option which gives a zero-phase periodic window
3. Adds a `padding` option which zero-pads the window (for the zero-phase windows this adds the padding to the middle of the returned vector)
4. Adds a bunch of documentation for each of the windows, including a plot for each one generated with UnicodePlots.jl.


ssfrr:
Also note this should be non-breaking - the default options for the keyword arguments match the previous behavior, and the refactored window functions pass the previous tests.

ssfrr:
coverage seems to have dropped because of code that's used to generate the docstrings, which apparently isn't counted in coverage calculations.

galenlynch:
Beautiful documentation. I'll try to review this sometime this week, if no one else does before then.

gen_winplots.jl
```diff
@@ -0,0 +1,40 @@
...
+        (blackman, ()),
+        (kaiser, (3)),
+        (dpss, (2, 1)),
+        ]
+
+        ymax = winfunc == dpss ? 0.2 : 1.0
+        n = winfunc == triang ? 7 : 69
+
+        fname = split(string(winfunc), ".")[end]
+        println(io, "$(fname)_winplot = padplot(\"\"\"")
```
galenlynch: Shouldn't this be `const`?

src/windows.jl
```diff
@@ -25,150 +25,479 @@ export  rect,
...
+function hanning(n::Integer; padding=0, zerophase=false)
+    makewindow(n, padding, zerophase) do x
+        0.5*(1+cos(2pi*x))
+    end
+end
+```
+"""
+function makewindow(winfunc::Function, n, padding, zerophase)
+    win = zeros(n+padding)
+    if n == 1
```
galenlynch: It would be nice to validate inputs to this function, since most (all?) window functions now use it to generate windows. e.g. ensure that `padding >= 0` (negative values currently throw a cryptic error), or even that `n >=0`, and provide helpful error messages.

src/windows.jl
```diff
@@ -25,150 +25,479 @@ export  rect,
...
+within the window, and 0 outside of it.
 
-Rectangular window function of length `n`.
+$(twoD_docs())
+
+$zerophase_docs
 """
-function rect(n::Integer)
-    ones(n)
+function rect(n::Integer; padding=0, zerophase=false)
```
galenlynch: It would be nice to have type bounds on KW args to window functions.

ssfrr: Agreed. Just pushed a commit that adds tightens the type bounds across the board.

src/windows.jl
ssfrr: Good catch - added some checks.

gen_winplots.jl
ssfrr: Sure could be. And now is. :)

PR #242, opened 2018-11-19
Sirtom67/remez func

sirtom67:
@martinholters I finally checked out your remez API changes - looks very good! I started with your latest on that branch and merged master in.

I wrote some improved doc strings and think we could merge it.

I still would like to add the inverse sinc filter example but I think that can be handled in addition to this branch.


src/Filters/remez_fir.jl
```diff
@@ -380,33 +353,134 @@ plot(f, 20*log10.(abs.(freqz(b2,f,1.0))))
...
+    or functions. If a function, the function should accept a real frequency and
+    return the real desired response or real weight. Examples:
+      LPF with unity weights. [(0, 0.475) => 1, (0.5, 1.0) => 0].
+      LPF with weight of 2 in the stop band. [(0, 0.475) => (1, 1), (0.5, 1.0) => (0, 2)].
+      BPF with unity weights. [(0, 0.375) => 0, (0.4, 0.5) => 1, (0.525, 1.0) => 0]
+      Hilbert transformer. [(0.1, 0.95) => 1]; neg=true.
+      Differentiator. [(0.01, 0.99) => (f -> f/2, f -> 1/f)]; neg=true.
+- `Hz::Real`: The sampling frequency in Hz. Default is 1.
+- `neg::Bool`: Whether the filter has negative symmetry or not. Default is false.
+    If false, the filter is symmetric. If true, the filter is asymmetric.
```
martinholters: I think "asymmetric" is the wrong term here. Should be "odd-symmetric", if that is a word.

martinholters:
Thanks for picking this up! (For those not following along, this builds on #230, which it replaces.)

Barring objections, I'll merge this in a couple of days. (And please bump me in the likely event I forget.)

src/Filters/remez_fir.jl
ssfrr: is the property that [scipy calls](https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.windows.hann.html#scipy.signal.windows.hann) "symmetric" vs "periodic"? In that nomenclature a "symmetric" window has the same number of points on both sides of the peak (and is usually an odd-length window so the peak is on a sample) vs a "periodic" which only includes one of the extrema (so for an even-length periodic window (the common case) there's the peak still falls on a sample). The latter is called periodic because if you consider it to wrap around then the minimum value isn't repeated.

martinholters: No, `neg=true` means that `h[n]=-h[end+1-n]`.

sirtom67: Thanks for the comments, I pushed a minor improvement to the "neg" argument description.

PR #244, opened 2018-11-23
Sirtom67/remez doc improvements

sirtom67:
Improve remez doc strings:
put simple API method first,
only highlight API differences in scipy API, 
and fix numerous small markup issues.


galenlynch:
It seems like this PR overlaps with #243?

sirtom67:
Thanks Galen!

galenlynch:
My pleasure!

Issue #245, opened 2018-11-28
Is the code in the /doc folder obsolete?


galenlynch:
I can't tell if it's being used, but I suspect that it is not. Should it be deleted?

galenlynch:
closed via #246

PR #183, opened 2017-10-09
Switch documentation to Documenter, part 1

martinholters:
This copies the documentation from the RST files to docstrings and markdown files and enables building and deploying using Documenter. I tried to reproduce the old docs as faithfully as possible; any doc improvements should be done in separate PRs.

Note: Before the deployment works, Github/Travis need to be configured accordingly for which I lack the permissions on the Github side. @ararslan could you do that?

One issue is that I'm getting spurious "docstrings potentially missing" warnings for
```
    DSP.Periodograms.periodogram :: Union{Tuple{AbstractArray{T,1}}, Tuple{T}} where T<:Number
    DSP.Periodograms.periodogram :: Union{Tuple{AbstractArray{T,2}}, Tuple{T}} where T<:Real
```
which I don't understand. The present docstrings for `periodogram` do appear in the docs, and those signatures look strange to me. Any idea what's going on there?

Once the deployment is verified to be working, in the next step I'll change the links in the README.md and replace the readthedocs docs with a dummy page pointing at the new docs.

Issue #182, opened 2017-10-05
Documentation doesn't build


martinholters:
From the readthedocs build log:
```
Running Sphinx v1.5.3
making output directory...
WARNING: sphinx.ext.pngmath has been deprecated. Please use sphinx.ext.imgmath instead.

Traceback (most recent call last):
  File "/home/docs/checkouts/readthedocs.org/user_builds/dspjl/envs/latest/local/lib/python2.7/site-packages/sphinx/cmdline.py", line 295, in main
    opts.warningiserror, opts.tags, opts.verbosity, opts.jobs)
  File "/home/docs/checkouts/readthedocs.org/user_builds/dspjl/envs/latest/local/lib/python2.7/site-packages/sphinx/application.py", line 199, in __init__
    self.setup_extension(extension)
  File "/home/docs/checkouts/readthedocs.org/user_builds/dspjl/envs/latest/local/lib/python2.7/site-packages/sphinx/application.py", line 523, in setup_extension
    ext_meta = mod.setup(self)
  File "/home/docs/checkouts/readthedocs.org/user_builds/dspjl/envs/latest/local/lib/python2.7/site-packages/sphinx/ext/mathjax.py", line 72, in setup
    raise ExtensionError('sphinx.ext.mathjax: other math package is already loaded')
ExtensionError: sphinx.ext.mathjax: other math package is already loaded

Extension error:
sphinx.ext.mathjax: other math package is already loaded
```

Maybe instead of fixing this, it's time to move to Documenter.jl?

ararslan:
+1 for Documenter.

martinholters:
I'll start porting things over, but it might be a few days...

PR #246, opened 2018-11-28
Remove unused doc folder

galenlynch:
I think the /doc folder may be unused (see #245).


galenlynch:
Looking through #182, #183, and #184, it seems like this folder in indeed obsolete. The documentation built just fine when it was removed, so I'm going to go ahead and merge this.

martinholters:
Ref. #184 which deleted most of `doc`. Yes, I think all of this can safely be removed now. But I don't know whether anything on readthedocs should be configured/disabled before. (And if so, who has the rights to do so.)

martinholters:
https://dspjl.readthedocs.io/en/latest/ still seems to be built from the most recent commit to master. If we can disable that, that would be good.

martinholters:
Ah, that's unfortunate. Then the best course of actions would be to fix the link in index.rst, have the fixed page build, disable building, merge this PR. OTOH, I'm not sure how bad it would be if the readthedocs page just disappeared.

galenlynch:
I don't think I have permissions to change the readthedocs webook. I can add the fixed link soon.

galenlynch:
Ok I fixed the link.

martinholters:
Great.

@ararslan, can you disable building the docs on readthedocs (e.g. by removing the integration on this repo that triggers it)?

ararslan:
I've deactivated the ReadTheDocs integration for this repo.

martinholters:
Thanks! Then I guess this should be good to go.

galenlynch:
Thanks guys!

PR #252, opened 2018-12-09
filter design documentation improvements

sirtom67:
1) include examples from the unittests in "remez" doc
2) reorganize the "filters.md" doc to more accurately
   reflect reality

### DIFF
--- a/docs/src/filters.md
+++ b/docs/src/filters.md
@@ -57,18 +57,24 @@ Most analog and digital filters are constructed by composing
 [response types](@ref response-types), which determine the frequency
 response of the filter, with [design methods](@ref design-methods),
 which determine how the filter is constructed.
+The response type is low-pass, high-pass, band-pass or
+band-stop and includes the edges of the bands.
+The design method is Butterworth, Chebyshev1, Chebyshev2, Elliptic, 
+or FIRWindow, and includes any
+necessary parameters for the method that affect the shape of the response,
+such as filter order, ripple, and attenuation.
 
 ```@docs
 analogfilter
 digitalfilter
 ```
 
-For some filters, the design method inherently implies a response type.
-Such filters are documented below.
+For some filters, the design method is more general or 
+inherently implies a response type;
+these [direct design methods](@ref direct-design-methods)
+include "remez" which designs equiripple FIR
+filters of all types, and "iirnotch" which designs a 2nd order "biquad" IIR notch filter.
 
-```@docs
-iirnotch
-```
 
 ### [Filter response types](@id response-types)
 
@@ -94,7 +100,12 @@ Elliptic
 
 ```@docs
 FIRWindow
+```
+
+### [Direct filter design methods](@id direct-design-methods)
+```@docs
 remez
+iirnotch
 ```
 
 ## Filter response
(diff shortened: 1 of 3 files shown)


docs/src/filters.md
```diff
@@ -57,18 +57,24 @@ Most analog and digital filters are constructed by composing
 [response types](@ref response-types), which determine the frequency
 response of the filter, with [design methods](@ref design-methods),
 which determine how the filter is constructed.
+The response type is low-pass, high-pass, band-pass or
```
galenlynch: Could you add links to the response types? i.e. instead of using "low-pass", use ```[`Lowpass`](@ref)```. More details on the link syntax [here](https://juliadocs.github.io/Documenter.jl/stable/man/syntax/#@ref-link-1).

docs/src/filters.md
```diff
@@ -57,18 +57,24 @@ Most analog and digital filters are constructed by composing
...
+band-stop and includes the edges of the bands.
+The design method is Butterworth, Chebyshev1, Chebyshev2, Elliptic,
```
galenlynch: Could you also add links here?

src/Filters/remez_fir.jl
```diff
@@ -324,7 +324,7 @@ Construct a length 35 filter with a passband at 0.15-0.4 Hz
 those bands - the transition bands - is unspecified.
 
 ```julia-repl
-julia> bpass = remez(35, [(0, 0.1)=>0, (0.15, 0.4)=>1, (0.45, 0.5)=>0])
```
galenlynch: I think it's common to have the `julia>` prompt inside `julia-repl` code blocks. E.g. see the second block [here](https://docs.julialang.org/en/v1/manual/documentation/index.html).

src/Filters/remez_fir.jl
```diff
@@ -347,6 +347,59 @@ plot(f, 20*log10.(abs.(freqz(b,f,1.0))))
 plot(f, 20*log10.(abs.(freqz(b2,f,1.0))))
 grid()
 ```
+
+# Examples from the unittests - standard (even) symmetry.
```
galenlynch: Would it be worth it to make these doc-tests? The syntax for doing so is [number six in this list](https://docs.julialang.org/en/v1/manual/documentation/index.html).

sirtom67:
Galen: thanks for the review! I put back the "julia>" in the examples, and made links to the response types and design methods. I don't think the doc-tests make sense here, because the actual tests do more than just design the filter (they design it and then compare it to the scipy reference filter that is stored in ascii form in the test directory).

martinholters:
Doctests are not meant to replace unit tests, but to make sure the documentation, especially the examples therein, stays up to date.

galenlynch:
Doctests must be runnable 'as-is', i.e. if you took the code and pasted it into a new Julia REPL session. Examples that use packages, such as the ones in your expanded doc-string, must also have the required `using` statements. I think the doctest documentation in [Documenter](https://juliadocs.github.io/Documenter.jl/stable/man/doctests/) is more complete, and show how to deal with multiple examples, as well as hiding the output of the examples. See specifically [this section](https://juliadocs.github.io/Documenter.jl/stable/man/doctests/#Setup-Code-1) on set-up prior to the doctest. I'm not 100% sure that the syntax in Documenter applies to doc-strings in Julia code, but I would try it to see if it works. If it doesn't work, you may have to prepend `using DSP` to your examples.

galenlynch:
See JuliaDocs/Documenter.jl#910

Issue #142, opened 2016-12-23
InexactError() constructing PolynomialRatio


ma-laforge:
The following throws an `InexactError()`:
```
using DSP
b=[1]; a=[10]
PolynomialRatio(b,a)
```

which can somehow be avoided by making `a` an array of `Float64`, instead:
```
using DSP
b=[1]; a=[10.0]
PolynomialRatio(b,a)
```

Also: there is no error if `b=[10]; a=[1]`, for some reason.

hamzaelsaawy:
Its from [here](https://github.com/JuliaDSP/DSP.jl/blob/73cfda49019143fe6ed1b6a72555778470f16e34/src/Filters/coefficients.jl#L53), where the numerator and denominator are divided by the leading term of the denominator polynomial.

Is there a need to do so?

ma-laforge:
No.  There is no real need to do so.  It was just a little unexpected.

I think from the ***point of view of usability***, it is quite confusing to see an inexact error when all you do is construct a polynomial with 2 arrays of integers.

However, I do find it a bit odd that the polynomials get normalized when they get created.  That is also unexpected.  I personally would not have done this normalization unless there was a really good reason for it (which there might be, I suppose).

PS Thanks for the feedback.

martinholters:
We could add an optional parameter to `PolynomialRatio` to suppress normalization, defaulting to the current behavior for compatibility.

ma-laforge:
I don't really have a strong opinion about the optional parameter.

It would not help with the confusing message - because you would still need to be aware that the default behaviour is to normalize, so the problem would remain.

But this brings up an interesting question: why ***does*** DSP.jl normalize the polynomial at all if it is still just as valid ***not*** to do the normalization?

Shouldn't `normalize()` be a function that is called by the functions operating on the polynomial (assuming that normalization is actually needed for the operation)?

hamzaelsaawy:
I think normalizing is used primarily in the  [FIR filtering algorithm](https://github.com/JuliaDSP/DSP.jl/blob/5abc50754793fc6cece466454e5330839125aa65/src/dspbase.jl#L76) but interestingly enough that code doesn't assume the coefficients are normalized, so maybe we can do without in the `PolynomialRatio` object?  

The only issue would be a `PolynomialRatio{Int}` might return a `Biquad{Float64}` since the latter is normalized by definition, and that would be weird for some people to see, but I think a lot of the filter coefficient objects change types when the representation changes.

[...]

PR #257, opened 2019-02-18
PolynomialRatio construction

galenlynch:
I got super carried away with issue #142, and ended up coming up with my sketch of a solution.

```julia
julia> PolynomialRatio([1, 1, 1], [1, 1, 1])
PolynomialRatio{Float64}(Poly(1.0 + 1.0*x + 1.0*x^2), Poly(1.0 + 1.0*x + 1.0*x^2))

julia> PolynomialRatio([1, 1, 1], [2, 1, 1])
PolynomialRatio{Float64}(Poly(0.5 + 0.5*x + 0.5*x^2), Poly(0.5 + 0.5*x + 1.0*x^2))


julia> PolynomialRatio{Int}([1, 1, 1], [2, 1, 1])
ERROR: ArgumentError: PolynomialRatio{<:Integer} requires normalized coefficients, i.e. a[end] must be 1

julia> PolynomialRatio{Int}([1, 1, 1], [1, 1, 1])
PolynomialRatio{Int64}(Poly(1 + x + x^2), Poly(1 + x + x^2))
```

I'll expand on the details later, but just want to get this PR out there for now and see what people think.

### DIFF
--- a/src/Filters/Filters.jl
+++ b/src/Filters/Filters.jl
@@ -3,7 +3,7 @@ using ..DSP: @importffts, mul!, rmul!
 using ..Unwrap
 using Polynomials, ..Util
 import Base: *
-using Compat: copyto!, undef, argmin
+using Compat: copyto!, undef, argmin, Nothing
 import Compat
 using Compat.LinearAlgebra: I
 using Compat.Statistics: middle
--- a/src/Filters/coefficients.jl
+++ b/src/Filters/coefficients.jl
@@ -207,7 +207,7 @@ Base.promote_rule(::Type{Biquad{T}}, ::Type{Biquad{S}}) where {T,S} = Biquad{pro
 ZeroPoleGain{Z,P,K}(f::Biquad) where {Z,P,K} = ZeroPoleGain{Z,P,K}(PolynomialRatio(f))
 ZeroPoleGain(f::Biquad) = ZeroPoleGain(convert(PolynomialRatio, f))
 
-function PolynomialRatio(f::Biquad{T}) where T
+function PolynomialRatio(f::Biquad{T}, coeff_type::Type = Nothing) where T
     if f.b2 == zero(T) && f.a2 == zero(T)
         if f.b1 == zero(T) && f.a1 == zero(T)
             b = T[f.b0]
@@ -221,7 +221,7 @@ function PolynomialRatio(f::Biquad{T}) where T
         a = T[one(T), f.a1, f.a2]
     end
 
-    PolynomialRatio(b, a)
+    PolynomialRatio(b, a, coeff_type)
 end
 
 function Biquad{T}(f::PolynomialRatio) where T
--- a/src/Filters/coefficients.jl
+++ b/src/Filters/coefficients.jl
@@ -45,41 +45,7 @@ Base.promote_rule(::Type{ZeroPoleGain{Z1,P1,K1}}, ::Type{ZeroPoleGain{Z2,P2,K2}}
 # Transfer function form
 #
 
-struct PolynomialRatio{T<:Number} <: FilterCoefficients
-    b::Poly{T}
-    a::Poly{T}
-
-    function PolynomialRatio{Ti}(b::Poly{Ti}, a::Poly{Ti}) where {Ti<:Number}
-        _polynomialratio_check_coeffs(b::Poly, a::Poly)
-        new(b / a[end], a / a[end])
-    end
-
-    function PolynomialRatio{Ti}(b::Poly{Ti}, a::Poly{Ti}) where {Ti<:Integer}
-        _polynomialratio_check_coeffs(b::Poly, a::Poly)
-        if a[end] != 1
-            throw(
-                ArgumentError(
-                    "PolynomialRatio{<:Integer} requires normalized " *
-                    "coefficients, i.e. a[end] must be 1"
-                )
-            )
-        end
-        new(b, a)
-    end
-end
-
-function _polynomialratio_check_coeffs(b::Poly, a::Poly)
-    if all(iszero, coeffs(b)) || all(iszero, coeffs(a))
-        throw(ArgumentError("filter must have non-zero numerator and denominator"))
-    end
-    nothing
-end
-
-# This allows users to easily specify coefficient data type while minimizing the
-# need for two sets of functions PolynomialRatio{T}(...) and
-# PolynomialRatio(...), by using standard default arguments and a sentinal
-# coefficient type Nothing to indicate that there is no desired coefficient type
-PolynomialRatio{T}(args...) where T = PolynomialRatio(args..., T)
+const NotClosedUnderDivision = Union{Integer, Complex{<:Integer}}
 
 """
     PolynomialRatio(b, a)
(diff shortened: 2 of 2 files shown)


hamzaelsaawy:
I think we should mirror the [Julia standard](https://docs.julialang.org/en/v1/manual/style-guide/index.html#Write-functions-with-argument-ordering-similar-to-Julia-Base-1) and have `PolynomialRatio(::T, args...)` rather than having `::T` at the end.

galenlynch:
Ok so here was my rationale:

Before this PR, you could construct `PolynomialRatio` objects with or without specifying the type by calling `PolynomialRatio{T}(...)` or `PolynomialRatio(...)`, respectively. Without specifying the type, `PolynomialRatio` figures out its own data type using standard promotion rules. Specifying the type overrides that. 

Using the parametric functions to override type promotion requires each step in the construction process to have two methods with mostly redundant content, except the parametric methods call the parametric version of the next step in the construction process, and non-parametric methods call the non-parmatric next step. Instead of having these two chains of methods, having an optional type parameter at the end with a "dummy" value of `Nothing` allows you to combine both methods. Finally, a convenience method `PolynomialRatio{T}(...) = PolynomialRatio(..., T)` allows users to create `PolynomialRatio` objects at any step of the construction process (with other filters, vectors, or polynomials, with various mismatching types), while still letting them override normal type promotion. 

While the optional `coeff_type` argument enables this, I was intending users to use the pre-existing interface to specify the parametric type, and not the `coeff_type` argument. Since optional arguments have to be at the end, it's certainly a lot easier to have the optional type not at the beginning of the argument list. Adding yet another interface, `PolynomialRatio(T, ...)` is possible, but it seems like a separate issue to me. Also, while I agree that interface is use a lot in base Julia, it's not in the style guide.

hamzaelsaawy:
I am actually in the process of gutting most of the parametric constructors since my take is that they are redundant, add complexity, and generic versions are cleaner vs parameterized constructors.
(There are other type-related changes focused on type-consistency, so I can definitely hold off until a separate PR.)

And the two interfaces can be merged like: 

```julia
PolynomialRatio(::Type{T}, b::Poly{<:Number}, a::Poly{<:Number}) where {T<:Number}= 
    PolynomialRatio{T}(b, a)

PolynomialRatio(b::Poly{T}, a::Poly{S}) where {T<:Number, S<:Number}= 
    PolynomialRatio(promote_type(Float64, S, T), b, a)
```

This is how julia defines the [`zeros`/`ones`](https://github.com/JuliaLang/julia/blob/5d4eaca0c9fa3d555c79dbacdccb9169fdf64b65/base/array.jl#L461)/etc.

Also, the [style guide does call out that the types (bullet 4) should be before other inputs (that aren't mutated) and everything else.](https://docs.julialang.org/en/v1/manual/style-guide/index.html#Write-functions-with-argument-ordering-similar-to-Julia-Base-1)

martinholters:
I don't like the constructors with the eltype argument. If I want a `PolynomialRatio{MyNumber}`, I find it much clearer to write `PolynomialRatio{MyNumber}(...)` than `PolynomialRatio(MyNumber, ...)`. I see how it makes sense implementation-wise as an optional argument to avoid code-duplication, but I'd hope we can refactor that is some other manner. A rough sketch of what I could imagine this might look like:
```julia
struct PolynomialRatio{T<:Number} <: FilterCoefficients
    b::Poly{T}
    a::Poly{T}
    function PolynomialRatio{Ti}(b::Poly{Ti}, a::Poly{Ti}) where {Ti<:Integer}
        _polynomialratio_check_coeffs(b, a)
        new(b, a)
    end
    PolynomialRatio{Ti}(b::Poly, a::Poly) where {Ti<:Integer} = 
        PolynomialRatio{Ti}(convert(Poly{TI}(b), convert(Poly{TI}(a))
    PolynomialRatio{Ti}(b, a) where {Ti<:Integer} =
        PolynomialRatio{Ti}(_to_polys(b, a)...)
end
PolynomialRatio(b,a) = PolynomialRatio(_to_polys(b, a)...)
function PolynomialRatio(b::Poly{T}, a::Poly{T}) where T
    b /= a[end]
    a /= a[end]
    PolynomialRatio{eltype(a)}(b, a)
end

function _to_polys(b, a)
    # converts single numbers, vectors, and Poly's of different type to Poly's of promoted type
    # ...
end
function _polynomialratio_check_coeffs(b, a)
    # as before, but also checks `a[end]` == one(a[end])`
end
```
That way, when explicitly giving the eltype, one has to provide normalized coefficients, though. I'd say that's ok. In fact, for `MyNumber` not being a subtype of `Integer`, is it valid to assume `/(::MyNumber, ::MyNumber)` to give a `MyNumber`? I'm not sure.

I'm probably missing some things here, which you probably can easily spot having just worked on this code. But maybe the general idea still makes sense?

martinholters:
Oh, and I while I'm not too happy about the current implementation here, as pointed out in https://github.com/JuliaDSP/DSP.jl/issues/142#issuecomment-465564433, I think I very much do like the obtained behavior.

galenlynch:
The constructor in the PR still has the old interface, provided by the method `PolynomialRatio{T}(...) = PolynomialRatio(..., T)`. The optional eltype argument is definitely a bit of hack to make things work, and was not intended to be documented or used. I agree that we should be able to do better.

Yeah I like your suggested implementation. The tricky part, which I was struggling with, is that if a user use a parametric constructor `PolynomialRatio{T}(...)` then it specifies the eventual type parameter, but should not interfere with conversion of arrays to polys etc. I think something like your `_to_polys` with an optional type parameter may do the trick. Basically just replace the constructor chain with the a `_to_polys` chain to remove the optional type parameter from the public interface.

galenlynch:
@martinholters I tried to address your comments. What do you think?

martinholters:
Maybe we need to think about the desired API a bit. In the current form, this PR normalizes the coefficients unless a) the type parameter is given explicitly and b) it is a subtype of `Integer` or `Complex{<:Integer}`. That is, if I have some custom type that is not an `Integer`, but still not closed under division (some fixed point format, maybe), I'll run into problems.

I therefore think we should consider the following options:
* Never normalize in `PolynomialRatio{T}(...)`, only normalize in `PolynomialRatio(...)`. (What I implied in my implementation sketch above.) This would mean that one might have to do the normalization before calling `PolynomialRatio{T}(...)`. Not sure how often one would want to call the typed constructor with not-yet-normalized coefficients.
* Allow normalization to be switched on/off with an optional parameter. We could make the _default_ here be chosen based on the coefficient type, but that might be too automagical.

galenlynch:
I think normalization should be required, whatever API we decide on.

martinholters:
Ah, yes, sorry, that wasn't clear. I didn't mean to propose an option that allows non-normalized coefficients to be stored in the end. It's just the question between automatically normalizing and just asserting that the given coefficients are already normalized.

galenlynch:
I'm fine with either suggestion!

hamzaelsaawy:
> Allow normalization to be switched on/off with an optional parameter.

I vote for this, since different effects from different constructors seems ... weird. 

> asserting that the given coefficients are already normalized.

@martinholters, did you have in mind something like `isnormalized(f::PolynomialRatio) = f.a[end] == 1`  or a type parameter?

Also, i think we should revisit forcing `a[end] == 1` before filtering (I know we discussed it [here](https://github.com/JuliaDSP/DSP.jl/issues/142#issuecomment-465564433):  
In a weird case, if someone wants to simulate integer (or fixed point) coefficients on integer (or fixed point) data and `a[end] != 1`, converting to floats could cause round off error in representing the coefficients and then loss of significance during filtering(as it would with integers, but the loss of 12 bits in doubles point increases the chance), resulting in a less accurate representation.  
Moreover, rounding the integer coefficients before construction would be more inaccurate, so maybe we could add a case to `filt` where it normalizes at every time step if the filter in un-normalized (but returns a float regardless of inputs to avoid type stability issues) but ensure that normalization is the default. 
 
The other alternative is not be able to simulate all integer filters/systems, which would be bad when designing for embedded applications

galenlynch:
Filtering with un-normalized coefficients seems like an extremely specialized use-case, to me.

martinholters:
Agree, let's restrict `PolynomialRatio` to normalized coefficients.

From the question in https://github.com/JuliaDSP/DSP.jl/pull/257#issuecomment-467693101, I get it that I still may not have been clear. So to clarify:
* `PolynomialRatio(b,a)` should always normalize the coefficients. If the division changes the coefficient type, then so be it, e.g. `PolynomialRatio(::Vector{Int},::Vector{Int})` will return a `PolynomialRatio{Float64}`.
* The question is whether `PolynomialRatio{T}(b,a)` normalizes, too, or throws an error if `a[end]!=1`. I'm not a big fan of making that depend on `T`, at least on the hard-coded form `T<:Union{Integer,Complex{<:Integer}}`. I'd be ok with never normalizing, but would also be ok with an optional parameter to control this. But to be clear: not normalizing here would mean throwing if  `a[end]!=1`; we should never construct a `PolynomialRatio` where the stored `a[end]` is not one.

galenlynch:
I'm probably not going to finish this up, anyone else should obviously feel free to.

PR #265, opened 2019-03-17
Increase conv performance by making fewer intermediate arrays

galenlynch:
`conv` pads its inputs to use fourier convolution with linear convolution.
Previously, this padding was done by making separate arrays for each pad, and
then concatenating them with the input arrays. I have eliminated the
intermediate pad arrays, and thereby eliminated the creation of two arrays for
each call of `conv`.

I have additionally added a new function, `_circ_conv`, which performs circular
convolution using fourier methods. By separating this out form the `conv`
function, dispatch on the element type can be done my Julia, instead of by a
branch in `conv`. This also allows future use of `_circ_conv` for intentional
circular convolution.

### DIFF
--- a/src/dspbase.jl
+++ b/src/dspbase.jl
@@ -117,27 +117,31 @@ function deconv(b::StridedVector{T}, a::StridedVector{T}) where T
     filt(b, a, x)
 end
 
-function _zeropad(u, ntot, nu = length(u))
-    padded = similar(u, ntot)
+function _zeropad!(padded, u, ntot, nu = length(u))
     copyto!(padded, 1, u, 1, nu)
     padded[nu + 1:ntot] .= 0
     padded
 end
+_zeropad(u, ntot, nu = length(u)) = _zeropad!(similar(u, ntot), u, ntot, nu)
 
-function _circ_conv(
-    upad::StridedVector{T}, vpad::StridedVector{T}, np2::Integer
+function _conv(
+    u::StridedVector{T}, v::StridedVector{T}, npad, nu, nv
 ) where T<:Real
-    p = plan_rfft(upad)
-    uf = p*upad
-    vf = p*vpad
-    uf .*= vf # Multiply in place
-    irfft(uf, np2)
+    padded = _zeropad(u, npad, nu)
+    p = plan_rfft(padded)
+    uf = p * padded
+    _zeropad!(padded, v, npad, nv)
+    vf = p * padded
+    uf .*= vf
+    irfft(uf, npad)
 end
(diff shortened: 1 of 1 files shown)


galenlynch:
This PR results in a sizeable performance increase in `conv`, at the cost of some code readability.

Some simple testing shows that with large input vectors (like I often deal with), the changes increase performance by about 50%, mostly a result of fewer GC interruptions. In addition, memory usage is halved.

Before changes:
```julia
using DSP, BenchmarkTools
julia> u = rand(1000000);
julia> @benchmark conv(u, u)
BenchmarkTools.Trial: 
  memory estimate:  122.08 MiB
  allocs estimate:  156
  --------------
  minimum time:     91.795 ms (2.14% GC)
  median time:      157.906 ms (36.03% GC)
  mean time:        130.520 ms (24.95% GC)
  maximum time:     202.017 ms (55.41% GC)
  --------------
  samples:          39
  evals/sample:     1
```

After changes:
```julia
using DSP, BenchmarkTools
julia> u = rand(1000000);
julia> @benchmark conv(u, u)
BenchmarkTools.Trial: 
  memory estimate:  61.04 MiB
  allocs estimate:  146
  --------------
  minimum time:     76.519 ms (1.51% GC)
  median time:      79.611 ms (1.49% GC)
  mean time:        82.073 ms (3.67% GC)
  maximum time:     191.095 ms (53.96% GC)
  --------------
  samples:          61
  evals/sample:     1
```

While `conv` was already fast, small changes make a big difference when doing Monte Carlo simulations, for example.

galenlynch:
Also it's amazing how much faster this is than Matlab: for the same problem (convolution of two million element vectors), Matlab takes 51.15 seconds, compared to 0.079 seconds in Julia. Perhaps they're doing time domain convolution or something?

galenlynch:
Any comments?

HDictus:
Similar still creates a new array right? You could consider trying PaddedViews.jl to see if that makes it even better, though it would be an additional dependency of course.
I can't think of a way to improve the readability of it, and I think it's only a small difference in any case.
It seems like it would be a good idea to add some default benchmarks for certain functions, so that one can test the effects of their changes on that without needing to rewrite it. Maybe there's even a way to incorporate that in the CI?

galenlynch:
@HDictus the benchmarks are a good idea, I think lots of repos have something like that built into CI to detect performance regressions. If you'd want to give it a shot, it would be a welcome addition.

Not sure about the PaddedViews, I imagine you'd have to benchmark them fair carefully to see if slower array access through the view would be worth the memory savings.

Issue #253, opened 2018-12-11
`xcorr` pads inputs to equal length


ssfrr:
Naively I'd expect `length(xcorr(X, Y)) == length(X) + length(Y) - 1`, but instead it's `2*max(length(X), length(Y))-1`. This is because it pads its inputs to equal length before handing off to `conv`. I'm not sure why it does this, given that `conv` pads its inputs as part of the FFT algorithm (but then it undoes the effect of the padding before returning the value).

Does anyone know whether this padding is desirable for some reason?

I've been bitten twice by this (once a while ago, and then apparently I forgot about it in the meantime), so at the very least I'd like to document it, but it seems somewhat questionable.

ssfrr:
As further justification why we might want to change this, it's weird that `conv` and `xcorr` will give results with different lengths.

I know this is a breaking change, so we'd need to figure out whether there's a clean way to deprecate it, maybe with a keyword argument that enables the new behavior, and then at some point in the future we can change the default of the kwarg.

galenlynch:
This behaviour is the same as Matlab's `xcorr`, as well as SciPy's `correlate` with default arguments (i.e. `mode = 'full'`). Perhaps adding a keyword to change the `mode` of the result would be nice, but I think the current behaviour is in line with other popular DSP packages.

ssfrr:
In those packages does `conv` do the same thing, or do they likewise have `conv` and `xcorr` behaving differently?

galenlynch:
Matlab - 
No, `conv` is different in both packages.

xcorr: always pads to `2*(max(N,M)) - 1` where `N` is length of the first vector and `M` is length of second vector
conv: padding controlled by `shape` optional paramter, defaults to `full` which is `N + M - 1`

SciPy -
correlate - same default behavior as matlab's xcorr
convolve - same default behavior as matlab's conv

Also, the current behavior returns a symmetric cross correlation, which is nice.

galenlynch:
@ssfrr now that I've been bitten by this a few times as well I agree with you.

galenlynch:
Fixed by #266

ssfrr:
🎉

Thanks @galenlynch!

galenlynch:
My pleasure!

PR #266, opened 2019-03-17
Make xcorr padding optional

galenlynch:
As brought up in #253, xcorr currently has the unintuitive behavior of doing
padding in addition to the padding done by `conv`. This means that for inputs of
different lengths, `xcorr` results are not the same length as `conv`, and that
the beginning of the result is filled with zeros. While Numpy and Matlab both
return padded results, they both fill the end of result with zeros, instead of
the beginning.

This commit changes the default behavior to be more in line with the
mathematical definition of cross-correlations, and also makes `xcorr` more in
line with the behavior of `conv`. This new behvaior can be disabled by
setting the new `mode` keyword to `:legacy`. Deprecation warnings are thrown if
the result of `xcorr` would be altered by the new behavior.

### DIFF
--- a/src/dspbase.jl
+++ b/src/dspbase.jl
@@ -212,10 +212,11 @@ function check_padmode_kwarg(padmode::Symbol, su::Integer, sv::Integer)
         if su != sv
             Base.depwarn(
             """
-            The default behavior for xcorr will changed in future versions.
-            For more details, see the documentation for xcorr. To avoid this
-            warning, specify padmode = :none or padmode = :longest where
-            appropriate
+            The default value of `padmode` will be changing from `:longest` to
+            `:none` in a future release of DSP. In preparation for this change,
+            leaving `padmode` unspecified is currently deprecated. To keep
+            current behavior specify `padmode=:longest`. To avoid this warning,
+            specify padmode = :none or padmode = :longest where appropriate.
             """
                 ,
                 :xcorr
@@ -231,19 +232,25 @@ end
     xcorr(u,v; padmode = :longest)
 
 Compute the cross-correlation of two vectors. The size of the output depends on
-`padmode` keyword argument: if `padmode = :none` then the result will be the same size
-as [`conv`](@ref) of `u` and `v`. If `padmode = :longest` then the result will have
-length `2*max(length(X), length(Y))-1`, where the beginning of the result will
-be padded with zeros. If `padmode = :none`, it will be `length(X) - length(Y) - 1`.
+the padmode keyword argument: with padmode = :none the length of the
+result will be length(u) + length(v) - 1, as with conv. With
+padmode = :longest the shorter of the arguments will be padded so they
+are equal length. This gives a result with length 2*max(length(u), length(v))-1,
+with the zero-lag condition at the center.
+
+!!! warning
+    The default value of `padmode` will be changing from `:longest` to `:none`
+    in a future release of DSP. In preparation for this change, leaving
+    `padmode` unspecified is currently deprecated.
 """
 function xcorr(u, v; padmode::Symbol = :default_longest)
     su = size(u,1); sv = size(v,1)
     padmode = check_padmode_kwarg(padmode, su, sv)
     if padmode == :longest
         if su < sv
-            u = [u;zeros(eltype(u),sv-su)]
+            u = _zeropad(u, sv, su)
         elseif sv < su
-            v = [v;zeros(eltype(v),su-sv)]
+            u = _zeropad(v, su, sv)
         end
         conv(u, Compat.reverse(conj(v), dims=1))
     elseif padmode == :none
--- a/src/dspbase.jl
+++ b/src/dspbase.jl
@@ -250,7 +250,7 @@ function xcorr(u, v; padmode::Symbol = :default_longest)
         if su < sv
             u = _zeropad(u, sv, su)
         elseif sv < su
-            u = _zeropad(v, su, sv)
+            v = _zeropad(v, su, sv)
         end
         conv(u, Compat.reverse(conj(v), dims=1))
     elseif padmode == :none
--- a/src/dspbase.jl
+++ b/src/dspbase.jl
@@ -207,14 +207,14 @@ conv2(A::StridedMatrix{T}, B::StridedMatrix{T}) where {T<:Integer} =
 conv2(u::StridedVector{T}, v::StridedVector{T}, A::StridedMatrix{T}) where {T<:Integer} =
     round.(Int, conv2(float(u), float(v), float(A)))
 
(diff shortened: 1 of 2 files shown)


HDictus:
You might want to have a look at the tests to make sure both modes are represented. You'll still have a problem with nightly, but the rest of the CI should go through.

galenlynch:
@HDictus Thanks for taking a look. Yeah I'll fix the tests if people think this is a good idea. I guess I didn't say so but I'm hoping for feedback on this PR. I think the change would be useful, and like the new default behavior better than having the extra zero padding end up in the results.

src/dspbase.jl
```diff
@@ -186,17 +186,45 @@ conv2(A::StridedMatrix{T}, B::StridedMatrix{T}) where {T<:Integer} =
...
+end
+
 """
-    xcorr(u,v)
+    xcorr(u,v; mode = :full)
 
-Compute the cross-correlation of two vectors.
+Compute the cross-correlation of two vectors. The size of the output depends on
+`mode` keyword argument: if `mode = :full` then the result will be the same size
+as [`conv`](@ref) of `u` and `v`. If `mode = :legacy` then the result will have
```
ssfrr: maybe specify here that the length along dimension `d` will be `size(u, d)+size(v, d)-1` for clarify? I guess it could even be `length` given that this currently only supports 1D (for now...)

src/dspbase.jl
```diff
@@ -186,17 +186,45 @@ conv2(A::StridedMatrix{T}, B::StridedMatrix{T}) where {T<:Integer} =
...
 """
-function xcorr(u, v)
+function xcorr(u, v; mode::Symbol = :default_full)
     su = size(u,1); sv = size(v,1)
-    if su < sv
-        u = [u;zeros(eltype(u),sv-su)]
-    elseif sv < su
-        v = [v;zeros(eltype(v),su-sv)]
+    mode = check_mode_kwarg(mode, su, sv)
+    if mode == :legacy
```
ssfrr: I agree that this mode shouldn't be the default, but I can see it being a useful feature (as you mentioned before, some folks might want the zero-lag condition to be centered). Maybe `padmode=:longest` or something, rather than calling it `:legacy`.

ssfrr:
Actually, on thinking about it a little bit more, we probably want the default behavior to throw the deprecation warning but keep the current behavior, then in a future release we can change the default.

So for `x` and `y` that are different lengths, you'd get:


- `xcorr(x, y; padmode=:none)` - get the new behavior
- `xcorr(x, y; padmode=:longest)` - get the old behavior
- `xcorr(x, y)` - throws depwarn that default `padmode` will be changing to `:none`, set `padmode = :longest`

galenlynch:
Yeah, I don't know what to call the different modes. I don't really like `:none` for the new beavior, because padding still happens in `conv`. In addition, no padding would correspond to circular cross-correlation, which could also be a supported mode down the road. The new behavior corresponds to matlab's `full` convolution. `:longest` seems to make sense for the old behavior.

ssfrr:
But that padding is an implementation detail because of the FFT-based convolution. Conceptually I’d still say that linear convolution or two discrete sequences will have support over the length N+M-1 if there is no padding.

I think circular convolution/correlation should be distinct operations from their linear counterparts, so in my mind the padding shouldn’t be related.

galenlynch:
Makes sense. I adopted your suggestions.

src/dspbase.jl
galenlynch: fixed

src/dspbase.jl
galenlynch: Does the new docstring address this?

src/dspbase.jl
```diff
@@ -207,17 +207,48 @@ conv2(A::StridedMatrix{T}, B::StridedMatrix{T}) where {T<:Integer} =
...
 """
-    xcorr(u,v)
+    xcorr(u,v; padmode = :longest)
 
-Compute the cross-correlation of two vectors.
+Compute the cross-correlation of two vectors. The size of the output depends on
+`padmode` keyword argument: if `padmode = :none` then the result will be the same size
+as [`conv`](@ref) of `u` and `v`. If `padmode = :longest` then the result will have
+length `2*max(length(X), length(Y))-1`, where the beginning of the result will
+be padded with zeros. If `padmode = :none`, it will be `length(X) - length(Y) - 1`.
```
ssfrr: The pre-padding or post-padding of the result depend on which of the arguments ends up getting padded (you can confirm in the tests below). So if the time-reversed argument (`v`) gets padded then the result is pre-padded, but if `u` gets padded then the result is post-padded.

also an editing nitpick, but I'd put the two clauses relating to `:none` together, so something like:

> Compute the cross-correlation of two vectors. The size of the output depends on
the `padmode` keyword argument: with `padmode = :none` the length of the
result will be  `length(u) + length(v) - 1`, as with [`conv`](@ref). With
`padmode = :longest` the shorter of the arguments will be padded so they
are equal length. This gives a result with length `2*max(length(u), length(v))-1`,
with the zero-lag condition at the center.

We should also add a note about the deprecation and upcoming change:
```
!!! warning
    The default value of `padmode` will be changing from `:longest` to `:none`
    in a future release of DSP. In preparation for this change, leaving `padmode` unspecified
    is currently deprecated.
```

src/dspbase.jl
```diff
@@ -207,17 +207,48 @@ conv2(A::StridedMatrix{T}, B::StridedMatrix{T}) where {T<:Integer} =
 conv2(u::StridedVector{T}, v::StridedVector{T}, A::StridedMatrix{T}) where {T<:Integer} =
     round.(Int, conv2(float(u), float(v), float(A)))
 
+function check_padmode_kwarg(padmode::Symbol, su::Integer, sv::Integer)
+    if padmode == :default_longest
+        if su != sv
+            Base.depwarn(
+            """
+            The default behavior for xcorr will changed in future versions.
```
ssfrr: couldn't hurt to be explicit here, and give a direct recommendation for what to do to keep things unchanged (for when you're updating old code): "the default pad mode of `xcorr` will be changed from `:longest` to `:none` in future versions. To keep current behavior specify `padmode=:longest`"

src/dspbase.jl
galenlynch: Whoops, didn't re-read the doc string before addressing your comment. Good catch!

src/dspbase.jl
galenlynch: done

galenlynch:
Yeah, once you start working with the different length inputs the old behaviour definitely bites. Now we just have to combine this code with that of `filt`, figure out an API for separable convolutions in multiple dimensions, merge #269, and then we'll be all set...

PR #273, opened 2019-04-02
Improve decision boundary between time domain and frequency filtering

galenlynch:
In my testing, `filt` is choosing to use frequency domain filtering in instances when time domain filtering would be faster. I have moved the decision boundary between the two algorithms to be better reflect benchmarking results from my laptop. To be honest, I'm not sure how well these benchmarks generalize.

In the process, I have also changed how `nfft` is chosen for the overlap-save algorithm. The new heuristic is simpler, and seems to follow the cost estimates provided by FFTW.

The choice between time and frequency domain filtering is still not perfect. Right now the decision seems to be correct when filtering large amounts of data. However, for smaller amounts of data, it may be advantageous to do time domain filtering to avoid the overhead of frequency domain filtering, even if it would be more expensive per output sample for large amounts of data.

### DIFF
--- a/src/Filters/filt.jl
+++ b/src/Filters/filt.jl
@@ -391,10 +391,11 @@ const SMALL_FILT_CUTOFF = 54
 #
 # filt implementation for FIR filters (faster than Base)
 #
-@generated function filt!(out, b::NTuple{N,T}, x) where {N, T}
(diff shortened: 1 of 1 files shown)


galenlynch:
Ok to merge?

Issue #262, opened 2019-03-13
Resample segmentation fault for rate::AbstractFloat


1oly:
Hello,

I'm getting a segmentation fault for resample when `rate` is a certain float value (example coming from an actual script where `s` and `y` are audio signals):
```
using DSP
s = rand(1920000);
y = rand(2629632);
rate = length(s)/length(y)
X = resample(y,rate)
```

by using Rational type, I don't see a segmentation fault: 

```
using DSP
s = rand(1920000);
y = rand(2629632);
rate = length(s)//length(y) # Rational here
X = resample(y,rate)
```

I tried to look in `resample_filter` https://github.com/JuliaDSP/DSP.jl/blob/270e6f25de3f9176493c3bfed863bf0ec94289ac/src/Filters/design.jl#L606
but cannot reproduce manually. 

Will be happy to help out if this is not an issue on my side.

Thanks!

Error: 
```
signal (11): Segmentation fault: 11
in expression starting at no file:0
setindex! at ./array.jl:767 [inlined]
filt! at /Users/oliver/.julia/packages/DSP/efNaR/src/Filters/stream_filt.jl:635
filt at /Users/oliver/.julia/packages/DSP/efNaR/src/Filters/stream_filt.jl:648
resample at /Users/oliver/.julia/packages/DSP/efNaR/src/Filters/stream_filt.jl:703
resample at /Users/oliver/.julia/packages/DSP/efNaR/src/Filters/stream_filt.jl:708
unknown function (ip: 0x11f37f318)
jl_fptr_trampoline at /Users/osx/buildbot/slave/package_osx64/build/src/gf.c:1864
do_call at /Users/osx/buildbot/slave/package_osx64/build/src/interpreter.c:323
eval_stmt_value at /Users/osx/buildbot/slave/package_osx64/build/src/interpreter.c:362 [inlined]
eval_body at /Users/osx/buildbot/slave/package_osx64/build/src/interpreter.c:759
jl_interpret_toplevel_thunk_callback at /Users/osx/buildbot/slave/package_osx64/build/src/interpreter.c:885
unknown function (ip: 0xfffffffffffffffe)
unknown function (ip: 0x10e2c7caf)
unknown function (ip: 0xffffffffffffffff)
jl_interpret_toplevel_thunk at /Users/osx/buildbot/slave/package_osx64/build/src/interpreter.c:894
[...]

PR #275, opened 2019-04-09
Check buffer lengths in stream filtering functions before access

galenlynch:
A number of exported functions in `stream_filt.jl` use `@inbounds` access to
buffers without verifying their lengths. This can cause segfaults, as seen in
issue #262.

### DIFF
--- a/src/Filters/stream_filt.jl
+++ b/src/Filters/stream_filt.jl
@@ -544,8 +544,6 @@ function filt!(buffer::AbstractVector{Tb}, self::FIRFilter{FIRDecimator{Th}}, x:
     history::Vector{Tx} = self.history
     bufIdx              = 0
 
-    bufLen >= xLen || error("buffer length must be >= x length")
-
     if xLen < kernel.inputDeficit
         self.history = shiftin!(history, x)
         kernel.inputDeficit -= xLen
@@ -554,6 +552,9 @@ function filt!(buffer::AbstractVector{Tb}, self::FIRFilter{FIRDecimator{Th}}, x:
 
     outLen              = outputlength(self, xLen)
     inputIdx            = kernel.inputDeficit
+
+    nbufout = fld(xLen - inputIdx, kernel.decimation)
+    bufLen >= nbufout || error("buffer length insufficient")
 
     while inputIdx <= xLen
         bufIdx += 1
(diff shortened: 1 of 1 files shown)


galenlynch:
Good to merge?

src/Filters/stream_filt.jl
```diff
@@ -552,6 +553,9 @@ function filt!(buffer::AbstractVector{Tb}, self::FIRFilter{FIRDecimator{Th}}, x:
     outLen              = outputlength(self, xLen)
     inputIdx            = kernel.inputDeficit
 
+    nbufout = fld(xLen - inputIdx, kernel.decimation)
+    bufLen >= nbufout || error("buffer length insufficient")
```
martinholters:
```suggestion
    bufLen >= nbufout || throw(ArgumentError("buffer length insufficient"))
```

src/Filters/stream_filt.jl
```diff
@@ -552,6 +553,9 @@ function filt!(buffer::AbstractVector{Tb}, self::FIRFilter{FIRDecimator{Th}}, x:
     outLen              = outputlength(self, xLen)
     inputIdx            = kernel.inputDeficit
 
+    nbufout = fld(xLen - inputIdx, kernel.decimation)
```
martinholters: If `xLen == inputIdx`, one value is stored to `buffer` below, so this can't be quite right. Might be just missing a `+ 1`, but I haven't really thought this through.

src/Filters/stream_filt.jl
```diff
@@ -600,10 +604,15 @@ function update(kernel::FIRArbitrary)
...
-function filt!(buffer::AbstractVector{Tb}, self::FIRFilter{FIRArbitrary{Th}}, x::AbstractVector{Tx}) where {Tb,Th,Tx}
+function filt!(
+    buffer::AbstractVector{Tb},
+    self::FIRFilter{FIRArbitrary{Th}},
+    x::AbstractVector{Tx}
+) where {Tb,Th,Tx}
     kernel              = self.kernel
     pfb                 = kernel.pfb
     dpfb                = kernel.dpfb
+    bufLen              = length(buffer)
```
martinholters:
```suggestion
```

galenlynch:
Thanks for the review!

PR #279, opened 2019-04-09
Abstract array conv

galenlynch:
This allows `conv` and friends to operate on arrays that don't start at index 1.
It also allows the use of other abstract vectors, such as ranges.

This makes it possible to convolutions with lots of different types of arrays:
```julia
julia> using DSP, OffsetArrays

julia> offset_arr = OffsetArray{Int}(undef, -1:2);

julia> offset_arr[:] = [1, 2, 1, 2];

julia> conv(offset_arr, 1:3)
6-element Array{Int64,1}:
  1
  4
  8
 10
  7
  6

```
It also fixes a problem with type promotion that existed prior to this commit. Previously, `conv` would fail for inputs with differing floating point element types. For example, using both `Float32` and `Float64` would cause method errors. This PR fixes this problem by adding an extra promotion method.

Finally, it allows other element types that are not BLAS floats or integers to be used. Due to the zero padding involved, and reuse of buffers for the different inputs, it is still necessary to have the core `conv` kernel operate on BLAS floats, but more types of inputs can now be accepted and converted to BLAS floats.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -3,3 +3,4 @@ Polynomials 0.1.0
 Reexport
 SpecialFunctions
 FFTW
+Compat 1.2
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,4 +1,4 @@
-julia 0.7
+julia 1
 Polynomials 0.1.0
 Reexport
 SpecialFunctions
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,4 +1,4 @@
-julia 1
+julia 0.7
 Polynomials 0.1.0
 Reexport
 SpecialFunctions
--- a/REQUIRE
+++ b/REQUIRE
@@ -3,4 +3,4 @@ Polynomials 0.1.0
 Reexport
 SpecialFunctions
 FFTW
-Compat 1.2
+Compat 1.3
--- a/src/dspbase.jl
+++ b/src/dspbase.jl
@@ -303,6 +303,16 @@ function check_padmode_kwarg(padmode::Symbol, su::Integer, sv::Integer)
     end
 end
 
+dsp_reverse(v, ::NTuple{<:Any, Base.OneTo{Int}}) = reverse(v, dims = 1)
+function dsp_reverse(v, vaxes)
+    vsize = length(v)
+    reflected_start = - first(vaxes[1]) - vsize + 1
+    reflected_axes = (range(reflected_start, reflected_start + vsize - 1),)
+    out = similar(v, reflected_axes)
+    copyto!(out, reflected_start, Iterators.reverse(v), 1, vsize)
+end
+
+
 """
     xcorr(u,v; padmode = :longest)
 
@@ -329,9 +339,9 @@ function xcorr(
         elseif sv < su
             v = _zeropad(v, su)
         end
-        conv(u, reverse(conj(v), dims=1))
+        conv(u, dsp_reverse(conj(v), axes(v)))
     elseif padmode == :none
-        conv(u, reverse(conj(v), dims=1))
+        conv(u, dsp_reverse(conj(v), axes(v)))
     else
         throw(ArgumentError("padmode keyword argument must be either :none or :longest"))
     end
--- a/src/dspbase.jl
+++ b/src/dspbase.jl
@@ -129,6 +129,7 @@ function _zeropad!(padded::AbstractVector, u::AbstractVector)
 end
 function _zeropad!(padded::AbstractArray{<:Any, N},
                    u::AbstractArray{<:Any, N}) where N
+    # Copy the data to the beginning of the padded array
     pad_ax = axes(padded)
     pad_first_I = first.(pad_ax)
     datasize = size(u)
@@ -138,14 +139,37 @@ function _zeropad!(padded::AbstractArray{<:Any, N},
     data_range = CartesianIndices(u)
     copyto!(padded, pad_data_range, u, data_range)
 
+    # Step through each dimension, and fill the trailing indices in that
+    # dimension with zeros.
+    #
+    # This is accomplished by going from the last to the first dimension, and
+    # for each dimension, finding the biggest rectangular region that needs to
+    # be zero padded. This region corresponds to the  trailing indices after the
+    # data in the selected dimension, all the indices in lower dimensions, and
+    # the same indices as the data for higher dimensions, corresponding to
+    # regions that were missed in that dimension. This should allow column-major
+    # friendly access to the array.
+    #
+    # Geometric intuition for two dimensions:
+    #   1 1 1 0 0     1 1 1    0 0
+    #   1 1 1 0 0     1 1 1    0 0
+    #   1 1 1 0 0  =  1 1 1    0 0
+    #   0 0 0 0 0              0 0
+    #   0 0 0 0 0     0 0 0    0 0
+    #                 0 0 0
+
     pad_ranges = Vector{UnitRange{Int}}(undef, N)
-    pad_ranges .= pad_ax
+    pad_ranges .= pad_ax # Initially select the entire dimension
     for i = N:-1:1
+        # Get the trailing indices for this dimension
         nu = size(u, i)
         first_pad_i = first(pad_ranges[i])
         pad_ranges[i] = first_pad_i + nu :last(pad_ranges[i])
+
+        # Make the rectangular region and set it to zero
         pad_range = CartesianIndices(NTuple{N, UnitRange{Int}}(pad_ranges))
         padded[pad_range] .= 0
+        # Set this dimension to be just the portion that was missed (same as data)
         pad_ranges[i] = first_pad_i: first_pad_i + nu - 1
     end
     padded
--- a/src/dspbase.jl
+++ b/src/dspbase.jl
@@ -118,31 +118,34 @@ function deconv(b::StridedVector{T}, a::StridedVector{T}) where T
 end
 
 function _zeropad!(padded::AbstractVector, u::AbstractVector)
-    ulen = length(u)
-    padlen = length(padded)
-    pad_inds = axes(padded, 1)
-    first_padded_i = first(pad_inds)
-    copyto!(padded, first_padded_i, u, first(axes(u, 1)), ulen)
-    padded[first_padded_i + ulen:last(pad_inds)] .= 0
+    u_len = length(u)
+    # Use axes to accommodate arrays that do not start at index 1
+    pad_ax = axes(padded, 1)
+    first_padded_i = first(pad_ax)
+    first_u_i = first(axes(u, 1))
+    copyto!(padded, first_padded_i, u, first_u_i, u_len)
+    padded[first_padded_i + u_len:last(pad_ax)] .= 0
     padded
 end
 function _zeropad!(padded::AbstractArray{<:Any, N},
                    u::AbstractArray{<:Any, N}) where N
-    padax = axes(padded)
-    padfirst = first.(padax)
+    pad_ax = axes(padded)
+    pad_first_I = first.(pad_ax)
     datasize = size(u)
-    pad_dest = CartesianIndices(range.(padfirst, padfirst .+ datasize .- 1))
-    datainds = CartesianIndices(u)
-    copyto!(padded, pad_dest, u, datainds)
+    pad_data_range = CartesianIndices(
+        range.(pad_first_I, pad_first_I .+ datasize .- 1)
+    )
+    data_range = CartesianIndices(u)
+    copyto!(padded, pad_data_range, u, data_range)
 
     pad_ranges = Vector{UnitRange{Int}}(undef, N)
-    pad_ranges .= padax
+    pad_ranges .= pad_ax
     for i = N:-1:1
         nu = size(u, i)
         first_pad_i = first(pad_ranges[i])
         pad_ranges[i] = first_pad_i + nu :last(pad_ranges[i])
-        padinds = CartesianIndices(NTuple{N, UnitRange{Int}}(pad_ranges))
-        padded[padinds] .= 0
+        pad_range = CartesianIndices(NTuple{N, UnitRange{Int}}(pad_ranges))
+        padded[pad_range] .= 0
         pad_ranges[i] = first_pad_i: first_pad_i + nu - 1
     end
     padded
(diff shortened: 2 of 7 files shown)


galenlynch:
This would implement #228.

martinholters:
So if I understand correctly, this just ignores the offset for `OffsetArray`s? If we later decide that returning an `OffsetArray` in that case would be preferable, that would be a breaking change.

galenlynch:
I think that's right. It uses `similar`, but I think that does not maintain the indexing of `OffsetArray`s.

galenlynch:
However, `_zeropad!` should work for `OffsetArray`s as is, which means it would be pretty easy to return a `OffsetArray` down the road.

galenlynch:
Reading over #228 in more detail, I agree that this behavior probably isn't the best. Should we just throw an error for offset arrays like you suggested?

galenlynch:
It feels a little weird to throw an error when a reasonable result could instead be produced, in order to leave room for future expansion of the API. Just because we can handle `OffsetArrays` doesn't mean we're obliged to maintain the indexing offset of the result do we? We don't even use `OffsetArrays` in DSP.jl, so I think it would be fine to change undefined behavior.

martinholters:
In my opinion, we should either
* not care about `OffsetArray`s; if something works for an `OffsetArray`, ok, if it errors, so be it, or
* explicitly ensure we're not given an `OffsetArray` and throw an error otherwise or
* figure out exactly what to do with `OffsetArray`s, but I'd propose to postpone that.

What I would not like is something half-baked, which goes through some trouble to ensure it does not error for an `OffsetArray` but does not have well thought-through semantics.

E.g.
> `_zeropad!` should work for `OffsetArray`s as is

If I read correctly, `_zeropad!` will copy `u` to the first `length(u)` indices of `padded` (talking about the vector case for simplicity, here) and set the rest to zero. Ok, sounds reasonable.  But we could also copy `u` to `padded[axes(u,1)]` (and set the rest to zero), so that `padded[i] == u[i]` would hold for `i` ∈ `axes(u,1)`. Sounds reasonable to me, too. And I have no idea which one we should pick in the long run. So should we commit to one behavior now?

galenlynch:
Just to be clear, the goal here is linear convolution. All this zero padding is completely internal and just allows you to do linear convolution using circular convolution machinery. This is just to say that where you put the zeros is an implementation detail, and should not depend on what offset the inputs have, nor should it change the result. The only meaningful difference in semantics that I can think of when working with `OffsetArray`s is what offset index the resulting vector containing the linear convolution should have, and perhaps implying the existence of more zeros than are actually in the inputs. As it is now, it will do the linear convolution correctly, and return a normal array, instead of an `OffsetArray`. Down the road, it could easily return an `OffsetArray` with some other offset. Even we did return an array with a offset, I think it would have the same content as the current implementation.

>  But we could also copy u to padded[axes(u,1)] (and set the rest to zero), so that padded[i] == u[i] would hold for i ∈ axes(u,1).

For this to work you would have to do the same thing, with the same location of zero padding, to the other input whether or not it is also an `OffsetArray`. You would also have to change which portion of the circular convolution you return to the user as the valid linear convolution result. All that would just get you the same linear convolution as copying `u` to the first `length(u)` indices and calling it a day.

The only way I can think of an `OffsetArray` changing the interpretation of what it means to convolve an array with some other array is that it might imply the existence of more zeros than are actually in the inputs. For example, if you're convolving a vector that starts at 0, with another that starts at 1, maybe the interpretation should be there's an extra zero before the start of the second vector. However, I think that naive semantics of convolving the two arrays, without inferring more data than is actually in the inputs, and finally returning a regular vector, is pretty reasonable.

galenlynch:
> You would also have to change which portion of the circular convolution you return to the user as the valid linear convolution result. 

Actually this part isn't true. Moving the zeros around wouldn't change the portion of the circular convolution that corresponds to linear convolution.

martinholters:
Right, of course, `_zeropad!` is purely internal and so we can do what we want, and of course, the only difference between having `OffsetArray` inputs to convolution and other array inputs should be the indices of the result. I guess what I tried to say above is that if we at some point decide what to do about the output indices for `OffsetArray` inputs, it is unclear (at least to me ATM) whether the `_zeropad!` behavior needs to be adjusted.

> Just because we can handle `OffsetArray`s doesn't mean we're obliged to maintain the indexing offset of the result do we?

Well, changing the result (in a potentially subtle way) can always cause trouble.

> We don't even use OffsetArrays in DSP.jl, so I think it would be fine to change undefined behavior.

Ok, but then we should clearly document that it is undefined behavior and should not be used. Modifying our implementation so that it does not throw an error, but instead returns an arbitrary result, and at the same time add documentation telling people not to use that seems weird. Now, of course the result is not arbitrary, which means people  will start to rely on it, no matter how clearly we document they shouldn't. That might put us into a situation where we need to go through some deprecation cycle in the future. I'd like to avoid that.

galenlynch:
That makes sense. The implementation of `_zeropad!` and the indexing logic would not need to change, regardless of the offset of the inputs or the padded buffer. Right now the indexing offset of the result is determined by `_zeropad(u, size) = _zeropad!(similar(u, size), u)`, which will return an array with 1-based indexing. To make the result have some other indexing offset, all you would need to do is change that to `_zeropad!(similar(u, offset:offset + size - 1), u)`. 

Maybe we could just decide the behavior we want right now before we open the type bound up to include all abstract arrays? I don't know what a sensible default would be for two vectors that have different offsets. If they had the same offset, then it seems like maintaining that offset would make sense. However, always returning a 1-based vector would also make sense, as the user could just convert it to a indexing offset of their choosing afterwards. Returning regular arrays would also avoid adding another dependency to DSP.jl. It seems like ND cases would just be easy to extrapolate from what we decide with vectors.

ssfrr:
I'd expect the `offset(conv(a, b))` to be `offset(a)+offset(b)+1`

So if you have an array indexed from `0:4` and another from `0:6`, the result would be indexed from `0:10`. Shifting either input shifts the convolved output. This is consistent with treating OffsetArrays as conceptually zero outside the indexed range.

So for offsets of 2 and -3 it would look like this:

image

galenlynch:
> This is consistent with treating OffsetArrays as conceptually zero outside the indexed range.

But only the offset would change, not the result of the convolution, right?

ssfrr:
yes, that's correct, it should still be true that `result.parent == conv(a.parent, b.parent)`

galenlynch:
> I'd expect the `offset(conv(a, b))` to be `offset(a)+offset(b)+1`

I'm assuming you mean the offset is the difference from one, not the start of the indexing.
So if both inputs started at index 2 (offset = 1), then the output would start at index 3?

ssfrr:
also correct. I'm talking about the number stored in `a.offsets`, which is the offset from `1`.

There's also a little wrinkle with `xcorr`. By default `reverse(a::OffsetArray)` reverses the data but keeps the indices the same. In our context it would make sense for the indices to be reflected across 0 as well. This has the nice property that you can interpret the indices of the result as the lag.
For example:

```julia
using Plots
using OffsetArrays
using DSP

# these need to be generalized for N-D
# define a reverse that also reverses the indices
rev(a::OffsetArray) = OffsetArray(reverse(a.parent), -a.offsets[1]-length(a)-1)
domain(a::OffsetArray) = (1:length(a)).+a.offsets[1]
DSP.conv(a::OffsetArray, b::OffsetArray) = OffsetArray(conv(a.parent, b.parent),
                                                    a.offsets.+b.offsets.+1)

a1 = OffsetArray(rand(4), -2)
a2 = OffsetArray(a1.parent, 0) # delayed by 2
c = conv(rev(a1), a2)

xlims = (minimum(first.([domain(a1), domain(a2), domain(c)])) - 1,
         maximum(last.([domain(a1), domain(a2), domain(c)])) + 1)

sticks([domain(a1), domain(a2), domain(c)],
       [a1.parent, a2.parent, c.parent],
       xlims=xlims, layout=(3,1),
       xticks=[false false true],
       legend=false,
       marker=:circle, markersize=5,
       color=1)
```

Gives:
image

galenlynch:
Your rule, which I agree with 100%, is basically `startindex_out = startindex_u + startindex_v`, but should only apply if the arrays have weird axes. I don't think our interface should actually use the `AbstractArrays` package, and stick with what's defined in base. I'll see what I can cook up.

galenlynch:
Ok, that should do it...

galenlynch:
Tests are failing on 0.7 for cryptic reasons. Can we just drop support for 0.7?

galenlynch:
I see, it's failing because `range` changed from 0.7 to 1.0.

src/dspbase.jl
```diff
@@ -117,24 +117,58 @@ function deconv(b::StridedVector{T}, a::StridedVector{T}) where T
...
-        pad_ranges[i] = 1:nu
+        # Trailing indices for this dimension
+        pad_ranges[i] = datasize[i] + 1 : padsize[i]
+
+        # Make the rectangular region and set it to zero
+        pad_region = CartesianIndices(NTuple{N, UnitRange{Int}}(pad_ranges))
+        padded[pad_region] .= 0
+
+        # Set this dimension to be just the portion that was missed (same as data)
+        pad_ranges[i] = pad_data_ranges[i]
```
ssfrr: Zeroing out large contiguous blocks of memory should be pretty fast, and when `u` and `v` are similarly sized and N-dimensional only `1/2^N` of the entries are going to be nonzero. I did a little benchmarking against a version that just prefills `padded` with zeros:

```julia
using BenchmarkTools
using DSP

function zeropad_prefill!(padded::AbstractArray{<:Any, N},
                          u::AbstractArray{<:Any, N}) where N
    # Copy the data to the beginning of the padded array
    fill!(padded, zero(eltype(padded)))
    datasize = size(u)
    pad_data_ranges = range.(1, datasize)
    copyto!(padded, CartesianIndices(pad_data_ranges), u, CartesianIndices(u))
    
    padded
end

padded = rand(1000, 1000)
u = rand(500, 500)

display(@benchmark DSP._zeropad!(padded, u))
display(@benchmark zeropad_prefill!(padded, u))
```

On my system the `prefill` version is faster until `u` gets to be about `900x900`

src/dspbase.jl
```diff
@@ -163,37 +197,63 @@ function _conv(u, v, paddims)
...
-
-conv(u::StridedArray{T, N}, v::StridedArray{T, N}) where {T<:Integer, N} =
+function conv(u::AbstractArray{<:BLAS.BlasFloat, N},
+              v::AbstractArray{<:BLAS.BlasFloat, N}) where N
+    fu, fv = promote(u, v)
+    conv(fu, fv)
+end
+conv(u::AbstractArray{T, N}, v::AbstractArray{T, N}) where {T<:Number, N} =
+    conv(float(u), float(v))
+conv(u::AbstractArray{T, N}, v::AbstractArray{T, N}) where {T<:Integer, N} =
```
ssfrr: This should probably be expanded to have `T1<:Integer` and `T2<:Integer` so it covers when `u` and `v` are both `Integer` types but they're not the same type. (or maybe it should be handled with promotion?) For example this currently throws a `MethodError`:

```
conv([1, 2, 5], Int32[6, 1, 4, 1])
```

martinholters:
> I see, it's failing because `range` changed from 0.7 to 1.0.

Ah, careful! The `range(start, stop)` method you are using was added in julia 1.1. The reason it works on 1.0 here is that Compat supports it, and Compat is loaded by at least one of our dependencies. If we want to continue using this, I suggest adding Compat back in here, to be sure DSP still works even in case all our dependencies get rid of Compat. Alternatively, use `range(start, stop=stop)`. That would also work on julia 0.7.

REQUIRE
```diff
@@ -3,3 +3,4 @@ Polynomials 0.1.0
 Reexport
 SpecialFunctions
 FFTW
+Compat 1.2
```
martinholters:
```suggestion
Compat 1.3
```
Relevant change is
https://github.com/JuliaLang/Compat.jl/commit/052636a58dcffc3c810d48bc3818293dde924902

galenlynch: Thanks for catching that, and sorry that my PRs have been a little sloppy! I should really be working on other stuff at the moment, so I might be in a bit too much of a rush to finish this stuff up.

martinholters: Sorry if my comments are rather terse and might therefore sound a bit harsh sometimes. I should also be doing other stuff...

galenlynch: Not at all, they are appreciated! At least I'm in good company while not doing what I really should be doing.

galenlynch:
Good to merge?

ssfrr:
There are still some `range(x, y)` calls that are throwing a bunch of warnings

See the CI output here:
https://travis-ci.org/JuliaDSP/DSP.jl/jobs/518729461#L583

Unfortunately they don't show a useful backtrace, but it's probably [here](https://github.com/JuliaDSP/DSP.jl/pull/279/files#diff-be18884b6f4046c11d8320c86380b90cR136) and [here](https://github.com/JuliaDSP/DSP.jl/pull/279/files#diff-be18884b6f4046c11d8320c86380b90cR185) and [here](https://github.com/JuliaDSP/DSP.jl/pull/279/files#diff-be18884b6f4046c11d8320c86380b90cR187).

On 1.1 without Compat `range(a, b)` throws an error.

galenlynch:
I think I fixed the warnings...

ssfrr:
I think the `Compat` errors are because now there's a `Project.toml`, so `Pkg` ignores the `REQUIRE`. Unfortunately until we retire `REQUIRE` for good I think we need to keep the two files in sync.

galenlynch:
But there's no `Project.toml` in the commit being tested, right? The PR hasn't been merged with master since `Project.toml` was added.

galenlynch:
I find the new Pkg to be pretty confusing. Is there a tool to edit `Project.toml` and add dependencies? What about dependencies at test time only? Even finding the UUID of other packages is cryptic. I understand the intent/design of Pkg, but not how it's intended to be used.

ssfrr:
Hmm, good point. Maybe our Travis.yml is doing something fishy and for some reason it's using the Project.toml from `master`? Anyways looks like it's good now that you merged. :)

So each package's `Project.toml` can be activated to put yourself in the environment of that package. I think the easiest way to add dependencies to a package is to activate that package (`pkg> activate DSP`) and then do `pkg> add Compat`. That should add it as a dependency and handle adding the UUID.

martinholters:
Travis runs against a merge commit between the PRs branch and master, hence it had picked up Project.toml.

ssfrr:
ah, of course. that makes sense. I guess otherwise you wouldn't know that your merged code was good.

PR #281, opened 2019-04-12
Minor deprecation fixes

martinholters:
* Use `@deprecate_binding` for the old filter coefficient names and stop using them in tests.
* Drop the deprecation for `firfilt` (deprecated in v0.0.9!)
* Use `xcorr` with `padmode` argument in `finddelay` (apparently missed in #266).

### DIFF
--- a/src/deprecated.jl
+++ b/src/deprecated.jl
@@ -1,8 +1,7 @@
-const ZPKFilter = ZeroPoleGain
-const TFFilter = PolynomialRatio
-const BiquadFilter = Biquad
-const SOSFilter = SecondOrderSections
-export ZPKFilter, TFFilter, BiquadFilter, SOSFilter
+Base.@deprecate_binding ZPKFilter ZeroPoleGain
+Base.@deprecate_binding TFFilter PolynomialRatio
+Base.@deprecate_binding BiquadFilter Biquad
+Base.@deprecate_binding SOSFilter SecondOrderSections
 
 @deprecate firfilt(h, x) filt(h, x)
 @deprecate conv2 conv
--- a/src/deprecated.jl
+++ b/src/deprecated.jl
@@ -1,7 +1,7 @@
+# deprecations after 0.5
 Base.@deprecate_binding ZPKFilter ZeroPoleGain
 Base.@deprecate_binding TFFilter PolynomialRatio
 Base.@deprecate_binding BiquadFilter Biquad
 Base.@deprecate_binding SOSFilter SecondOrderSections
 
-@deprecate firfilt(h, x) filt(h, x)
 @deprecate conv2 conv
--- a/src/util.jl
+++ b/src/util.jl
@@ -361,11 +361,11 @@ julia> finddelay([1, 2, 3], [0, 0, 1, 2, 3])
 ```
 """
 function finddelay(x::AbstractVector{<: Real}, y::AbstractVector{<: Real})
-    s = xcorr(y, x)
+    s = xcorr(y, x, padmode=:none)
     max_corr = maximum(abs, s)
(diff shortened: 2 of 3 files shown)


galenlynch:
Whoops, good catch on `finddelay` without `padmode`

src/deprecated.jl
```diff
@@ -1,8 +1,7 @@
-const ZPKFilter = ZeroPoleGain
-const TFFilter = PolynomialRatio
-const BiquadFilter = Biquad
-const SOSFilter = SecondOrderSections
-export ZPKFilter, TFFilter, BiquadFilter, SOSFilter
+# deprecations after 0.5
+Base.@deprecate_binding ZPKFilter ZeroPoleGain
```
galenlynch: What's the difference between `@deprecate` and `@deprecate_binding`?

src/util.jl
```diff
@@ -361,11 +361,11 @@ julia> finddelay([1, 2, 3], [0, 0, 1, 2, 3])
 ```
 """
 function finddelay(x::AbstractVector{<: Real}, y::AbstractVector{<: Real})
-    s = xcorr(y, x)
+    s = xcorr(y, x, padmode=:none)
```
galenlynch: I think the safest thing would be to use `padmode=:longest`, which maintains the behavior that this function assumed.

src/util.jl
```diff
@@ -361,11 +361,11 @@ julia> finddelay([1, 2, 3], [0, 0, 1, 2, 3])
...
     max_corr = maximum(abs, s)
     max_idxs = findall(x -> abs(x) == max_corr, s)
 
-    center_idx = cld(length(s), 2)
```
galenlynch: I don't think this changed with padding

src/deprecated.jl
martinholters: `@deprecate` only works for functions. I.e. after `@deprecate ZPKFilter ZeroPoleGain`, one could call `ZPKFilter(z, p, k)`, but could not use it as a type as in `f::ZPKFilter`.

src/util.jl
martinholters: Well, it shifts the reference point, but I actually find this version a bit simpler.

src/util.jl
martinholters: The `:longest` to `:none` change requires this. Compare:
```julia
julia> xcorr([0, 0, 1, 2, 3], [1, 2, 3], padmode=:longest)
9-element Array{Int64,1}:
  0
  0
  0
  0
  3
  8
 14
  8
  3
```
Maximum is two indices after the middle => delay of two.
```julia
julia> xcorr([0, 0, 1, 2, 3], [1, 2, 3], padmode=:none)
7-element Array{Int64,1}:
  0
  0
  3
  8
 14
  8
  3
```
Maximum is two indices after the length of the second argument => delay of two.

If you insist, we can change the `padmode` to `:longest` and keep this as is. But I'm more used to the `padmode=:none` behavior.

galenlynch: Oh never mind, I get it now.

src/util.jl
galenlynch: Just noticed you were taking the length of `x` below, not `s`. Yeah that is simpler!

src/util.jl
galenlynch: Whoops, I was typing that as you were also t yping. Thanks for the explanation!

src/deprecated.jl
galenlynch: Ahh ok

ssfrr:
Just wanted to say I'm really pumped about how much activity DSP is seeing lately. It's really improving quickly. Even though (maybe because?) we should all be doing other things. :smile:

galenlynch:
Definitely because

PR #282, opened 2019-04-12
Add Project.toml

martinholters:
We'll need this for future version registrations.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,0 +1,22 @@
+name = "DSP"
+uuid = "717857b8-e6f2-59f4-9121-6e50c889abd2"
+
+[deps]
+FFTW = "7a1cc6ca-52ef-59f5-83cd-3a7055c09341"
+LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
+Polynomials = "f27b6e38-b328-58d1-80ce-0feddd5e7a45"
+Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
+Reexport = "189a3867-3050-52da-a836-e630ba90ab69"
+SpecialFunctions = "276daf66-3868-5448-9aa4-cd146d93841b"
+Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
+
+[compat]
+Polynomials = ">= 0.1.0"
+julia = "0.7,1"
+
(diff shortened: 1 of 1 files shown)


galenlynch:
Thanks for doing this!

Project.toml
```diff
@@ -0,0 +1,22 @@
...
+FFTW = "7a1cc6ca-52ef-59f5-83cd-3a7055c09341"
+LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
+Polynomials = "f27b6e38-b328-58d1-80ce-0feddd5e7a45"
+Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
+Reexport = "189a3867-3050-52da-a836-e630ba90ab69"
+SpecialFunctions = "276daf66-3868-5448-9aa4-cd146d93841b"
+Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
+
+[compat]
+Polynomials = "≥ 0.1.0"
```
galenlynch: I can't find the relevant documentation in [Pkg's compatibility section](https://julialang.github.io/Pkg.jl/v1/compatibility/) for `≥`. Is it the same as a caret?

Project.toml
```diff
@@ -0,0 +1,22 @@
+name = "DSP"
+uuid = "717857b8-e6f2-59f4-9121-6e50c889abd2"
+
```
galenlynch: Don't we need a version?

Project.toml
martinholters: https://julialang.github.io/Pkg.jl/v1/compatibility/#Inequality-specifiers-1

Project.toml
martinholters: Good catch, we do! (Well, at least once we actually want to tag the next version, but no harm in setting it to the current version now.)

Project.toml
galenlynch: Ah gotcha. Could we use the ASCII version? I haven't gotten around to integrating latex completions into my editor, and I feel like other people may be in a similar situation. Unicode symbols can be a real drag to work with if you're not using the right editor.

martinholters: Sure. This is what the automatic script produces.

galenlynch:
Thanks again for doing this!

PR #283, opened 2019-04-12
Make ESPRIT test more robust

martinholters:
AFAICT ESPRIT doesn't guarantee a specific order for the frequencies  returned, so sort them before comparing to the expected ones.


martinholters:
Ah, passes on nightly :tada:

galenlynch:
Thanks for finally fixing this!

PR #284, opened 2019-04-12
Rework filter coefficients

### LINKED ISSUE #170: Inconsistencies for filters with different number of poles and zeros
I think for discrete filters, it is unclear whether powers of z or powers of z⁻¹ are being used. If numerator and denominator have the same order, that doesn't matter, but otherwise this gives rise to some inconsistencies:
```julia
julia> h = Biquad(0., 1., 0., 0., 0.); # just a unit delay z⁻¹

julia> impz(h, 3) # OK
3-element Array{Float64,1}:
 0.0
 1.0
 0.0

julia> h2 = PolynomialRatio(h)
DSP.Filters.PolynomialRatio{Float64}(Poly(1.0), Poly(1.0*x))

julia> impz(h2, 3) # Oh oh
3-element Array{Float64,1}:
 1.0
 0.0
 0.0

julia> freqz(h, linspace(0, pi)) ≈ freqz(h2, linspace(0, pi)) ≈ exp.(linspace(0, pi)*im) # here it's treated like z instead of z⁻¹
true
```

Looks like the last one to make any major changes to the involved functions was @simonster three years ago. Any opinions here?

### LINKED ISSUE #310: phasez returns negative phase
For example:
```julia
julia> angle.(freqz(digitalfilter(Lowpass(.5), Butterworth(2))))
250-element Array{Float64,1}:
  0.0
 -0.008921688379769978
 -0.017844797030280853
  ⋮
 -3.123747856559511
 -3.132670965210026
 -3.141592653589793

julia> phasez(digitalfilter(Lowpass(.5), Butterworth(2)))
250-element Array{Float64,1}:
 -0.0
  0.008921688379769978
  0.017844797030280853
  ⋮
  3.123747856559511
  3.132670965210026
  3.141592653589793
```
In this case `phasez` gives back phase that rises with frequency which is associated with negative group delay that should not appear with any causal filter. 

The function definition states:
```julia
function phasez(filter::FilterCoefficients, w = range(0, stop=π, length=250))
    h = freqz(filter, w)
    unwrap(-atan.(imag(h), real(h)); dims=ndims(h))
end
```
and I'm not sure why that minus is before the tangent. Is there any reason for this or should I try to put in my first ever PR with an easy fix?

martinholters:
This a start on adding a `Domain` type parameter on `FilterCoefficients` and its subtypes, as discussed in #170. I've chosen `:z` and `:s` instead of a tag type as proposed in https://github.com/JuliaDSP/DSP.jl/issues/170#issuecomment-462582469 for no particular reason. If a tag type is generally preferred, I'm happy to change it.

Instead of defining the polynomials for `PolynomialRatio{:z}` in terms of z⁻¹, I've kept (?) them at z, but zero-pad numerator and denominator to same length as necessary. That suffices to fix #170, but re-defining them to be in z⁻¹ might be more intuitive going forward.

For `PolynomialRatio{:s}`, I've dropped the coefficient normalization as it doesn't make as much sense there as it does for digital filters. I haven't touched upon coefficient normalization for `PolynomialRatio{:z}` for now, but have it on my to-do.

One issue I have discovered is that `Biquad` doesn't work well for analog filters due to the implicit `a0==1`, corresponding to a s² term always being present in the denominator. Hence, we can't represent e.g. s or s², although they clearly are (degenerate) biquads. Not sure what to do about this. Include `a0` and keep it at one for digital filters (similar to `PolynomialRatio{:z}`)?

That said, this branch should be fairly usable, but due to the open questions, I keep it as draft for now.

EDIT: Also fixes #310.


hamzaelsaawy:
I remember starting to implement this but it seems I forgot to follow through.
I did change the printing in `Polynomials.jl` to prep, because I was thinking of toggling the `reverse` in `PolynomialRatio` for discrete vs analog.

That way `Poly` eats up "leading" zeros in the s-domain and trailing zeros in the z-domain, which I thought was more intuitive.

martinholters:
I'll try not reversing for digital and see how that feels.

martinholters:
Yes, I think having the polynomials of `PolynomialRatio{:z}` be defined in terms of z⁻¹ is nicer. The conversion to/from `ZeroPoleGain` is a bit more complicated now. At the moment, we interpret `ZeroPoleGain{:z}` as `k*prod(z-z₀ᵢ)/prod(z-pᵢ)`. It might make things a little easier if we treat it as `k*prod(1-z₀ᵢ*z⁻¹)/prod(1-pᵢ*z⁻¹)` instead (which only makes a difference if the number of zeros and poles differ). I'll give that a shot...

martinholters:
I've tackled #142 in the latest change. Behavior now:
```julia
julia> PolynomialRatio{:z}([1, 2], [3, 4]) # normalizes and promotes to float
PolynomialRatio{:z,Float64}(Poly(0.3333333333333333 + 0.6666666666666666*z^-1), Poly(1.0 + 1.3333333333333333*z^-1))

julia> PolynomialRatio{:z,Int}([1, 2], [1, 4]) # does not normalize/promote
PolynomialRatio{:z,Int64}(Poly(1 + 2*z^-1), Poly(1 + 4*z^-1))

julia> PolynomialRatio{:z,Int}([1, 2], [3, 4]) # ...but asserts the coefficients are normalized
ERROR: ArgumentError: filter must have denominator with leading coefficient equal to one

julia> PolynomialRatio{:z,Float64}([1.0, 2.0], [3.0, 4.0]) # ...also if normalizing could be done within the type
ERROR: ArgumentError: filter must have denominator with leading coefficient equal to one
```
(stacktraces omitted for brevity)

That is normalization is done if no element type is specified; otherwise the coefficients are verified to be normalized and an error is thrown otherwise. Alternatively, one could try to normalize and only throw if that doesn't work, i.e. the normalized coefficients can't be converted to the given element type. Would that be preferable?

hamzaelsaawy:
I vote for a normalize error if it cant work, like `Int(3.0)` vs `Int(3.1)`.

martinholters:
Ok, I've changed the failing cases from above to
```julia
julia> PolynomialRatio{:z,Int}([2, 2], [2, 4]) # normalizes, keeps Int
PolynomialRatio{:z,Int64}(Poly(1 + z^-1), Poly(1 + 2*z^-1))

julia> PolynomialRatio{:z,Float64}([1.0, 2.0], [3.0, 4.0]) # normalizes, no problem
PolynomialRatio{:z,Float64}(Poly(0.3333333333333333 + 0.6666666666666666*z^-1), Poly(1.0 + 1.3333333333333333*z^-1))

julia> PolynomialRatio{:z,Int}([1, 2], [3, 4]) # fails because 1/3 cannot be represented as an Int
ERROR: InexactError: Int64(0.3333333333333333)

julia> PolynomialRatio{:z,Float64}([1.0, 2.0], [0.0, 4.0]) # fails because normalization would need to divide be zero
ERROR: ArgumentError: filter must have non-zero leading denominator coefficient

**Comment by martinholters** [comment]
I think this is (and has been for a while...) in a state where it could use some serious review. Is this the API we want?

**Comment by galenlynch** [comment]
It seems like a major improvement to me. I'll try to review it soon but could probably use some help.

**Comment by galenlynch** [comment]
@martinholters I totally dropped the ball on this. I'm assuming you're still interested in this PR? I'll try to look at it soon!

**Comment by martinholters** [comment]
Oh yes, would be nice to move this finally forward...

**Comment by kumarbal** [comment]
@martinholters Is it possible to add an evaluation routine for `PolynomialRatio` as in evaluate R(z), where z could be e.g. `im \omega`? This is in line with a similar function in Polynomial.

**Comment by galenlynch** [comment]
We need to get this out of purgatory! Martin I will try to review it soon, but realistically I won't be able to before Sunday. Given how important it is, and how long it's been sitting in purgatory, maybe someone else could also help out?

**Comment by galenlynch** [comment]
I'm for the API change.

**Comment by kumarbal** [comment]
I am up to my neck with the day job, but I can probably help with review
and testing over time. Sorry I cannot spare more. I think the discussion is
easy to translate into a design template. I must admit I did not consider
all the ramifications of how the changes will affect behavior of existing
functionality.

Regards

**Comment by martinholters** [comment]
Rebased and CI is still green. The feedback so far, if little, has been generally in favor, so I think I'll go ahead and merge this in a few day unless anyone objects. I guess we'll have to iterate on it a bit in the future, but even in the current state I'm pretty confident it's an improvement, and having it on master seems the only way to get broader feedback and figure out the rough edges.

**Comment by galenlynch** [comment]
That sounds great!

File: src/Filters/coefficients.jl [hunk 1]
```diff
@@ -63,52 +83,101 @@ H(z) = \\frac{\\verb!b[1]! + \\ldots + \\verb!b[n]! z^{-n+1}}{\\verb!a[1]! + \\l
...
+end
+PolynomialRatio{D}(f::ZeroPoleGain{D,Z,P,K}) where {D,Z,P,K} =
+    PolynomialRatio{D,promote_type(real(Z),real(P),K)}(f)
+
+ZeroPoleGain{D,Z,P,K}(f::PolynomialRatio{D}) where {D,Z,P,K} =
+    ZeroPoleGain{D,Z,P,K}(ZeroPoleGain{D}(f))
+function ZeroPoleGain{:s}(f::PolynomialRatio{:s,T}) where {T}
+    z = roots(f.b)
+    p = roots(f.a)
+    # work around `roots` returning Any[] for polynomials of degree zero
```
martinholters, line 137: Ref. https://github.com/JuliaMath/Polynomials.jl/pull/286. If that gets accepted, we could remove this work-around if we bump the required `Polynomials` version.

galenlynch:
I know I keep saying this but I'll really try to review it in the next
three days

martinholters:
I have rewritten `PolynomialRatio` to use `LaurentPolynomial`s. That allows to represent the z-case as Polynomials with negative powers of z instead of positive powers of z^-1 and simplifies a lot of things. Unfortunately, it relies on https://github.com/JuliaMath/Polynomials.jl/issues/287 getting fixed, so I haven't pushed it, yet. It's probably worth waiting for that.

martinholters:
Pushed the updated version. This will fail CI because it depends on the as-of-now untagged version 1.1.11 of Polynomials. But for testing locally, `dev`ing Polynomials should do the trick.

So my current plan is to wait for the Polynomials release, retrigger CI, if if that comes out green, I'll merge unless anyone objects.

martinholters:
...speaking of versions, I've bumped the DSP version to `0.7.0-dev` in the latest commit, as I think this the PR warrants a minor version bump (being somewhat break), and this way, we won't forget.

martinholters:
OMG, I finally did it. :boom:

galenlynch:
THANK YOU!! This is huge!

PR #297, opened 2019-05-08
Combine FIR filtering from dspbase and DSP.jl

galenlynch:
There were two identical functions, `_tdfilt!`, and `filt!` that both accomplished
FIR filtering. It turns out that the function in dspbase, `filt!` was faster,
and also allowed you to change the initial filter state to achieve, e.g., a steady state
step response. I therefore removed the implementation in `_tdfilt!` in favor of
`filt!`.

The one benefit that `_tdfilt!` had over `filt!` was a set of small-kernel
filtering functions with an unrolled loop. I have combined the functionality of
these small filter functions with `filt!`, while making these small filter
functions also accept some initial state.

Finally, I have added `muladd` in the direct form filtering kernels, where
possible. This makes a big performance difference for the unrolled loop kernel,
but for the other kernels it doesn't seem to matter much. However, it should
produce more accurate filtering results on platforms that support fma
instructions.

### DIFF
--- a/src/Filters/Filters.jl
+++ b/src/Filters/Filters.jl
@@ -4,7 +4,7 @@ using Polynomials, ..Util
 import Base: *
 using LinearAlgebra: I, mul!, rmul!
 using Statistics: middle
-import ..DSP: filt, filt!
+import ..DSP: filt, filt!, SMALL_FILT_CUTOFF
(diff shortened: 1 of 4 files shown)


galenlynch:
Ok to merge?

PR #300, opened 2019-05-15
Accommodate AbstractFFTs.jl#26 fftfreq etc.

crstnbr:
This PR accommodates https://github.com/JuliaMath/AbstractFFTs.jl/pull/26.

The definitions for `fftfreq`, `rfftfreq`, as well as the `Frequencies` type has been moved to AbstractFFTs.jl. Hence these definitions aren't necessary in DSP.jl anymore.

All tests pass locally. They will fail here, since https://github.com/JuliaMath/AbstractFFTs.jl/pull/26 hasn't been merged yet.

Do not merge before https://github.com/JuliaMath/AbstractFFTs.jl/pull/26 has been merged!

### DIFF
--- a/src/periodograms.jl
+++ b/src/periodograms.jl
@@ -218,11 +218,11 @@ See also: [`fftfreq`](@ref), [`rfftfreq`](@ref)
 freq(p::TFR) = p.freq
 freq(p::Periodogram2) = (p.freq1, p.freq2)
 FFTW.fftshift(p::Periodogram{T,F}) where {T,F<:Frequencies} =
-    Periodogram(p.freq.nreal == p.freq.n ? p.power : fftshift(p.power), fftshift(p.freq))
+    Periodogram(p.freq.n_nonnegative == p.freq.n ? p.power : fftshift(p.power), fftshift(p.freq))
 FFTW.fftshift(p::Periodogram{T,F}) where {T,F<:AbstractRange} = p
 # 2-d
 FFTW.fftshift(p::Periodogram2{T,F1,F2}) where {T,F1<:Frequencies,F2<:Frequencies} =
-    Periodogram2(p.freq1.nreal == p.freq1.n ? fftshift(p.power,2) : fftshift(p.power), fftshift(p.freq1), fftshift(p.freq2))
+    Periodogram2(p.freq1.n_nonnegative == p.freq1.n ? fftshift(p.power,2) : fftshift(p.power), fftshift(p.freq1), fftshift(p.freq2))
 FFTW.fftshift(p::Periodogram2{T,F1,F2}) where {T,F1<:AbstractRange,F2<:Frequencies} =
     Periodogram2(fftshift(p.power,2), p.freq1, fftshift(p.freq2))
 FFTW.fftshift(p::Periodogram2{T,F1,F2}) where {T,F1<:AbstractRange,F2<:AbstractRange} = p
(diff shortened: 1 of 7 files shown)


galenlynch:
Could you also:
1) Add a deprecation for these functions, pointing users to AbstractFFTs
2) Remove tests in `test/util.jl`
3) Remove the documentation from`docs/src/util.jl`

ssfrr:
Agreed there should be deprecation warnings here, though I don't think they should guide the user to use `AbstractFFTs`. As far as I understand the FFT usage, users shouldn't generally import `AbstractFFTs` directly. Most folks should import `FFTW`, which re-exports all functions exported from `AbstractFFTs`. `CuArrays` is another provider of FFT implementations, which doesn't currently use `@reexport` to re-export all the symbols but I [think it should](https://github.com/JuliaGPU/CuArrays.jl/issues/341).

I guess people writing generic code that takes in arrays of unknown types would use `AbstractFFTs` and then rely on the user to have the correct implementations loaded for whatever arrays they're using?

Perhaps @ararslan or @stevengj can clarify what the expectation is for how the FFT packages are supposed to be used?

ararslan:
From the AbstractFFTs README:
> This package is mainly not intended to be used directly.

So I agree that deprecations shouldn't point directly to AbstractFFTs.

crstnbr:
@galenlynch I implemented you suggestions. Regarding the deprecation I just `@deprecate`d functions to equivalents reexported by `FFTW` for now.

src/util.jl
```diff
@@ -108,52 +108,9 @@ fftabs2type(::Type{Complex{T}}) where {T<:FFTW.fftwReal} = T
...
-Return discrete fourier transform sample frequencies for use with
-`rfft`. The returned Frequencies object is an AbstractVector
-containing the frequency bin centers at every sample point. `fs`
-is the sample rate of the input signal.
-"""
-rfftfreq(n::Int, fs::Real=1) = Frequencies((n >> 1)+1, (n >> 1)+1, fs/n)
-FFTW.fftshift(x::Frequencies) = (x.nreal-x.n:x.nreal-1)*x.multiplier
+@deprecate Frequencies(nreal::Int, n::Int, multiplier::Float64) FFTW.Frequencies(nreal, n, multiplier)
+@deprecate fftfreq(n::Int, fs::Real=1) FFTW.fftfreq(n, fs)
+@deprecate rfftfreq(n::Int, fs::Real=1) FFTW.rfftfreq(n, fs)
```
martinholters: Should move these to `deprecated.jl`.

galenlynch:
@crstnbr could you please implement @martinholters suggestions so we can merge this once a new tagged version on AbstractFFTs.jl is available?

crstnbr:
@galenlynch I'm currently on vacation. Will try to get back to this next week!

crstnbr:
Alright, tests pass locally (with `AbstractFFTs#master` and `FFTW#master`).

crstnbr:
Now that https://github.com/JuliaMath/AbstractFFTs.jl/pull/26 has been merged and a new release has been tagged (https://github.com/JuliaMath/AbstractFFTs.jl/releases/tag/v0.5.0) this should be merged.

ararslan:
This should set `AbstractFFTs = "0.5"` in Project.toml.

crstnbr:
> This should set `AbstractFFTs = "0.5"` in the Project.toml `[compat]` section.

Why? DSP doesnt have AbstractFFTs as a direct dependency. So this shouldn't be necessary, right? Instead I believe we should set the FFTW compat, which I do in https://github.com/JuliaDSP/DSP.jl/pull/321.

ararslan:
We just need to make sure that DSP can't get AbstractFFTs < 0.5.

ararslan:
Something like https://github.com/JuliaMath/FFTW.jl/pull/124 will be necessary, I think. (cf https://github.com/JuliaDSP/DSP.jl/pull/321#discussion_r344464994)

src/periodograms.jl
```diff
@@ -6,6 +6,7 @@ using ..Util, ..Windows
 export arraysplit, nextfastfft, periodogram, welch_pgram, mt_pgram,
        spectrogram, power, freq, stft
 using FFTW
+import FFTW: Frequencies, fftfreq, rfftfreq
```
ararslan:
```suggestion
```
I believe this isn't necessary, since there's a `using FFTW` and FFTW reexports the symbols from AbstractFFTs. Same with the import from DSP in the tests.

ararslan:
I put this and #321 together with an adjustment to the FFTW bound and a version bump in #322 with Carsten's authorship preserved. Thanks so much for your work on this, @crstnbr!

Issue #306, opened 2019-06-10
Error when trying to use DSP


adamnemecek:
```
julia> using DSP
[ Info: Precompiling DSP [717857b8-e6f2-59f4-9121-6e50c889abd2]
WARNING: both FFTW and Util export "Frequencies"; uses of it in module Periodograms must be qualified
ERROR: LoadError: LoadError: UndefVarError: Frequencies not defined
Stacktrace:
 [1] top-level scope at none:0
 [2] include at ./boot.jl:326 [inlined]
 [3] include_relative(::Module, ::String) at ./loading.jl:1038
 [4] include at ./sysimg.jl:29 [inlined]
 [5] include(::String) at /Users/adamnemecek/.julia/dev/DSP/src/DSP.jl:3
 [6] top-level scope at none:0
 [7] include at ./boot.jl:326 [inlined]
 [8] include_relative(::Module, ::String) at ./loading.jl:1038
 [9] include(::Module, ::String) at ./sysimg.jl:29
 [10] top-level scope at none:2
 [11] eval at ./boot.jl:328 [inlined]
 [12] eval(::Expr) at ./client.jl:404
 [13] top-level scope at ./none:3
in expression starting at /Users/adamnemecek/.julia/dev/DSP/src/periodograms.jl:186
in expression starting at /Users/adamnemecek/.julia/dev/DSP/src/DSP.jl:16
ERROR: Failed to precompile DSP [717857b8-e6f2-59f4-9121-6e50c889abd2] to /Users/adamnemecek/.julia/compiled/v1.1/DSP/OtML7.ji.
Stacktrace:
 [1] compilecache(::Base.PkgId, ::String) at ./loading.jl:1197
 [2] _require(::Base.PkgId) at ./loading.jl:960
 [3] require(::Base.PkgId) at ./loading.jl:858
 [4] require(::Module, ::Symbol) at ./loading.jl:853
```

galenlynch:
The root cause is JuliaMath/AbstractFFTs.jl#26 taking these functions from DSP.jl and putting them in AbstractFFTs.jl instead. #300 should fix the problem but is still waiting for a response to review, and a new tagged version of AbstractFFTs.jl.

mchitre:
Struggled with this for a bit, so a quick note here of what worked:
- worked with `AbstractFFTs` `v0.4.1`
- failed with `AbstractFFTs` `v0.5.0`

crstnbr:
[...]

PR #307, opened 2019-06-10
use fully qualified util.frequencies in periodogram

adamnemecek:
This is related to #306. I'm not sure if this is the best way of fixing it but it seems to work.


galenlynch:
Thanks for submitting a PR to fix this!

See my response to #306: this should be fixed by #300 instead, when a newly tagged version of AbstractFFTs.jl is available.

PR #312, opened 2019-07-19
Add functionnality to create eye diagrams.

ma-laforge:
Hi,

I was wondering if it would be appropriate to add functionality to create eye diagrams in DSP.jl.

For those who are less familiar with this "analog" visualization of digital signals, there is information here:

https://en.wikipedia.org/wiki/Eye_pattern

Note that I personally consider this part of DIGITAL signal processing, because we are typically observing digital signals, and often correcting with digital filters (ex: FIR pre-emphasis on the TX output).

If this is ok, I will create a test set to validate the functionality.

Thanks.


ma-laforge:
Added test set & fixed algorithm.

ma-laforge:
Actually, by looking online at other numerical tools, it seems that eye diagrams would be better suited to a "Communications" module

..Maybe "PhysicalCommunications" (Phy layer) - because "Communications" is a bit broad a category.

PR #313, opened 2019-08-03
add modes to conv function

jasonsychau:
add modes to conv function

### DIFF
--- a/src/dspbase.jl
+++ b/src/dspbase.jl
@@ -289,41 +289,50 @@ function _conv_clip!(y::AbstractArray, minpad, axesu, axesv)
 end
 
 """
-    conv(u,v)
+    conv(u,v; mode = "full")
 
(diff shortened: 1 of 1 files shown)


galenlynch:
This is a very big change to how `conv` works with no explanation. Besides changing convolution from frequency domain to time domain, and thereby change its algorithmic complexity, this commit would also make `conv` no longer work with arbitrary dimensions (and instead only work with vectors), as well as no longer accepting inputs with arbitrary indexing (and instead only work with arrays that start at index one).

Supporting something like Matab's convolution modes could be done just by cropping the output of `conv`. If you want to support these modes at the library level, then that seems like a better place to start.

galenlynch (changes requested):
If you want to add this functionality, please implement it in a way that preserves the existing functionality and performance.

jasonsychau:
Sure, I’ll review at later time.


Thanks, Jason

PR #316, opened 2019-09-05
Fix filtfilt so it works for real coefficents and complex data.

bhawkins:
Trying to run `filtfilt` on complex data fails with `InexactError` due to calling `mul!(Y, A, B)` with `eltype(Y) <: Real`. I tried to correct the type logic and added a test.


galenlynch:
Looks like a simple fix, and the test is much appreciated. Thanks for putting in the effort!

PR #331, opened 2019-11-27
Add Julia 1.3 to CI

galenlynch:
Also remove Julia 1.1.

I don't know how many versions we should be testing? 1.0 is the LTS so it makes sense to keep it. 1.3 is the most recent version, and 1.2 is the version most people are probably running. I figured that it doesn't make that much sense to keep testing on 1.1, but happy to change that if people disagree.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,8 +3,7 @@ os:
   - linux
 julia:
   - 1.0
-  - 1.2
-  - 1.3
+  - 1
   - nightly
 matrix:
   allow_failures:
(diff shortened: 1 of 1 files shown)


ssfrr:
If you just put `1` it will pull the latest released version. For instance in another package I [test on 1.0, 1, and nightly, with failures allowed on nightly](https://github.com/JuliaAudio/SampledSignals.jl/blob/2e17121d438246dfe6f2c5490c680f097551775c/.travis.yml#L6)

martinholters:
As the tests don't take that long, I'd keep the rest on 1.1 as long as we support it.

Issue #235, opened 2018-10-16
Exponent Operator for ZeroPoleGain Type (works on 0.6, but not on 1.0)


standarddeviant:
The exponent operator of a ZeroPoleGain object works on Julia 0.6, but not on Julia 1.0.

Here's using that operator on 0.6.4:
```julia
julia> typeof(Butterworth(1))
DSP.Filters.ZeroPoleGain{Float64,Complex{Float64},Int64}

julia> methodswith(ZeroPoleGain)
6-element Array{Method,1}:
 *(f::DSP.Filters.ZeroPoleGain, g::Number) in DSP.Filters at C:\Users\sd\.julia\v0.6\DSP\src\Filters\coefficients.jl:37
 *(g::Number, f::DSP.Filters.ZeroPoleGain) in DSP.Filters at C:\Users\sd\.julia\v0.6\DSP\src\Filters\coefficients.jl:38
 *(f1::DSP.Filters.ZeroPoleGain, f2::DSP.Filters.ZeroPoleGain) in DSP.Filters at C:\Users\sd\.julia\v0.6\DSP\src\Filters\coefficients.jl:39
 *(f1::DSP.Filters.ZeroPoleGain, fs::DSP.Filters.ZeroPoleGain...) in DSP.Filters at C:\Users\sd\.julia\v0.6\DSP\src\Filters\coefficients.jl:41
 freqs(filter::DSP.Filters.ZeroPoleGain, w::Number) in DSP.Filters at C:\Users\sd\.julia\v0.6\DSP\src\Filters\response.jl:95
 freqz(filter::DSP.Filters.ZeroPoleGain, w::Number) in DSP.Filters at C:\Users\sd\.julia\v0.6\DSP\src\Filters\response.jl:14

julia> Butterworth(1)^2
DSP.Filters.ZeroPoleGain{Float64,Complex{Float64},Int64}(Float64[], Complex{Float64}[-1.0+0.0im, -1.0+0.0im], 1)
```

Here's using that operator on 1.0.0:
```julia
julia> typeof(Butterworth(1))
ZeroPoleGain{Float64,Complex{Float64},Int64}

julia> methodswith(ZeroPoleGain)
[1] freqs(filter::ZeroPoleGain, w::Number) in DSP.Filters at C:\Users\sd\.julia\packages\DSP\Cq0iO\src\Filters\response.jl:95
[...]

PR #336, opened 2019-12-26
Added support for exponent operator on FilterCoefficients types (#235)

nathangoenner:
First time contributing on Github so please let me know if I have something wrong.


galenlynch:
Thanks for doing this! Please add tests. I will review the code soon.

galenlynch:
And also add documentation.

src/Filters/coefficients.jl
```diff
@@ -380,3 +380,4 @@ SecondOrderSections(f::FilterCoefficients) = SecondOrderSections(ZeroPoleGain(f)
 *(f1::Biquad, fs::Biquad...) = SecondOrderSections([f1, fs...], 1)
 *(f1::SecondOrderSections, f2::Biquad) = SecondOrderSections([f1.biquads; f2], f1.g)
 *(f1::Biquad, f2::SecondOrderSections) = SecondOrderSections([f1; f2.biquads], f2.g)
+^(f::FilterCoefficients, n::Integer) = *(fill(f, n)...)
```
galenlynch: What happens if `n <= 0`?

nathangoenner: Good point.  I suppose for negative values of n it could return the reciprocal of the positive n value?  Values of n of 0 or 1 also need to be handled.

nathangoenner:
I still have to add some tests.  What would you like to see in the documentation for this?

src/Filters/coefficients.jl
```diff
@@ -380,3 +380,19 @@ SecondOrderSections(f::FilterCoefficients) = SecondOrderSections(ZeroPoleGain(f)
 *(f1::Biquad, fs::Biquad...) = SecondOrderSections([f1, fs...], 1)
 *(f1::SecondOrderSections, f2::Biquad) = SecondOrderSections([f1.biquads; f2], f1.g)
 *(f1::Biquad, f2::SecondOrderSections) = SecondOrderSections([f1; f2.biquads], f2.g)
+inv(f::FilterCoefficients) = begin
+    filt_pr = PolynomialRatio(f)
+    return typeof(f)(PolynomialRatio(filt_pr.a, filt_pr.b))
+end
+^(f::FilterCoefficients, n::Integer) = begin
```
galenlynch: Why not make it a multi-line function definition:

```julia
function ^(f::FilterCoefficients, n::Integer)
    ...
end
```

nathangoenner: If that's the preferred style I'll happily change it.

src/Filters/coefficients.jl
```diff
@@ -380,3 +380,19 @@ SecondOrderSections(f::FilterCoefficients) = SecondOrderSections(ZeroPoleGain(f)
 *(f1::Biquad, fs::Biquad...) = SecondOrderSections([f1, fs...], 1)
 *(f1::SecondOrderSections, f2::Biquad) = SecondOrderSections([f1.biquads; f2], f1.g)
 *(f1::Biquad, f2::SecondOrderSections) = SecondOrderSections([f1; f2.biquads], f2.g)
+function inv(f::FilterCoefficients)
```
nathangoenner: I added a "inv" function because by default taking f::FilterCoefficients ^ -1 invoked Base.inv and caused an error.

src/Filters/coefficients.jl
```diff
@@ -385,14 +385,8 @@ function inv(f::FilterCoefficients)
...
-        filt_copy = *(fill(f, abs(n))...)
-    elseif n == 0
-        filt_copy = PolynomialRatio(Complex[1], Complex[1])
-    end
-    if sign(n) == -1
-        filt_copy = inv(filt_copy)
-    end
+    filt_copy = PolynomialRatio(f)
+    filt_copy = PolynomialRatio(filt_copy.b^(abs(n)), filt_copy.a^(abs(n)))
+    (sign(n) == -1) && (filt_copy = inv(filt_copy))
```
nathangoenner: I think this usage of a conditional is a bit hard to read, but it's used many other places and keeps the code small.

nathangoenner:
One thing I noticed while writing tests is that `Butterworth(1) == Butterworth(1)` evaluates to false.  Is this something worth opening a ticket over?  It would simplify this unit test if filters could be tested for equality directly.

martinholters:
From a brief look, I'm not too happy about always converting to `PolynomialRatio` for the actual computation and then converting back. For a `Biquad`, converting back is probably impossible (unless `n==1` or `n==-1`), so returning a `PolynomialRatio` might be preferable. For `SecondOrderSections` it would probably be nice to keep the previous decomposition. For `ZeroPoleGain` it works in principle, but seems to be a bad idea numerically. Generally, I think supporting `^` is nice (though I don't know how often one would use it in practice), so if you could implement it individually for the different `FilterCoefficients` subtypes, that would be appreciated.

PR #348, opened 2020-03-22
Explicitly test Julia 1.3

rob-luke:
Julia 1.4 has been released, so `1` will now test `1.4`. And `1.3` needs to be explicitly added.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ julia:
   - 1.0
   - 1.1
   - 1.2
+  - 1.3
   - 1
   - nightly
 matrix:


rob-luke:
So it looks like Travis is still pointing to 1.3.1 according to the build above. I will rerun the tests once it looks like travis is pointing to 1.4.

https://travis-ci.org/github/JuliaDSP/DSP.jl/jobs/665358533#L166

rob-luke:
Could some please review now. The 1 test is now testing 1.4


https://travis-ci.org/github/JuliaDSP/DSP.jl/jobs/670887301#L165

ssfrr:
Thanks for looking out for this.

With most of my packages I skip the non-LTS 1.x versions, and just to `1.0` and `1`, but given that DSP is a pretty core package I think it makes sense to be a little more conservative and test on all of the 1.x releases.

PR #359, opened 2020-05-13
fix poly depwarn and 50x performance drop

SimonDanisch:
I haven't checked anything, besides running the tests and my code and verified that it works^^


galenlynch:
Thanks for doing this. Ultimately we should just figure out what's changed with Polynomials.jl and use the new interface. However, I don't see why we shouldn't use this much needed stop-gap until then. Should we just merge this and release a new patch version?

ararslan:
> Should we just merge this and release a new patch version?

:+1:

Issue #352, opened 2020-04-23
conv between fixed-point array and integer array results in OOM


fghzxm:
`conv` between an `Array{FixedPointNumbers.Fixed{T,f}}` and an `Array{Int}` results in OOM.

Reproduction:
```julia
using FixedPointNumbers, DSP;
conv(Array{Q0f15}(undef, 100000), [-1, 1])
```

galenlynch:
Thanks for posting this bug, I'll look into it tonight.

galenlynch:
It seems like one of the `conv` conversion methods has overly strict type bounds. I'll fix it in a PR.

Right now, it looks like `conv` will automatically convert fixed point numbers into BLAS floating point numbers, which is behavior inherited from Julia base. I'm not sure if there's still a good reason to do this, since I think this conversion was for the sake of the `FFTW` calls inside of `conv`. I need to do some testing, but I think that all the code in `conv` should be fine without this conversion. If tha'ts the case, then perhaps it would make sense to remove this conversion to keep the element type of the inputs, though it would technically be a breaking change.

galenlynch:
This example triggers the same error, without using `FixedPointNumbers`:

```julia
using DSP;
conv([1//2, 1//3, 1//4], [1, 2])
```

rob-luke:
I have a question about the fixed point numbers, not the `x//y` operator you provide above. I comment here because the fix in #353 looks great for `//`.

If you make the breaking changes to keep the element type, will the entire conv operation be performed using fixed point numbers/operations? Or will it convert to something float like somewhere along the line for computation? I am trying to put myself in the mind of a `FixedPointNumbers` user. I assume these users are trying to understand the effect that operating in fixed point will have on their processing, perhaps in preparation for converting their code to a fixed point DSP hardware. In this case, the user would want all operations to be performed in fixed point, not converted to float and then converted back to fixed along the way.

Personally I would love to have full support for fixed point operations. I have not looked in to it (so would be pleased to be corrected here), but it may be that enforcing all downstream operations to operate in fixed point is currently beyond the scope of `DSP.jl`, and in that case should we throw an error or warning if a user passes in fixed point numbers. Something like `@warn(Input contains fixed point numbers. Some processing steps may be computed after conversion to float.)`. What are your thoughts @galenlynch?

galenlynch:
[...]

Issue #355, opened 2020-04-27
Implementation of `conv()` Uses Overlap and Save While Direct Method Is Faster


RoyiAvital:
The current implementation uses in many cases the [Overlap and Save](https://en.wikipedia.org/wiki/Overlap%E2%80%93save_method) method for Linear Convolution. 
While this method is optimized to a streamed data it is not optimized for cases where all data is given in memory.

See https://discourse.julialang.org/t/convolution-conv-with-same-size-output/38260/11:

![Figure0001](https://user-images.githubusercontent.com/7347975/80370992-d568a400-8899-11ea-821d-b8bec750cf31.png)

I think it is better to implement direct loop with `@simd` and `@inbounds`. It will probably be faster than the current method for most cases. For longer kernels / signals it should use the Frequency Domain as it is implemented now,

ssfrr:
See this similar plot (but for longer filters) that @galenlynch made that compares regular FFT processing to overlap-save. I think that the thing to do is to add another check for short convolutions that uses the direct method for those.

https://github.com/JuliaDSP/DSP.jl/pull/286#issuecomment-484977111

@RoyiAvital do you have the code for that plot? I'm curious about the weird diagonal lines where the overlap-save is faster. Did you use `nextfastfft` for the FFT-based plots?

galenlynch:
Much of this is a repeat of what I said on Discord, but I think the tradeoff between direct and fft or overlap-save is not a question of whether everything is in memory or not, but rather what the kernel size is. FFT convolution has an algorithmic advantage over direct convolution (NlogN vs N^2), so you should only choose direct convolution when the overhead is not worth the savings in complexity. Those savings are only [increased](https://github.com/JuliaDSP/DSP.jl/pull/286#issuecomment-484977111) with overlap-save compared to fft convolution. In my testing, the break-even point between fft convolution and direct convolution happens when the kernel is length 58. You are welcome to compare for yourself as direct convolution is already implemented in DSP.jl in the `td_filt` function for 1d arrays, which is chosen inside of [`filt`](https://github.com/JuliaDSP/DSP.jl/blob/f6c051d562dd7422c52f9628fccf391c6413723b/src/Filters/filt.jl#L495) if the kernel is less than 59 elements long. It is already vectorized (i.e. benefits from `@simd` and `@inbounds`), and furthermore we have unrolled the inner loop for kernel lengths less than 59.

[...]

PR #360, opened 2020-05-13
Reject non-numerical arrays as arguments to conv

fghzxm:
Related to #352 (#353)

Currently `conv` has a method `conv(::AbstractArray, ::AbstractArray)`
that is intended to cast arrays of different ndims into
dimension-compatible ones before delegating to more detailed methods.
Unfortunately, this method ends up catching arrays with non-`Number`
eltypes, and since no detailed methods accept such arrays as arguments,
this method ends up calling itself and results in stack overflow.  This
commit fixes that by restricting the signature of the abovementioned
method to `conv(::AbstractArray{<:Number, N}, ::AbstractArray{<:Number,
P}) where {N, P}` so that calling `conv` with non-numerical arrays
is now a type error.

### DIFF
--- a/src/dspbase.jl
+++ b/src/dspbase.jl
@@ -718,10 +718,14 @@ function conv(u::AbstractArray{<:BLAS.BlasFloat, N},
     conv(u, float(v))
 end
 
-function conv(A::AbstractArray{<:Number, N},
-              B::AbstractArray{<:Number, P}) where {N, P}
-    maxnd = max(ndims(A), ndims(B))
-    return conv(cat(A, dims=maxnd), cat(B, dims=maxnd))
+function conv(A::AbstractArray{<:Number, M},
+              B::AbstractArray{<:Number, N}) where {M, N}
+    if (M < N)
+        conv(cat(A, dims=N), B)
+    else
+        @assert M > N
+        conv(A, cat(B, dims=M))
+    end
 end
 
 """
--- a/src/dspbase.jl
+++ b/src/dspbase.jl
@@ -721,10 +721,10 @@ end
 function conv(A::AbstractArray{<:Number, M},
               B::AbstractArray{<:Number, N}) where {M, N}
     if (M < N)
-        conv(cat(A, dims=N), B)
+        conv(cat(A, dims=N)::AbstractArray{eltype(A), N}, B)
     else
         @assert M > N
-        conv(A, cat(B, dims=M))
+        conv(A, cat(B, dims=M)::AbstractArray{eltype(B), M})
     end
 end
 
--- a/src/dspbase.jl
+++ b/src/dspbase.jl
@@ -718,7 +718,8 @@ function conv(u::AbstractArray{<:BLAS.BlasFloat, N},
     conv(u, float(v))
 end
 
-function conv(A::AbstractArray, B::AbstractArray)
+function conv(A::AbstractArray{<:Number, N},
+              B::AbstractArray{<:Number, P}) where {N, P}
     maxnd = max(ndims(A), ndims(B))
     return conv(cat(A, dims=maxnd), cat(B, dims=maxnd))
 end
--- a/test/dsp.jl
+++ b/test/dsp.jl
@@ -64,6 +64,8 @@ end
         @test conv(offset_arr, 1:3) == OffsetVector(expectation, 0:5)
         # Issue #352
         @test conv([1//2, 1//3, 1//4], [1, 2]) ≈ [1//2, 4//3, 11//12, 1//2]
+        # Non-numerical arrays should not be convolved
+        @test_throws MethodError conv([sin], [cos])
     end
 
 


src/dspbase.jl
```diff
@@ -718,7 +718,8 @@ function conv(u::AbstractArray{<:BLAS.BlasFloat, N},
     conv(u, float(v))
 end
 
-function conv(A::AbstractArray, B::AbstractArray)
+function conv(A::AbstractArray{<:Number, N},
+              B::AbstractArray{<:Number, P}) where {N, P}
     maxnd = max(ndims(A), ndims(B))
     return conv(cat(A, dims=maxnd), cat(B, dims=maxnd))
```
galenlynch, line 724: You can use the static parameters here to find maxnd: `maxnd = max(N, P)`

Also definitely picking at nits but why `P`? `M` and `N` would be more common choices for the number of dimensions.

martinholters:
This is fine, I guess, but doesn't fix all the problems. Even for `Number` types, we expect that `float()` will turn them into `BlasFloat`s, which may not be the case, e.g.
```julia
julia> conv(BigFloat[1], [1])
ERROR: StackOverflowError:
Stacktrace:
 [1] conv(::Array{BigFloat,1}, ::Array{Float64,1}) at DSP/src/dspbase.jl:713 (repeats 79984 times)
```
I don't think this PR fixes that. Of course, it doesn't have to; just pointing out that there is more to do. Whether as part of this PR or for another day is another matter.

src/dspbase.jl
```diff
@@ -718,9 +718,14 @@ function conv(u::AbstractArray{<:BLAS.BlasFloat, N},
     conv(u, float(v))
 end
 
-function conv(A::AbstractArray, B::AbstractArray)
-    maxnd = max(ndims(A), ndims(B))
-    return conv(cat(A, dims=maxnd), cat(B, dims=maxnd))
+function conv(A::AbstractArray{<:Number, M},
+              B::AbstractArray{<:Number, N}) where {M, N}
+    if (M < N)
+        conv(cat(A, dims=N), B)
```
martinholters, line 724: While at it: `cat` is not type-stable, but I think we can at least make the return type of `conv` inferable by adding a type assertion like
```suggestion
        conv(cat(A, dims=N)::AbstractArray{typof(A), N}, B)
```
and similar below.

fghzxm:
> This is fine, I guess, but doesn't fix all the problems. Even for `Number` types, we expect that `float()` will turn them into `BlasFloat`s, which may not be the case, e.g.
> 
> ```julia
> julia> conv(BigFloat[1], [1])
> ERROR: StackOverflowError:
> Stacktrace:
>  [1] conv(::Array{BigFloat,1}, ::Array{Float64,1}) at DSP/src/dspbase.jl:713 (repeats 79984 times)
> ```
> 
> I don't think this PR fixes that. Of course, it doesn't have to; just pointing out that there is more to do. Whether as part of this PR or for another day is another matter.

See also JuliaImages/juliaimages.github.io#129

src/dspbase.jl
fghzxm: Do you mean `eltype` instead of `typeof`?

src/dspbase.jl
```diff
@@ -718,9 +718,14 @@ function conv(u::AbstractArray{<:BLAS.BlasFloat, N},
...
 
-function conv(A::AbstractArray, B::AbstractArray)
-    maxnd = max(ndims(A), ndims(B))
-    return conv(cat(A, dims=maxnd), cat(B, dims=maxnd))
+function conv(A::AbstractArray{<:Number, M},
+              B::AbstractArray{<:Number, N}) where {M, N}
+    if (M < N)
+        conv(cat(A, dims=N)::AbstractArray{eltype(A), N}, B)
+    else
+        @assert M > N
```
galenlynch, line 726: Is it your intent here to stop infinite recursion when there is no other applicable `conv` methods (as would happen with arrays of `Gray` values)?

src/dspbase.jl
martinholters: Oh yes, of course.

src/dspbase.jl
fghzxm: My intent is to ensure all methods of `conv` reject arrays of non-`Number` eltype.  Since `Gray` is not `Number`, my version will raise `MethodError` before executing any method of `conv`.

What martinholters points out as a problem I believe is what would happen if `Gray` were actually made a subtype of `Number`, but `float(::Gray{<:Number})` remained unchanged (still returning a `Gray{<:AbstractFloat}`).

fghzxm:
```
@assert M > N
```
should always succeed, since if `M == N` it would be some above-defined method that got called.  I put it here essentially as a comment, and also so that debugging would be easier should other methods would get messed up in the future.

martinholters: Hm, is there any chance some code defined `conv(u::AbstractArray{<:Foo, N}, v::AbstractArray{<:Foo, N}) where {N}` for some non-`Number` `Foo` and relied on the method herein to get arguments to the same dimensionality first?

In general, this method really doesn't need its argument eltypes to be restricted to `Number`. Maybe a better approach would be to define `conv(u::AbstractArray{<:Any, N}, v::AbstractArray{<:Any, N}) where {N}` (i..e. for eltypes not otherwise handled) to either throw an informative error or provide a time-domain implementation that only requires `+` and `*` be defined for the elements. Maybe throw an error for now and leave time-domain convolution for a future improvement.

fghzxm: @martinholters I think it's better to provide time-domain convolution as a seperate function from `conv`, since the user may want to be able to choose between time-domain and FFT even if their arrays are numeric.  If we're to provide two implementations both under `conv` I think we should provide an explicit flag parameter for specifying the algorithm.  It seems strange to me that nominally different eltypes like `Float32` and `Gray{Float32}` can invoke completely different algorithms.

A better solution to the `Gray` problem in my opinion is something like a function `LinearAlgebra.blasfloat()` that is guaranteed to return a naked `BlasFloat`, unlike `float()` which may return a still wrapped type.

Also there is a package that does time-domain convolution: [aamini/FastConv.jl](aamini/FastConv.jl)

galenlynch: @fghzxm allowing the user to specify the algorithm seems like an excellent idea. I personally like what ImageFiltering did for this, make a flag that is a type to allow dispatching to eliminate branches in some conditions (e.g. if the element type is not compatible with fourier transforms). Thanks for the suggestion with `blasfloat`, seems like a good solution. In terms of a time-domain convolution, I'll check out FastConv.jl. My impression last time I looked was that it was not actively maintained, but it seems like there's been some recent activity. Despite this, there are still some signs that it's not that up to date, e.g. it uses `conv2`, which was formerly in Julia base, then migrated to DSP.jl, and then deprecated. However, FastConv does not use DSP.jl. I'm not sure it would work atm. Also, looking at the code in FastConv, it looks like N dimensional case it's using `@nloops`, which has been soft-replaced by `CartesianIndices` etc.

DSP.jl already has fast time domain convolution for 1D inputs as part of the `filt` function, which could be easily adapted for general convolution. I'll try to add a time-domain branch of `conv` sometime soon for the 1D case, and think about how to do the ND case, and ideally also support custom array indexing. As was discussed in great detail in #355, I think the performance savings might be minimal, but perhaps a better reason to add time-domain convolution is to support convolution of arrays with any element type that has `+` and `*` defined, as @martinholters suggested. It could then act as a fall-back for inputs that are incompatible with fourier methods.

galenlynch:
Ok to merge @martinholters ?

Btw I implemented ND time-domain convolution and added multi threading to all convolution methods. I'm still working on making all of the conv methods handle custom indexing, and still need to make some heuristic function that decides when to use each method.

galenlynch:
I would be eager to hear your thoughts on `conv` @martinholters, since I have been putting a lot of effort into it in the last few weeks. I haven't touched any of the "front-end" methods like the ones changed here, but have simply been adding more conv algorithms. So far the performance gains are pretty sizeable, and would also allow convolution of anything that supports `*` and `+`.

martinholters:
Well, yes, I haven't really thought trough where I would like to take it, just realized that it's a bit of a mess. But if you have methods that work on anything with `* ` and `+` and have good performance, then :100:! Restructuring the dispatch in the front-end can happen independently.

PR #362, opened 2020-05-24
Workaround for performance regression introduced by FFTW

### LINKED ISSUE #339: Performance regression in conv for images
From slack: I recently noticed a performance regression in `conv` 
```julia
using Images, DSP

# before pulling master
img = randn(1000,1000)
kernel = Kernel.LoG(4)
@btime DSP.conv($img, $(kernel)) # 42.050 ms (861 allocations: 61.68 MiB)
@btime imfilter!($storage, $img, $(kernel)) # 56.002 ms (967 allocations: 101.53 MiB)
```
After pulling DSP master to make use of the new algorithm I noticed that it is actually way slower
```julia
@btime DSP.conv($img, $(kernel)) # 73.747 ms (460183 allocations: 47.83 MiB)
```

galenlynch:
Convolutions in DSP currently rely on FFTW.jl, and a recent change in FFTW.jl
(JuliaMath/FFTW.jl#105) has introduced a large performance regression in `conv`
whenever Julia is started with more than one thread. Since v1 of FFTW.jl, it uses multi-threaded
FFTW transformations by default whenever Julia has more than one thread. This
new default causes small FFT problems to run much more slowly and use much more
memory. Since the overlap-save method of `conv` in DSP breaks a convolutions
into small convolutions, and therefore performs a large number of small FFTW
transformations, this change can cause convolutions to be slower by two orders
of magnitude, and similarly use two orders of magnitude more memory. While
FFTW.jl does not provide an explicit way to set the number of threads used by a
FFTW plan without changing a global variable, generating the plans with the
planning flag set to `FFTW.PATIENT` (instead of the default `MEASURE`) allows
the planner to consider changing the number of threads. Adding this flag to the
plans generated by the overlap-save convolution method seems to rescue the
performance regression on multi-threaded instances of Julia.

Fixes #339
Also see JuliaMath/FFTW.jl#121

### DIFF
--- a/src/dspbase.jl
+++ b/src/dspbase.jl
@@ -301,8 +301,9 @@ unnormalized.
     bufsize = ntuple(i -> i == 1 ? nffts[i] >> 1 + 1 : nffts[i], N)
     fdbuff = similar(u, Complex{T}, NTuple{N, Int}(bufsize))
 
-    p = plan_rfft(tdbuff)
-    ip = plan_brfft(fdbuff, nffts[1])
+    # PATIENT flag needed if Julia has more than one thread (See #339)
+    p = plan_rfft(tdbuff, flags = FFTW.PATIENT)
+    ip = plan_brfft(fdbuff, nffts[1], flags = FFTW.PATIENT)
 
     tdbuff, fdbuff, p, ip
 end
@@ -310,8 +311,9 @@ end
 @inline function os_prepare_conv(u::AbstractArray{<:Complex}, nffts)
     buff = similar(u, nffts)
 
-    p = plan_fft!(buff)
-    ip = plan_bfft!(buff)
+    # PATIENT flag needed if Julia has more than one thread (See #339)
+    p = plan_fft!(buff, flags = FFTW.PATIENT)
+    ip = plan_bfft!(buff, flags = FFTW.PATIENT)
 
     buff, buff, p, ip # Only one buffer for complex
 end


martinholters:
This seems to take significant time when running the tests---enough for travis to cancel the jobs after 10 minutes without output.

martinholters:
The planning time seems to become excessive in the higher-dimensional case:

```julia
julia> x = zeros(128, 128, 128);

julia> out = zeros(255, 255, 255);

# master
julia> @time DSP.unsafe_conv_kern_os!(out, x, x, size(x), size(x), size(out), (256, 256, 256));
  9.859362 seconds (13.47 M allocations: 1.152 GiB, 3.15% gc time)

julia> @time DSP.unsafe_conv_kern_os!(out, x, x, size(x), size(x), size(out), (256, 256, 256));
  5.168003 seconds (2.27 k allocations: 515.127 MiB, 2.97% gc time)

# this PR
julia> @time DSP.unsafe_conv_kern_os!(out, x, x, size(x), size(x), size(out), (256, 256, 256));
751.674809 seconds (13.06 M allocations: 1.381 GiB, 0.04% gc time)

julia> @time DSP.unsafe_conv_kern_os!(out, x, x, size(x), size(x), size(out), (256, 256, 256));
  3.693583 seconds (2.25 k allocations: 772.126 MiB, 4.53% gc time)
```
(Note: this is without multiple julia threads.)

This looks prohibitive, I'm afraid, but I don't have an alternative idea.

galenlynch:
Ah, bummer :(

I guess we'll have to wait to see if JuliaMath/FFTW.jl#150 gets merged...

PR #366, opened 2020-06-03
Use FFTW 1.2.2 to fix performance regression

### LINKED ISSUE #339: Performance regression in conv for images

galenlynch:
FFTW 1.2.2 restored single-threaded FFT plans by default, which fixes a
performance regression in DSP.jl (#339).

Closes #339.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "DSP"
 uuid = "717857b8-e6f2-59f4-9121-6e50c889abd2"
-version = "0.6.7"
+version = "0.6.8"
 
 [deps]
 FFTW = "7a1cc6ca-52ef-59f5-83cd-3a7055c09341"
@@ -13,7 +13,7 @@ SpecialFunctions = "276daf66-3868-5448-9aa4-cd146d93841b"
 Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 
 [compat]
-FFTW = "1.1"
+FFTW = "1.2.2"
 IterTools = "1"
 OffsetArrays = "0.11"
 Polynomials = "1.0"


galenlynch:
It looks like this PR is incompatible with Julia versions 1.0 through 1.2 because FFTW has dropped support for those versions of Julia. Maybe what we could do is introduce two patch versions, one that maintains compatibility with Julia 1.0-1.2 and limits FFTW compat to the versions to before it used multithreaded FFT plans, and a second patch version where support for Julia versions 1.0-1.2 is dropped and the new FFTW is required.

galenlynch:
Also worth mentioning: I'm currently working on a PR to use multi-threading for convolution, which would require dropping support for Julia 1.0-1.2, so we will have to face this issue at some point. What do people think about dropping support for Julia 1.0?

martinholters:
I do prefer supporting the latest LTS version (which still is 1.0, but I believe there are rumors it might be replaced soon) if possible at all. Maybe the multi-threaded convolution could live behind an `if VERSION` for the time being?

Anyway, I don't think this PR is necessary. We still support older `FFTW` versions, but for performance reasons, use should update to the latest one. I doubt we should force people to update `FFTW` (which will only happen if the do update `DSP`, anyway).

galenlynch:
ok sounds good

PR #375, opened 2020-08-23
Expand compat with OffsetArrays to include v1

galenlynch:
This is a test dependency only, and was not caught by the GitHub action.


silenj:
Great! Thank you.

Issue #376, opened 2020-09-04
Function for group delay?


githubtomtom:
We have `freqz()` and `phasez()`. I wonder if there’s any implementation for group delay ? (which is  -d/dw phase)

Thanks.

galenlynch:
I don't think we currently have a function that characterizes the group delay of a filter.

martinholters:
Fixed by #377.

PR #377, opened 2020-09-13
Add grpdelay

JuiceWizerd:
This pull request is to solve #376 and create a group delay function for use. The implementation is the Smith algorithm [1], modified to work for FIR and IIR filters [2], that is used in MATLAB and SciPy. 

Please let me know if there is anything to change. This is my first pull request.

[1]  Richard G. Lyons, "Understanding Digital Signal Processing, 3rd edition", p. 830.
[2] J. Smith, "Numerical Computation of Group Delay", Ccrma.stanford.edu. [Online]. Available: https://ccrma.stanford.edu/~jos/filters/Numerical_Computation_Group_Delay.html


galenlynch:
Thanks for the contribution, and congrats on your first pull request! I, or one of the other maintainers, will take a look soon.

src/Filters/response.jl
```diff
@@ -58,6 +60,25 @@ function phasez(filter::FilterCoefficients, w = range(0, stop=π, length=250))
...
+function grpdelay(filter::FilterCoefficients, w = range(0, stop=π, length=250))
+    filter = convert(PolynomialRatio, filter)
+    b, a = coefb(filter), filter.a.coeffs # reversed a
+    c = conv(b, conj(a))
+    cr = range(0, stop=length(c)-1) .* c
+    ejw = exp.(-im * w)
+    num = Polynomial(cr).(ejw)
+    den = Polynomial(c).(ejw)
+    gd = real(num ./ den) .- length(a) .+ 1
+    [x == Inf ? 0 : x for x in gd] # Set Infs to zeros
```
martinholters, line 78: Why is `Inf` replaced with `0`, but neither `-Inf` nor `NaN`? And why is it legitimate to replace it with zero anyway? IIUC, `gd` has non-finite values if the filter has poles and/or zeros on the unit circle. In both cases, there may well be a continuous extension of the group delay which will not necessarily be zero. E.g. for `filter = PolynomialRatio([1, 0, 1], [1])` and `w=π/2`, if it wasn't for numerical precision issues, one would get `num=-2` and `den=0`, hence either `gd=-Inf` (if `den==0.0`) or `gd=Inf` to be replaced with zero (if `den==-0.0`), while the only reasonable finite result would be `1`.

martinholters:
As we use `FilterCoefficients` for both analog and digital filters, we have `freqz`, `impz`, `stepz`, `phasez`. Should this be `grpdelayz`?

src/Filters/response.jl
```diff
@@ -58,6 +60,25 @@ function phasez(filter::FilterCoefficients, w = range(0, stop=π, length=250))
...
+
+Group delay of a digital 'filter' at normalized frequency
+or frequencies 'w' in radians/sample.
+"""
+function grpdelay(filter::FilterCoefficients, w = range(0, stop=π, length=250))
+    filter = convert(PolynomialRatio, filter)
+    b, a = coefb(filter), filter.a.coeffs # reversed a
+    c = conv(b, conj(a))
+    cr = range(0, stop=length(c)-1) .* c
+    ejw = exp.(-im * w)
```
martinholters, line 74:
```suggestion
    ejw = exp.(-im .* w)
```
gets rid of a temporary array.

src/Filters/response.jl
JuiceWizerd: I didn't think of those values. Looking at MATLAB it seems that they let Inf values occur in their implementation. SciPy checks to see if any values in ```den < 10 * EPSILON```. Should I just take this part out?

martinholters: Yes, I think returning `gd` should be fine. If the need arises, we can always figure out how to close any gaps.

JuiceWizerd:
> As we use `FilterCoefficients` for both analog and digital filters, we have `freqz`, `impz`, `stepz`, `phasez`. Should this be `grpdelayz`?

MATLAB and Scipy both implicitly support their group delay functions for digital filters. Is it common to check the group delay for an analog filter? This a change that I wouldn't mind making to align with the other functions if it makes more sense.

martinholters:
I think it would be more consistent to name it `grpdelayz`, even if it breaks with the relationship to Matlab and Scipy. But maybe others can chime in here?

src/Filters/response.jl
```diff
@@ -58,6 +60,24 @@ function phasez(filter::FilterCoefficients, w = range(0, stop=π, length=250))
...
+"""
+function grpdelay(filter::FilterCoefficients, w = range(0, stop=π, length=250))
+    filter = convert(PolynomialRatio, filter)
+    b, a = coefb(filter), filter.a.coeffs # reversed a
+    c = conv(b, conj(a))
+    cr = range(0, stop=length(c)-1) .* c
+    ejw = exp.(-im .* w)
+    num = Polynomial(cr).(ejw)
+    den = Polynomial(c).(ejw)
+    gd = real(num ./ den) .- length(a) .+ 1
```
martinholters, line 77:
```suggestion
    return real.(num ./ den) .- (length(a) - 1)
```
The `return` for stylistic reasons, the `.` to avoid unnecessary temporaries by allowing broadcast fusion, and the `- 1` to do the subtraction only once instead of once per element. We could avoid further allocations by substituting `num` and `den` with their definitions, but that might be at the cost of readability, not sure if worth it.

test/filter_response.jl
```diff
@@ -213,3 +213,29 @@ end
...
+    b2 = [1, -1.0166, 1]
+    a1 = [1, -0.683]
+    a2 = [1, -1.4461, 0.7957]
+    b = b0*conv(b1, b2)
+    a = conv(a1, a2)
+    df = PolynomialRatio(b, a)
+    w = matlab_delay[:, 1]
+
+    grpdelay_matlab = matlab_delay[:, 2]
+    @test grpdelay(df, w) ≈ grpdelay_matlab
```
martinholters, line 240: Maybe we can also test for a few cases where the correct result can be obtained "by hand"? E.g.
```julia
@test grpdelay(PolynomialRatio([1, 2, 2, 1], [1]) ≈ fill(1.5, 250)
@test grpdelay(PolynomialRatio([1, 2, 3, 2, 1], [1]) ≈ fill(2.0, 250)
```
Maybe also a first-order IIR filter?

galenlynch:
Yeah, `grpdelayz` sounds good to me.

src/Filters/response.jl
JuiceWizerd: I added tests for type I-IV filters. I don't think that they pass. There are cases where NaN occurs from division by 0 + 0*im. Some Inf values also occur. I think I should check for values less than eps(Float64) and maybe make a case for linear phase FIR filters where it returns an array of the "by hand" values.

src/Filters/response.jl
```diff
@@ -128,3 +154,13 @@ end
 function hz_to_radians_per_second(hz, fs)
     hz * ((2 * pi) / fs)
 end
+
+function _is_sym(x::AbstractArray)
+    n = length(x) ÷ 2
+    return all([x[1+i] == x[end-i] for i in 0:n])
```
martinholters, line 160:
```suggestion
    return all(x[1+i] == x[end-i] for i in 0:n-1)
```
Instantiating the vector is unnecessary, it creates an avoidable temporary and needs to perform all `==` tests, even if the first one already gives `false`. Also we can stop at `n-1`, I think.

src/Filters/response.jl
```diff
@@ -128,3 +154,13 @@ end
...
+end
+
+function _is_anti_sym(x::AbstractArray)
+    n = length(x) ÷ 2
+    return all([x[1+i] == -x[end-i] for i in 0:n])
```
martinholters, line 165:
```suggestion
    return all(x[1+i] == -x[end-i] for i in 0:n)
```
Here, too, we can optimize by avoiding the temporary vector. We have to go to `n` though, to ensure the middle element for odd length is zero (by comparing with itself negated). Or we use `n = (length(x) + 1) ÷ 2` and stop at `n-1`; not sure if worth it.

martinholters (changes requested):
Some last performance tweaks, then this should be good to go from my side.

JuiceWizerd:
Thanks for the performance tweaks Martin. I've implemented your changes in the latest commit.

martinholters (approved):

JuiceWizerd:
Should issue #376 be linked to this pull request? When this is merged will that automatically close that issue if it is linked?

martinholters:
Can just close manually. Otherwise, just stating "fixes #376" in the description would have sufficed to automatically close it.

BTW, thanks for the contribution and welcome!

PR #384, opened 2020-11-04
Rework filter response functions

martinholters:
With #284 finally in, it is no longer necessary to distinguish between `freqz` and `freqs`. As `freq` is already taken, I opted for `freqresp` as the common name. I also DRYed up that code quite a bit. Allowing phase response to be computed for analog filters followed directly, leading to `phasez` -> `phaseresp`. It followed to rename `grpdelayz` -> `grpdelay`, `impz` -> `impresp`, and `stepz` -> `stepresp`. As a bonus, I've added `grpdelay` for analog filters. Considering `impresp` and `stepresp` for analog filters, it would be possible to come up with something, but the API would have to be different, taking a vector of sampling times instead of the number of samples to produce as second input. Not sure if that justifies staying with the old names here.

~I'm also not entirely happy about the API. Are we sure we want the default `w` for `freqresp`/`phaseresp`/`grpdelay` for the digital case? (It makes much less sense in the analog case because the relevant frequency range is arbitrary.) Also, why does `freqresp` support taking `hz` and `fs` instead of `w`, but neither `phaseresp` nor `grpdelay` do? And does it really make sense to have `hz` and `fs` instead of `w` for the analog case?~ EDIT: If not given the frequencies `w`, `freqresp` now returns them so that the calling code does not have to rely on specific values chosen. Also, a default for analog filters (based on the pole/zero loci) is added. And I've dropped the `hz`+`fs` method altogether; can be re-added later if anyone actually misses it.

~Finally, I'm considering turning `_freq()` (evaluating the transfer function for a certain `z` (or `s`)) into methods of the `FilterCoefficients` object, so that e.g. `H = PolynomialRatio(b, a); H(exp(im*W))` would work.~ EDIT: I still like the idea, but it's better postponed to another PR.

Opinions?


martinholters:
Pinging @galenlynch and @ssfrr for feedback. Not necessarily code review, yet, but just some general feedback on the questions posed in the OP. (Others are welcome to give feedback, too, of course.)

galenlynch:
I think renaming the `freqs` and `freqz` functions sounds great, given that the `s` vs `z` distinction no longer needs to be made with the function name. Similarly, I think having consistent naming for functions that characterize filters is strictly a good thing, as long as we have all the necessary deprecations and support the old function names during the deprecation period.

Your question about whether to keep a default sampling frequency for digital filters is an interesting one. Are you worried that about having an inconsistent type signature for these characterization functions across both analog and digital filters? I personally find characterizing digital filters with normalized sampling frequencies quite natural, and therefore a reasonable (time-saving) default. If you think that throwing a method error is too cryptic when a user tries to use these functions with analog filters and only one argument, then perhaps we could throw a more informative error, directing users to the right method for analog filters. I'm inclined to agree that it would be a bit cryptic and unfriendly to just throw a method error. I'm all for making the type signature as consistent as possible across `freqresp`, `phaseresp`, and `grpdelay`.

I like the final suggestions quite a lot, though I haven't thought about it very deeply.

martinholters:
To be clear about the default for `w`, I'm totally fine with the limits being `0` and `pi`, those seem pretty natural. (Although if we properly supported complex-valued filters, `-pi` to `pi` would also be reasonable.) It's the length of 250 that seems quite arbitrary to.

martinholters:
Another idea: We could return `w` if it wasn't provided. That is, provide these two:
```julia
resp = freqresp(H, w)
resp, w = freqresp(H)
```
Then changing the default for `w` in the future wouldn't be breaking and for analog filters, we could choose it e.g. based on pole/zero radii.

galenlynch:
I see, yeah you're right that the length of 250 is very arbitrary. Returning `w` if it is not specified seems like a good idea, and probably should be done anyways. In addition to returning `w`, perhaps we should also just make it explicit that the length of the default `w` is not part of the API and is subject to change.

martinholters:
Updated to the "return `w` if not provided" approach and also implemented it for analog filters. (The heuristic for the latter might need tweaking, but that can happen later once actual problems surface, if any.)

martinholters:
To move this forward, I think I'll postpone the question of whether `_freq` should be a method the `FilterCoefficients` subtypes. We can always do that later, but once we have it, it is public API and we cannot take it away anymore that easily.

I'm still skeptical about the `hz, fs` methods of `freqresp`. For analog filters, it seems bogus to me, but it just falls out, so taking it away doesn't save us anything (except for a bit of documentation). But given that `phasez` never had such a method and nobody complained and asked for it, I wonder whether we need it at all. As again, adding public API in the future is easier than taking it away, I'm tempted to just remove this method for now and add it back in case complaints start coming in. Objections?

martinholters:
Barring objections, I'll go ahead and merge this in a few days.

galenlynch:
Thank you!

PR #385, opened 2020-11-06
Fix bugs in corner cases of conversion from ZeroPoleGain to SecondOrderSections

martinholters:
The first two commits fix bugs in corner cases, ~the third commit rewrites the while thing. I'm not sure 100% the rewrite is actually beneficial, so in the likely event I don't get a review on it within reasonable time, I'll just take it out. Therefore, in case anyone has only a little time to spend on this one, review of the first two commits only would also be quite welcome.~

### DIFF
--- a/src/Filters/coefficients.jl
+++ b/src/Filters/coefficients.jl
@@ -324,7 +324,7 @@ function split_real_complex(x::Vector{T}) where T
     r = typeof(real(zero(T)))[]
     for k in keys(d)
         if imag(k) != 0
-            if !haskey(d, conj(k))
+            if !haskey(d, conj(k)) || d[k] != d[conj(k)]
                 # No match for conjugate
                 return (c, r, false)
             elseif imag(k) > 0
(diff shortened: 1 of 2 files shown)


martinholters:
Removed third commit to trim this down to a bug fix (and very minor NFC cleanup). Barring objections, I plan to merge soon as I have other work in the pipeline blocked on this.

src/Filters/coefficients.jl
```diff
@@ -317,9 +322,9 @@ function split_real_complex(x::Vector{T}) where T
 
     c = T[]
     r = typeof(real(zero(T)))[]
-    for k in sort!(collect(keys(d)), by=x -> (real(x), imag(x)))
```
martinholters, line 320: Unrelated clean-up: The order in which this loop proceeds doesn't matter and the results are sorted afterward anyway (and by distance to the unit circle, not by this canonical order).

src/Filters/coefficients.jl
```diff
@@ -317,9 +322,9 @@ function split_real_complex(x::Vector{T}) where T
...
+    for k in keys(d)
```
martinholters, line 325: Unrelated clean-up: The order in which this loop proceeds doesn't matter and the results are sorted afterwards anyway (and by distance to the unit circle, not by this canonical order).

PR #387, opened 2020-11-13
Add test for error path in `PolynomialRatio{:s}` ctor

martinholters:


martinholters:
Coverage of `Filters/coefficients.jl` now at 100%.

galenlynch:
Yay! Thanks for all the work that you've been putting it recently!

PR #392, opened 2020-12-14
Add Github Actions for CI and docs

martinholters:


martinholters:
This runs tests on currently seven Julia versions across five OS/architecture combinations for a total of 35 runs of the test suite. Is that excessive? Given that our tests don't take too long and this repository is not among the most active, I'd say it's acceptable. We could also consider running e.g. all OS/architecture combinations with Julia LTS (`1.0`) and stable (`1`) and the other Julia versions just on one OS/architecture or similar. Opinions?

martinholters:
And bam, this has uncovered something. With Julia nightly on MacOS:
```
Chebyshev II: Test Failed at /Users/runner/work/DSP.jl/DSP.jl/test/FilterTestHelpers.jl:44
  Expression: err1 <= relerr * err2
   Evaluated: 13.17532926311455636223009553964575801024112963471409445842444535807558676549236 <= 12.35065852622911272446019107929151602048225926942818891684889071615117353098472
Stacktrace:
 [1] accuracy_check(err1::BigFloat, err2::BigFloat, part::String, relerr::Int64)
   @ FilterTestHelpers ~/work/DSP.jl/DSP.jl/test/FilterTestHelpers.jl:44
 [2] zpkfilter_accuracy(f1::ZeroPoleGain{:s, ComplexF64, ComplexF64, Float64}, f2::ZeroPoleGain{:z, ComplexF64, ComplexF64, Float64}, accurate_f::ZeroPoleGain{:s, Complex{BigFloat}, Complex{BigFloat}, BigFloat}; relerr::Int64, eps::Nothing)
   @ FilterTestHelpers ~/work/DSP.jl/DSP.jl/test/FilterTestHelpers.jl:87
 [3] macro expansion
   @ ~/work/DSP.jl/DSP.jl/test/filter_design.jl:222 [inlined]
 [4] macro expansion
   @ /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.7/Test/src/Test.jl:1147 [inlined]
 [5] top-level scope
   @ ~/work/DSP.jl/DSP.jl/test/filter_design.jl:140
```
From a brief glance, I haven't seen anything that uses random numbers in this test, so it's unclear why this depends on Julia version/OS.

galenlynch:
Nice! This seems like a great idea.

Been meaning to finish up my faster convolution PR, which I will hopefully do over the holidays.

martinholters:
For anyone (including myself) looking into the test failure: The failing test compares the result of `Chebyshev2(19, 1)` and a reference produced with Matlab to `Chebyshev2(BigFloat, 19, 1)`, in particular the comparison of the gain `k` fails. Locally, I get the same difference `err2` for the Matlab reference, so `Chebyshev2(19, 1)` seems to yield a slightly different `k` on Julia nightly/MacOS. Locally I get
```julia
julia> f1.k
37.33930783884508

julia> loss(f1.k, accurate_f.k)
12.17532926311455636223009553964575801024112963471409445842444535807558676549236
```
which is just below the allowed tolerance. However:
```julia
julia> loss(prevfloat(f1.k), accurate_f.k)
13.17532926311455636223009553964575801024112963471409445842444535807558676549236
```
which is the value of `err1` in the failing test. So the value computed with Julia nightly/MacOS seems to be off by just 1ulp compared to the other runs. So I'd say our `Chebyshev2` should either have higher accuracy everywhere so that we can afford one extra ulp here or there, or we should loosen the tolerance here.

martinholters:
Just re-ran CI to verify the failure reproduces, which it does.

martinholters:
The accuracy problem is remedied with
```diff
--- a/src/Filters/design.jl
+++ b/src/Filters/design.jl
@@ -97,7 +97,7 @@ function Chebyshev2(::Type{T}, n::Integer, ripple::Real) where {T<:Real}
         ze = complex(zero(T), -inv(cospi(w)))
         z[2i-1] = ze
         z[2i] = conj(ze)
-        k *= abs2(p[2i])/abs2(ze)
+        k *= abs2(p[2i] / ze)
     end
     isodd(n) && (k *= -real(p[end]))
 
```
But I don't know whether that is just lucky coincidence or actually is numerically favorable.

That aside, I'm seeing reproducible timeouts during installation of dependencies with Julia nightly on Ubuntu, both 32 and 64 bit. That also affects Travis, though, so there might be something up with Julia?

martinholters:
> But I don't know whether that is just lucky coincidence or actually is numerically favorable.

Prodding at this a bit more, I think it's just lucky coincidence. But I wonder whether our test here is flawed. Consider:
```julia
julia> loss(f.z, accurate_f.z)
6.369039732697331763086323332255388336332929362097955656436434839285139027418413

julia> loss(matlab_z, accurate_f.z)
24.57405098180143395239230488780927719972670649019816761329079532733345373011809

julia> loss(f.p, accurate_f.p)
32.94309177694550608597135747727250740911564359250072639832998462291671447180645

julia> loss(matlab_p, accurate_f.p)
142.0386807343219773467905149178635861530204199936753665503618674707803321521224
```
Both poles and zeros we compute are closer to the result when using `BigFloat` than what Matlab produces. Good. But:
```julia
julia> loss(f.k, accurate_f.k)
12.1753292631145563622300955396457580102411296347140944584244455136509687120209

julia> loss(matlab_k, accurate_f.k)
6.175329263114556362230095539645758010241129634714094458424445513650968712020905
```
So our gain is worse (in the same comparison), and pretty close to the factor of two we allow in the test. Bad? Not necessarily. I'd argue that it is more important that the gain is consistent with the poles and zeros. In particular, we want `k == abs(prod(p)/prod(z))`. Now if we compute the loss against that `k` computed with `BigFloat`, things look different:
```julia
julia> loss(f.k, abs(prod(Complex{BigFloat}.(f.p))/prod(Complex{BigFloat}.(f.z))))
1.347809958931986324930879207469776558361386149924386494295630223360945952758663

julia> loss(matlab_k, abs(prod(Complex{BigFloat}.(matlab_p))/prod(Complex{BigFloat}.(matlab_z))))
10.48911443815117938395987126694023000293543486941194758200991221566153935067621
```
By that measure, we are again better than Matlab.

Let me ponder that a bit more, and if I don't see any problems, I'll open a separate PR to update the test.

martinholters:
Cherry-picked #396 into here and tests pass :tada:

rob-luke:
> 
> 
> This runs tests on currently seven Julia versions across five OS/architecture combinations for a total of 35 runs of the test suite. Is that excessive? Given that our tests don't take too long and this repository is not among the most active, I'd say it's acceptable. We could also consider running e.g. all OS/architecture combinations with Julia LTS (`1.0`) and stable (`1`) and the other Julia versions just on one OS/architecture or similar. Opinions?

I dont think its excessive at all. As this package will be fundamental to many others, detailed testing is worth the slightly extra run time.

There are a number of PRs waiting to be merge. @galenlynch could we get this merged before the other PRs to give extra confidence in the testing?

rob-luke (approved):

.github/workflows/Documentation.yml
```diff
@@ -0,0 +1,22 @@
...
+  pull_request:
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: julia-actions/setup-julia@v1
+        with:
+          version: '1.1'
```
rob-luke, line 17: Should this be? Probably doesnt matter as tests should pass on all versions

```suggestion
          version: '1'
```

rob-luke:
I hope I didn't just commit a faux pas. But I just bumped the tested julia version to kick start the CI again and check it still runs correctly (the initial commit was 4 months ago). I wanted to check its all still green before merging.

.github/workflows/Documentation.yml
martinholters: I think printed output can sometimes change between Julia minor versions, so the doctests might suddenly start failing if we don't fix this to a specific version. At this point in time, we should probably fix it at 1.6, though...

martinholters: ...which indeed happens to break the doctests as tested locally due to changed printing of `Array{Int64,1}` as `Vector{Int64}`. So I'd propose leaving this at 1.1 for now and then doing another PR for bumping it to 1.6.

rob-luke: Excellent, thanks for taking the time to explain in detail.

PR #396, opened 2020-12-17
Better handling of gain in filter accuracy tests

martinholters:
When testing the accuracy of a filter in zero/pole/gain (zpk) form, directly comparing the gain k may be misleading: As k depends on the poles and zeros, deviation in those may have arbitrary effects on k. So a filter that is less accurate in terms of poles and zeros may just accidentally have better accuracy in k.

Therefore, this changes the `zpkfilter_accuracy` function to take an optional keyword parameter `compare_gain_at`, which, if given, lets it compare the filter gain computed at the given frequency (e.g. 0 for a lowpass) instead of plain k. By using this in a few tests, it is possible to tighten the error bounds.

Ref. https://github.com/JuliaDSP/DSP.jl/pull/392#issuecomment-745104583.

### DIFF
--- a/test/FilterTestHelpers.jl
+++ b/test/FilterTestHelpers.jl
@@ -58,7 +58,7 @@ function tffilter_accuracy(f1, f2, accurate_f)
     accuracy_check(loss(a1, accurate_a), loss(a2, accurate_a), "a")
 end
 
-function zpkfilter_accuracy(f1, f2, accurate_f; relerr=1, eps=nothing)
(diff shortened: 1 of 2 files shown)


galenlynch (approved):

galenlynch:
Looks like a great addition. I don't think @simonster is very active on this repo in the last few years, so I'm not sure we should wait on his review. You obviously don't need my permission to merge @martinholters , but I support this! I think you should just merge it whenever you're happy with it.

galenlynch:
I tried to merge this PR but something strange happened with the commit.

galenlynch:
Ah I see, it was cherry picked in #392

PR #401, opened 2021-02-09
add in-place `mt_pgram!`, and multitaper spectrogram, multitaper csd, multitaper coherence 

ericphanson:
This code was orignally written by @ararslan at Beacon Biosignals (and modified by @kolia?) and I've modified it (and possibly introduced mistakes) by adapting it to match DSP.jl's `spectrogram` input and output type, for ease of use. It's still a work in progress because

* needs tests and docstrings
* might need to somehow communicate it is for real data only, analogous to the one-sided spectrogram
* it outputs a `time x freq` matrix, whereas `spectrogram` outputs a `freq x time` matrix

I'm not sure how to test it; is there another open source implementation I can compare against?


galenlynch:
Thanks for the contribution! Could you comment on how it differs or improves upon the [`mt_pgram`](https://github.com/JuliaDSP/DSP.jl/blob/4cf717503e856a11b551ed3e463589dd5889f6d1/src/periodograms.jl#L384-L399) multitaper spectrogram function already in DSP.jl?

galenlynch:
Ah yes, it's a spectrogram. Nevermind!

ericphanson:
Made a bit of progress here today: I reworked things to have an `MTConfig` object that stores the settings and temporary variables for `mt_pgram`, and added an in-place version that `mt_pgram!` that uses it. Then changed the multitaper spectrogram config to use it too. This way we don't have a separate implementation for `mt_pgram` vs `multitaper_spectrogram!` and that you can do both in-place, and even reuse config object s between them. We also have a multitaper coherence implementation that I'd like to add, which can also re-use the same config object.

I also changed the order to be `freq x time` to match spectrogram and added some docstrings. Still needs tests though.

galenlynch:
I'll try to take a look this weekend, sorry for the delay.

ericphanson:
No worries! It wasn't really ready yet before anyway. Now is a good time to take a look though, if you get a chance!

I'll try to address review comments and add tests on Monday (assuming other priorities don't come up first...). The cross-spectra stuff is probably best as a separate PR, so I'll move that out then too (right now it's just commented out and unfinished).

ericphanson:
I'm going to force push to clean up the git history and properly credit some of the code's origins with co-author commits. Hope you don't mind!

ericphanson:
Ok, I think this is ready for review. Thanks for the git help, @palday!

I don't have permission to formally ask for review on github, but I'd appreciate it if @kolia and @ararslan could give this a review too, since it's a big bunch of code. I will start making some PRs to Beacon private repos to try out this branch as well.

Choices I made that possibly should be questioned:

* configuration objects named `MTConfig{T}, MTSpectrogramConfig{T}, MTCrossSpectraConfig{T}, MTCoherenceConfig{T}` hold settings and workspaces. These can be re-used to perform the same operations re-using workspace variables on different signals of the same size (with the same settings). (If you're using multiple threads, you should use a separate workspace for each thread, but note FFTW can be threaded too, so don't oversubscribe!).
* I did not make a default numeric type, e.g. `MTConfig(...) = MTConfig{Float64}(...)`, since it's not clear what the default should be. One option is to make e.g. `MTConfig(signal)` methods to extract size and eltype info from given signal data.
* `allocate_output(config)` produces an output object of the correct size given the configuration
* I did not name the "cross spectra" things as "csd" since there is not a normalization to make it a true "density". Is that annoying / too pendantic? I did not really know what to call those functions.
* I ended up putting the cross spectra and coherence stuff in this PR instead of splitting it out, since I realized some of my decisions were guided by needing that code to work too. However, if it's too bothersome to review this way please let me know and I'll split that out into a separate PR.
* `mt_pgram` now allocates a factor of `ntaper` more memory, because I save the results of each taper separately in `tapered_spectra!` before combining them in `mt_pgram!`. However, that memory is a workspace in the `MTConfig` object, so if you re-use the config you of course save a lot of memory. Why save each taper separately? Because then the same codepath and inner config objects can be reused for `mt_cross_spectral!`. I thought it was simpler and less error-prone this way, but if that's too big of an issue then I can make a separate code path for each case.
* I thought this one was kinda nice: we pass `FFTW.MEASURE` as the default flag when you create a config object, but `FFTW.ESTIMATE` when you use the out-of-place methods that create the config on the fly. The idea is if you're bothering to create a config object, then you're probably going to re-use it, so it's worth spending a bit of time optimizing the plan. If you're not going to re-use it (since you don't even get access to it with the out-of-place methods), then let's just get a quick plan as usual.
* AFAIK, the cross spectra and coherence code is the first multi-channel code in DSP.jl, and thus a convention of which dimension is the channel and which is time is needed. I chose the ordering `channels x samples` since that's what we use in [Onda.jl](https://github.com/beacon-biosignals/Onda.jl), but of course the other ordering is equally valid.
* I tried to make as "simple" as possible by layering the config objects, so for example `MTConfig` settings are used basically whenever you're doing an FFT in this code, and then if you're doing a bunch of them as a spectrogram, `MTSpectrogramConfig` uses an `MTConfig` to govern the individual ones and just holds the configuration needed for how you are combining them.
* I tried to match the existing API, by naming things `mt_*` to match `mt_pgram`, and using `fs`, `nfft`, `nw`, etc, which are existing parameter names. I also tried to match which things are keyword arguments vs positional, especially between creating the config objects and the out-of-place functions. Reviewers, please take a look and see if I missed anything!

## Quick benchmark

Using the first test from `test/periodogram.jl` as example data, we have
```julia
julia> using BenchmarkTools

julia> @btime mt_spectrogram!($out, $x0, $spec_config);
  14.186 μs (1 allocation: 2.12 KiB)

julia> @btime mt_spectrogram(x0, 256, 128; fs=10);
  638.610 μs (64 allocations: 118.73 KiB)

julia> @btime spectrogram(x0, 256, 128; fs=10);
  32.185 μs (47 allocations: 10.73 KiB)
```

You'll note that the in-place one still has one pesky allocation. Not sure where that is, but if someone spots it, let me know! And also the in-place multitaper spectrogram is faster than the existing out-of-place spectrogram, which is kind of fun (since it does like 7x less work), but almost certainly that advantage will go away as the data gets bigger.

src/multitaper.jl
```diff
@@ -0,0 +1,518 @@
...
+    if length(signal) != config.n_samples
+        throw(DimensionMismatch("""Expected `signal` to be of length `config.n_samples`;
+                got `length(signal) == $(length(signal)) and `config.n_samples` = $(config.n_samples)"""))
+    end
+    tmp = config.fft_output_tmp
+    tapered_spectra!(tmp, signal, config)
+
+    output .= 0
+    for j in 1:(config.ntapers)
+        fft2pow!(output, @view(tmp[:, j]), config.nfft, config.r, config.onesided)
```
ericphanson, line 172: here, we make use that `fft2pow!` increments in-place, so calling it repeatedly adds the contribution over each taper

src/multitaper.jl
```diff
@@ -0,0 +1,518 @@
...
+    overlap_duration = n_overlap_samples / fs
+
+    hop = window_duration - overlap_duration
+    time = range(window_duration / 2; step=hop, length=n_time_points)
+    return MTSpectrogramConfig{T,typeof(mt_config)}(n_samples, n_overlap_samples,
+                                                              n_time_points, time,
+                                                              mt_config)
+end
+
+function MTSpectrogramConfig{T}(n_samples, samples_per_window, n_overlap_samples;
```
ericphanson, line 213: order of `samples_per_window, n_overlap_samples` chosen to match `n, n_overlap` ordering of `spectrogram` (and `mt_spectrogram`)

src/multitaper.jl
```diff
@@ -0,0 +1,518 @@
...
+    end
+    if length(signal) != config.n_samples
+        throw(DimensionMismatch("""Expected `signal` to be of length `config.n_samples`;
+                got `length(signal) = $(length(signal)) and `config.n_samples` = $(config.n_samples)"""))
+    end
+    samples_per_window = config.mt_config.n_samples
+    subepochs = arraysplit(signal, samples_per_window, config.n_overlap_samples)
+    for (time_index, subepoch) in enumerate(subepochs)
+        mt_pgram!(destination[:, time_index], subepoch, config.mt_config)
+    end
```
ericphanson, line 246: five lines to do an in-place multitapered spectrogram! I think this kind of shows the point of layering the config objects, since the layout there kind of reflects how the code actually works, i.e. parameters are not mixed between `MTSpectrogramConfig` and `MTConfig` so hopefully it's clear what parameter is needed for what bit of the process.

src/multitaper.jl
```diff
@@ -0,0 +1,518 @@
...
+#####
+##### Multitapered cross spectral matrix
+#####
+
+struct CrossSpectral{T, F}
+    values::T
+    freq::F
+end
+
+freq(c::CrossSpectral) = c.freq
```
ericphanson, line 278: no helper for the `values`, so I guess the user should just use-field access. Open to adding something if there's a name that makes sense

src/multitaper.jl
```diff
@@ -0,0 +1,518 @@
...
+Any keywords accepted by [`MTConfig`](@ref) may be passed here.
+"""
+function MTCrossSpectraConfig{T}(n_channels, n_samples; fs=1, demean=true, low_bias=true, freq_range=nothing, nw=4, ntapers = 2 * nw - 1, kwargs...) where T
+    if demean
+        mean_per_channel = Vector{T}(undef, n_channels)
+        demeaned_signal = Matrix{T}(undef, n_channels, n_samples)
+    else
+        mean_per_channel = nothing
+        demeaned_signal = nothing
+    end
```
ericphanson, line 315: I don't know how standard demeaning the input to a csd is, but MNE does it and we wanted to match their coherences, so I wanted the option to do it here. Of course, you could just ask the user to do it, but since we have all the structure here anyway to do things in-place and allocation-free we might as well offer it for this.

src/multitaper.jl
```diff
@@ -0,0 +1,518 @@
...
+    
+    window = dpss(n_samples, nw, ntapers)
+    evals = dpsseig(window, nw)
+
+    if low_bias
+        keep_evals_mask = evals .> 0.9
+        window, evals = window[:, keep_evals_mask], evals[keep_evals_mask]
+    end
```
ericphanson, line 323: again, an option borrowed from MNE

src/multitaper.jl
```diff
@@ -0,0 +1,518 @@
...
+    if config.demean
+        mean!(config.mean_per_channel, signal)
+        config.demeaned_signal .= signal .- config.mean_per_channel
+        signal = config.demeaned_signal
+    end
+
+    x_mt = config.x_mt
+
+    for k in 1:config.n_channels
+        tapered_spectra!(x_mt[:,:,k], signal[k, :], config.mt_config)
```
ericphanson, line 375: note these are views (due to the `@views` annotation out front)

src/multitaper.jl
```diff
@@ -0,0 +1,518 @@
...
+    end
+    x_mt[1, :, :] ./= sqrt(2)
+    if iseven(size(signal, 2))
```
ericphanson, line 378: I kind of wonder if this should be `nfft` instead of `size(signal, 2)`, but MNE does it this way and I don't fully understand what it's for 🤷

test/multitaper.jl
```diff
@@ -0,0 +1,156 @@
...
+    # writedlm("data/csd_array_multitaper_values_im.txt", imag(csd_array_multitaper_values))
+    csd_array_multitaper_frequencies = readdlm(joinpath(@__DIR__, "data", "csd_array_multitaper_frequencies.txt"))
+    csd_array_multitaper_values_re = reshape(readdlm(joinpath(@__DIR__, "data", "csd_array_multitaper_values_re.txt")), (2,2,512))
+    csd_array_multitaper_values_im = reshape(readdlm(joinpath(@__DIR__, "data", "csd_array_multitaper_values_im.txt")), (2,2,512))
+    csd_array_multitaper_values = csd_array_multitaper_values_re + im*csd_array_multitaper_values_im
+
+    signal = dropdims(data; dims=1)
+    result = mt_cross_spectral(signal; fs=fs)
+    @test freq(result)[2:end] ≈ csd_array_multitaper_frequencies
+    @test result.values[:,:,2:end] ≈ csd_array_multitaper_values rtol=1e-3
```
ericphanson, line 148: only matches up to `rtol=1e-3`; I think that means it doesn't actually match to numerical precision, but I spent a long time trying to see what was different and failed to find anything more

ericphanson: seems to work now!

ericphanson:
I've been using this at Beacon a bit and have been ironing out edge cases and things like that, but I think it should be fairly solid at this point. But AFAIK no one else has tried out the API or given feedback on that, so it would be great to hear what other people think.

I also added another `mt_spectrogram` method for reusing the `MTConfig` object used for the FFTs within the window, which should be pretty re-usable (if you are doing a bunch of spectrograms on possibly different sized signals but with the same window size) and improve performance in that case since it then only needs to allocate the spectrogram output but nothing internally.

galenlynch:
I'm going to start looking this over now, sorry for the delay.

src/multitaper.jl
```diff
@@ -0,0 +1,583 @@
...
+    evals = dpsseig(window, nw)
+
+    if low_bias
+        keep_evals_mask = evals .> 0.9
+        window, evals = window[:, keep_evals_mask], evals[keep_evals_mask]
+    end
+    ntapers = size(window, 2)
+
+    denom = fs * sum(evals) / 2
+    weighted_evals = evals ./ denom
```
galenlynch, line 350: Could you provide a reference for weighting each Slepian taper before using it to estimate the spectra? I'm unfamiliar with this, and it's not used in any other spectral estimation code that I've seen.

galenlynch (commented):
I think this is likely to be the starting point of the review rather than the conclusion, but this is a very, very nice addition to DSP.jl.

I've learned some things about multitaper estimation by reviewing this PR, but I would like to read and discuss more before setting some things as the default. Specifically, I'm unfamiliar with weighting each Slepian taper by its eigenvalue when doing spectral estimation, which seems to be how cross spectral densities are calculated in this PR. I'm also unfamiliar with the 'low bias' flag in this PR, which discards Slepian tapers based on their eigenvalue instead of taking the first 2NW - 1 of them. I would be grateful for any references you could provide so we could better discuss whether these are good defaults. I will perhaps have to change how I do cross-spectral estimation after reading more!

I think the terminology of the names, fields, and documentation for many cross-spectral estimation is a bit vague on what is actually being computed. Presumably you are computing the cross power spectral density, but from reading the documentation it was a little bit hard for me to discern that.

I also think that users should be invited to access the `values` field, and we should provide an accessor function instead. Perhaps `power`?

I'm curious how you ensure FFTW alignment using this code, as you do not use any FFTW malloc function. I'm sure this is ignorance on my part, but does Julia already allocate arrays so that they have the correct SIMD alignment? I've read that this can involve both placing the start of the array on a 64 byte boundary, but also putting each element on a 64 byte boundary. I'm not sure exactly what FFTW expects for both Float32 and Float64 inputs, so if you could help me understand how you're assuring this alignment I would be very grateful.

Additionally, I think it would be good to not de-mean the inputs by default, but keep that option for users. However, I'm definitely open for discussion on this and all of the other points that I've made so far.

Very nice work, thank you for spending all this time on it, and sorry that it's taken me so long to look at this in detail.

src/multitaper.jl
ericphanson: I based this off of MNE's `csd_multitaper` method. I saw MATLAB's `pmtm` also provides an option to weight by the eigenvalues, https://www.mathworks.com/help/signal/ref/pmtm.html?searchHighlight=psd%20multitaper&s_tid=srchtitle#btultt2-method, but I don't know specifically how they do it (and that is not for cross-spectra). I found it discussed briefly in Section 3.3.1 of this reference as well:

Candy, J. V. “MULTITAPER SPECTRAL ESTIMATION: An Alternative to the Welch Periodogram Approach.” Lawrence Livermore National Lab. (LLNL), Livermore, CA (United States), August 19, 2019. https://doi.org/10.2172/1560107.

Would it be better to provide an option about whether or not to weight it, and how to do so?

src/multitaper.jl
```diff
@@ -0,0 +1,583 @@
...
+"""
+    MTCrossSpectraConfig{T}(n_channels, n_samples; fs=1, demean=true, low_bias=true, freq_range=nothing, nw=4, kwargs...) where T
+
+Creates a configuration object used for [`mt_cross_spectral!`](@ref) as well as [`mt_coherence!`](@ref).
+
+
+* `n_samples`: the number of samples to be used as input when computing multitaper periodograms
+    with this configuration. Used for pre-allocating temporary buffers.
+* `fs`: the number of samples per second of the input signal
+* `demean`: if `true`, the channelwise mean will be subtracted from the input signals before the cross spectral values are computed.
```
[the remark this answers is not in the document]
ericphanson: Ok, I'll make that change! No problem

src/multitaper.jl
```diff
@@ -0,0 +1,583 @@
...
+    fft_input[(length(signal) + 1):(config.nfft)] .= 0
+
+    # do the FFT
+    mul!(fft_output, config.plan, fft_input)
+    return nothing
+end
+
+
+"""
+    mt_pgram!(output::AbstractVector, signal::AbstractVector, config::MTConfig) -> Periodogram
```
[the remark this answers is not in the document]
ericphanson: I think there are four possible methods:

* out of place, pass a config
* in place, pass a config (*)
* out of place, construct the config inside the function (*)
* in place, construct the config inside the function

where the (*) methods are currently implemented (except `mt_spectrogram`, which has a variant on "out of place, pass a config": it has an out of place method where you pass a `MTConfig` instead of a `MTSpectrogramConfig`).

I chose these two as extremes of max-convienence vs max-perf. I can add "in place, construct the config inside the function" as you suggest too, or even both the other options. I was a bit worried about it being too confusing if there are many methods but perhaps it's nice to have all the options.

src/periodograms.jl
```diff
@@ -1,10 +1,16 @@
...
-export arraysplit, nextfastfft, periodogram, welch_pgram, mt_pgram,
-       spectrogram, power, freq, stft
+using Statistics: mean!
+export arraysplit, nextfastfft, periodogram, welch_pgram,
+       spectrogram, power, freq, stft,
+       MTConfig, mt_pgram, mt_pgram!,
+       MTSpectrogramConfig, mt_spectrogram, mt_spectrogram!,
+       MTCrossSpectraConfig, mt_cross_spectral, mt_cross_spectral!,
+       MTCoherenceConfig, mt_coherence, mt_coherence!,
+       allocate_output
```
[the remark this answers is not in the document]
ericphanson: Sure, I can do that

src/multitaper.jl
```diff
@@ -0,0 +1,583 @@
...
+    MTCrossSpectraConfig{T}(n_channels, n_samples; fs=1, demean=true, low_bias=true, freq_range=nothing, nw=4, kwargs...) where T
+
+Creates a configuration object used for [`mt_cross_spectral!`](@ref) as well as [`mt_coherence!`](@ref).
+
+
+* `n_samples`: the number of samples to be used as input when computing multitaper periodograms
+    with this configuration. Used for pre-allocating temporary buffers.
+* `fs`: the number of samples per second of the input signal
+* `demean`: if `true`, the channelwise mean will be subtracted from the input signals before the cross spectral values are computed.
+* `low_bias`: if `true`, keeps only tapers with eigenvalues > `0.9`.
```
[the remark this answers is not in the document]
ericphanson: This option was also borrowed from MNE. Based on the LLNL report I cited elsewhere, 

Candy, J. V. “MULTITAPER SPECTRAL ESTIMATION: An Alternative to the Welch Periodogram Approach.” Lawrence Livermore National Lab. (LLNL), Livermore, CA (United States), August 19, 2019. https://doi.org/10.2172/1560107.

the smaller the eigenvalue is relative to 1, the larger the spectral leakage (my interpretation of the bottom of p. 18 there).

ericphanson (commented):
Thanks a lot for the review @galenlynch!

I should say, many of the choices made by the cross-spectral code + coherence code are based on those made by `mne-python`'s <https://mne.tools/stable/generated/mne.time_frequency.csd_multitaper.html#mne.time_frequency.csd_multitaper>. I think I should have added that to a comment to credit it. In particular, the `low_bias` option is taken directly from the keyword argument there, and the choice of weighting by the eigenvalues was borrowed from them as well. I've written in more detail in some of the comments.

> I think the terminology of the names, fields, and documentation for many cross-spectral estimation is a bit vague on what is actually being computed. Presumably you are computing the cross power spectral density, but from reading the documentation it was a little bit hard for me to discern that.

I should say that I am fairly new to this field and am confused by the terminology myself, and that probably unfortunately is reflected in the lack of information here. I think that is the root issue behind `values` being a bad name etc. So I'm very glad you commented on this so we can ensure it is more clear by the time this is merged.

As for what it does: I tried to match the results of `mne.time_frequency.csd_multitaper_array` (as you can see in the tests), and there they call it a "cross spectral density". However, I don't really see how it is a density. There is no normalization by time / length of the signal, or by total energy or anything like that. So I am still a bit confused here and will try to consult my colleagues. If you have any insight I'd be glad to hear it.

galenlynch:
After reading a excerpt of [1] it seems like weighting the taper estimates by their eigenvalue is standard, but has a negligible impact if you only take the first 2NW-1 tapers. It seems like we should probably give users the option to use that weighting or not, but also document what we're doing. It would be nice to have all multitaper estimation use the same weighting or have the same weighting interface, instead of changing the weight estimation is done between spectral estimation and cross power spectral density.

[1] Percival, Donald B., and Andrew T. Walden. Spectral Analysis for Physical Applications: Multitaper and Conventional Univariate Techniques. Cambridge; New York, NY, USA: Cambridge University Press, 1993.

src/multitaper.jl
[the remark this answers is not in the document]
ericphanson: Fixed in https://github.com/JuliaDSP/DSP.jl/pull/401/commits/b25b6309ffeb396a9d443d94264dcfb14c8d1f4f

src/multitaper.jl
```diff
@@ -0,0 +1,583 @@
...
+    return Matrix{real(T)}(undef, config.cs_config.n_channels, config.cs_config.n_channels)
+end
+
+"""
+    coherence_from_cs!(output::AbstractMatrix, cs_matrix)
+
+Compute the pairwise channel coherences from a cross spectral matrix, storing the result in `output`
+and returning a `Symmetric` view of `output`.
+"""
+function coherence_from_cs!(output::AbstractMatrix{T}, cs_matrix) where T
```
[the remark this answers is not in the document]
ericphanson: Fixed in https://github.com/JuliaDSP/DSP.jl/pull/401/commits/808577a28e938811516bff15cff80db3fa223198

test/multitaper.jl
```diff
@@ -0,0 +1,244 @@
...
+        n_samples = 100
+        nfft = 200
+        ntapers = 5
+        fs = 1
+        window = rand(ntapers, n_samples)
+        for T in (Float32, Float64, Complex{Float32}, Complex{Float64})
+            fft_input_tmp = Vector{T}(undef, nfft)
+            onesided = T <: Real
+            fft_flags = FFTW.ESTIMATE
+            freq = onesided ? rfftfreq(nfft, fs) : fftfreq(nfft, fs)
```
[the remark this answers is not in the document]
ericphanson: Fixed in https://github.com/JuliaDSP/DSP.jl/pull/401/commits/af9464ad8db336cf8d036279c150111fda5040b2

src/periodograms.jl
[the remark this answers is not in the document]
ericphanson: Fixed in https://github.com/JuliaDSP/DSP.jl/pull/401/commits/c28e766cc374756d7148d2d03db21f600a971034

ericphanson:
> After reading a excerpt of [1] it seems like weighting the taper estimates by their eigenvalue is standard, but has a negligible impact if you only take the first 2NW-1 tapers. It seems like we should probably give users the option to use that weighting or not, but also document what we're doing. It would be nice to have all multitaper estimation use the same weighting or have the same weighting interface, instead of changing how the weight estimation is done between spectral estimation and cross power spectral density.

Thanks very much for the reference! Very helpful for me too. I agree consistency + options are better than varying arbitrary choices 😅 . I will improve the docstrings and add the option to the MTConfig object and thread it through everywhere (... tomorrow, hopefully).

src/multitaper.jl
```diff
@@ -0,0 +1,604 @@
...
+Fields:
+
+* `coherences::A`: `n_channels` x `n_channels` x `length(freq)` array consisting of the pairwise coherences between each channel, for each frequency
+* `freq::F`: frequencies; accessed by `freq(::Coherence)`
+
+"""
+struct Coherence{T,F,A<:AbstractArray{T, 3}}
+    coherences::A
+    freq::F
+end
```
ericphanson, line 556: Since we aren't averaging over frequency anymore, I thought it was better to return a structured object that holds the frequencies, like we do in other cases. Should I add an accessor like

```julia
coherence(c::Coherence) = c.coherences
```
?

src/multitaper.jl
```diff
@@ -0,0 +1,604 @@
...
+    @inbounds for ch2 in 1:n_channels
+        for ch1 in (ch2 + 1):n_channels # lower triangular matrix
+            for f in 1:n_freqs # average over frequency
+                output[ch1, ch2, f] += real(abs(cs_matrix[ch1, ch2, f]) /
+                                         (sqrt(cs_matrix[ch1, ch1, f] *
+                                               cs_matrix[ch2, ch2, f])))
+            end
+        end
+    end
+    output .+= PermutedDimsArray(output, (2, 1, 3)) # symmetrize
```
ericphanson, line 536: we don't have a `Symmetric` type like in the matrix case, so I just symmetrize it by hand. If I haven't messed things up this should at least be non-allocating, and probably the expensive bit is all the FFTs anyway, so I think it's probably not that bad to copy in the right values here.

src/periodograms.jl
```diff
@@ -1,13 +1,18 @@
 # The periodogram module contains functions which compute non-parametric
 # estimates of the periodogram P[s] of a signal s.
 module Periodograms
-using LinearAlgebra: mul!
+using LinearAlgebra: mul!, Symmetric, diagind
```
ericphanson, line 4:
```suggestion
using LinearAlgebra: mul!
```
don't need these anymore

src/multitaper.jl
galenlynch: Sounds good to me!

src/multitaper.jl
ararslan: Typically wouldn't you divide by 2 when symmetrizing this way?

ericphanson: Not here because we only fill in the lower half (and not the diagonal), and leave the upper half zero. So then when we add the “transpose” here it just copies the lower half up to the upper half. (Then we set the diagonal to 1).

src/multitaper.jl
ericphanson: Seems to be an MNE bug: https://github.com/mne-tools/mne-python/issues/9350

src/multitaper.jl
galenlynch: Did you figure out if this should be `nfft` or the size of the signal? It seems to me like it should be `nfft`.

galenlynch:
Just seeing where this stands, were you going to add an option to weight each taper's estimation, and add a similar option to multitaper estimation beyond spectrograms?

ericphanson:
> Just seeing where this stands, were you going to add an option to weight each taper's estimation, and add a similar option to multitaper estimation beyond spectrograms?

Yep! Just pushed up a commit. Some updates:

* I am pretty sure the `cross_spectral` code is comupting power not energy (i.e. there's a normalization by the length of the sequence), and that's built into the tapering. I wrote up my understanding of this here: https://mne.discourse.group/t/understanding-mne-time-frequency-csd-multitaper-implementation/3043/2. Therefore I've adjusted the names to reflect that.
* The `low_bias` and eigenvalue weighting stuff seems a bit ad-hoc so I've pulled it out of the main API. At the moment, it's exposed by a `dpss_config` function to create a config with those parameters, which we need for the reference tests. I'm not totally sure how it should be exposed.
* The parameter `r` already is used for normalization, so I've expanded it to allow per-taper normalization involving any desired `taper_weights`.
* I've added methods so we have all four of the methods discussed in https://github.com/JuliaDSP/DSP.jl/pull/401#discussion_r616799486

src/multitaper.jl
ericphanson: Done!

src/multitaper.jl
```diff
@@ -0,0 +1,583 @@
...
+
+Compute a multitapered cross-spectral matrix.
+
+* `signal`: `n_channels` x `n_samples`
+
+Any keyword arguments accepted by [`MTCrossSpectraConfig`](@ref) may be passed here.
+See also [`mt_cross_spectral!`](@ref).
+
+Produces a `CrossSpectral` object holding the `n_channels` x `n_channels` x `n_frequencies`
+output array (in the `values` field) and the corresponding frequencies (accessed by [`freq`](@ref)).
```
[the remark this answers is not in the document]
ericphanson: Now it's called `power` and I removed the reference to the field

src/multitaper.jl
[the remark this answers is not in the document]
ericphanson: I added the remaining methods so we have all 4 for each function in the latest commit

ericphanson:
I am guessing the codecov will take a hit with the new methods I added, so I'll add tests for them. I pushed up what I had so far so you could take a look already in case you had any comments.

src/multitaper.jl
ericphanson: done in the latest commit

src/multitaper.jl
ericphanson: I've updated the code here to use `nfft` instead

ericphanson:
I've added tests for the new methods and also added an error if the user does `mt_cross_spectra` (or `mt_coherence`, which uses that method) with complex input or a two-sided FFT, because I realized the current code assumes a one-sided FFT. I don't think it should be that hard to generalize it, but I'd prefer to leave that for later work in the interest of getting this in quicker.

src/multitaper.jl
ericphanson: I am pretty sure it should be `nfft`; I've filed an issue with MNE and someone there agreed as well. So I've changed it here to be `nfft`.

SimonDanisch:
Would be amazing to merge this :) Is there anything we can do to help?
@ararslan, @kolia, could you take another look at this, now that it's ready to get merged?

rob-luke:
Can you add the new functions to the docs? Unless I missed it in the long and impressive file changes.

ericphanson:
> Can you add the new functions to the docs?

Good point, done!

ericphanson:
Bump 🙂

ericphanson:
Would really love a re-review @galenlynch! Or maybe another DSP maintainer can help out, e.g. @ararslan?

I've got a pile of branches depending on this branch and they've started to propagate further downstream...

galenlynch:
Hey sorry that I disappeared. I'm happy with this now! I'll approve the PR and merge it tomorrow unless someone objects.

galenlynch (approved):

ericphanson:
Whoo! Very exciting :). Could we tag a release?

PR #408, opened 2021-03-08
modifications for breaking version of Polynomials

jverzani:
The just-tagged version 2.0.0 of `Polynomials` contains some breaking changes. Germaine to this project is the inclusion of the symbol into the type hierarchy. (You already have done this in the impacted `PolynomialRatio` type.) The modifications necessary were fairly minimal: the internal means to get the indeterminate is not through a method call, not field access; the use of `LaurentPolynomial{T,Domain}` in `PolynomialRatio` gives a concrete type.


martinholters (approved):
I generally approve, but wonder if this might block any downstream packages from updating their DSP because they depend on Polynomials v1. I don't think we need to worry about this too much, but wanted to point it out in case anyone goes "Oh yes, we can't do that!".

Is there any merit in restricting the argument types in functions to polynomials in the right variable? I guess that only changes the error thrown so can still do that later, just asking.

jverzani:
The blocking issue is definitely something that comes up (e.g., in another package I have `FastTransforms` holds things up). The first change to use `indeterminate` is in v2.0 or in v1.2.1. The change in the type though is only v2.0. I'm not sure exactly what you mean by restricting argument types to the right variable, but if you think that helps great. It could also just not be changed at the expense of a non-concrete type in the struct.

martinholters:
The next release of DSP.jl will be breaking anyway (and I have no idea about the timeline), so I guess I'm fine with just requiring Polynomials v2. If that does cause any troubles downstream, it's good to know that v1.2.1 might be sufficient if we accept the non-concrete member type.

> I'm not sure exactly what you mean by restricting argument types to the right variable

Those (few) functions in DSP.jl that accept a `LaurentPolynomial` as argument could be restricted to taking a `LaurentPolynomial{<:Number,Domain}` (for the proper value of `Domain`). E.g. constructing a `PolynomialRatio{:s}` from `LaurentPolynomial{Float64,:z}` should throw, and we could just make it a `MethodError`. I'm not saying we should, just that we could.

Anyway, I'm fine with merging as is, so lest someone raises objections, or beats me to it, or I forget, I'll merge in a day or so.

Issue #410, opened 2021-03-14
Integer convolutions have rounding errors


rpigott:
DSP contains the following definition:

```julia
conv(u::AbstractArray{<:Integer, N}, v::AbstractArray{<:Integer, N}) where {N} =
    round.(Int, conv(float(u), float(v)))
```

The float conversion and subsequent rounding introduces rounding errors when the floating point epsilon near the computed values is greater than 1, even though it returns an array of Int64. For example, the following result is unexpected to me:

```julia
using DSP
n = 314159265 # n < isqrt(typemax(Int)), no overflow
conv([n], [n]) == [n^2] # false, rounding error!
```

With a "naive" direct convolution, there is no error:

```julia
function dconv(a, b)
  na, nb = length(a), length(b)
  b = [b; zeros(eltype(b), na - 1)]
  a = [a; zeros(eltype(a), nb - 1)]
  [sum(a[i]*b[k - i + 1] for i in 1:k) for k in 1:(na+nb-1)]
end
n = 314159265
dconv([n], [n]) == [n^2] # true
```

The julia FAQ correctly (imo) [identifies many reasons](https://docs.julialang.org/en/v1/manual/faq/#faq-integer-arithmetic) why integer overflow behavior is an acceptable error where rounding/truncation is not; it may even be desirable. For julia, I also think it's not very good in terms of type stability, considering the following:

```julia
a = UInt32[1, 2, 3]
typeof(a) # Array{UInt32, 1}
typeof(conv(a, a)) # Array{Int64, 1}
```

since it always casts to Int, regardless of the input integer element type. FWIW, scipy uses machine integer arithmetic:

```python
> import numpy as np
> import scipy.signal as ss
> n = 314159265
> ss.convolve([n], [n]) == [n**2]
array([True])
> n = 2**32
> ss.convolve([n], [n])
array([0]) # int64 overflow
```

There are a couple ways to tackle this:

1. Remove the above definition of conv for integer type arrays. Users who are OK with rounding errors can perform the cast themselves, and are more likely to avoid unexpected type stability issues this way if they use an integer type other than signed Int64.
[...]

PR #411, opened 2021-03-14
Make conv return floats when given integer arrays

rpigott:
Ref #410, suggestion from the comments.

This removes an unnecessary round. Dispatch for this case is now handled the same as for arrays of generic Number type, so the return type of conv([1, 2, 3], [4, 5, 6]) is now Float64 instead of Int, correctly reflecting the internal precision. Users can get the old behavior by calling round themselves, i.e. `round.(Int, conv(a, b))` or perhaps even `round.(promote_type(eltype(a), eltype(b)), conv(a, b))` for other integer types.

If DSP learns a method of integer convolution in the future, the return type for that method should be an integer.

### DIFF
--- a/src/dspbase.jl
+++ b/src/dspbase.jl
@@ -702,9 +702,6 @@ function conv(u::AbstractArray{<:BLAS.BlasFloat, N},
     conv(fu, fv)
 end
 
-conv(u::AbstractArray{<:Integer, N}, v::AbstractArray{<:Integer, N}) where {N} =
-    round.(Int, conv(float(u), float(v)))
-
 conv(u::AbstractArray{<:Number, N}, v::AbstractArray{<:Number, N}) where {N} =
     conv(float(u), float(v))
 


galenlynch:
I'm hoping for comments on this PR, I think this is worth discussing before merging.

galenlynch:
Any comments?

ssfrr:
I know that convolution is used in non-DSP contexts for other purposes (like multiplying huge numbers) and in those contexts true integer convolution makes sense, and should probably be implemented without doing FFTs.

That said, this is a DSP package and if I remember correctly, supporting the integer case has added complexity in the past. I think it's reasonable to say that convolution returns floats.

galenlynch:
Well I'm in favor of making this change for now until we have real integer convolution. Could you please fix the tests?

martinholters:
I'd like to have a true integer convolution, but while we don't have it, being honest and returning floats seems better than rounding. So generally in favor, unless someone surprises me by submitting a proper integer-capable convolution.

galenlynch:
I have one that is basically complete, though I've stalled out on how to select which algorithm to use. I've added a bunch of direct convolution algorithms that perform differently for different inputs, and have also added a bunch of functions that try to estimate the run time based on some simple fits that I did to a battery of benchmarks. 

I would love some help brain-storming how to structure how to pick the best algorithm for a given input size and type. The code is on this branch: https://github.com/galenlynch/DSP.jl/blob/merge_conv_filt/src/dspbase.jl

I can make a draft PR soon and discuss where I've stalled out.

danilo-bc:
> I'd like to have a true integer convolution, but while we don't have it, being honest and returning floats seems better than rounding. So generally in favor, unless someone surprises me by submitting a proper integer-capable convolution.

I agree with @martinholters. Integer convolutions have their niche, but if there's no proper true integer convolution, it's better to be fair and return a float.

martinholters:
Hers is a simple N-D time-domain convolution I've cooked up:
```julia
function tdconv(u::AbstractArray{<:Number, N}, v::AbstractArray{<:Number, N}) where {N}
    return [
        sum(u[m] * v[n-m]
            # valid ranges (u and v assumed zero outside):
            #    firstindex(u) <= m <= lastindex(u)
            # and
            #    firstindex(v) <= n-m <= lastindex(v)
            # or equivalently
            #    n-lastindex(v) <= m <= n-firstindex(v)
            for m in CartesianIndices(ntuple(d ->
                max(firstindex(u,d),n[d]-lastindex(v,d)):min(lastindex(u,d), n[d]-firstindex(v,d)),
                N
            ))
        )
        for n in CartesianIndices(ntuple(d -> 2:size(u,d)+size(v,d), N))
    ]
end
```
I hope to find some time to hook it up so that `conv`  dispatches to it for integer input and probably also for small-sized float inputs, but I need to do some benchmarking to figure out what a reasonable cut-off would be.

martinholters:
> Should be obsoleted by #545. Close?

Should be auto-closed when #545 is merged, but if you prefer, you can also close, and I think we're pretty sure we'll go with #545 once that is fully in shape.

rpigott:
#545 looks good! Thanks for working on this!

Issue #400, opened 2021-02-02
Add Dirichlet kernel `diric`


JeffFessler:
It seems that right now no Julia package has the Dirichlet kernel (aka periodic sinc) in it:
https://encyclopediaofmath.org/index.php?title=Dirichlet_kernel

Matlab and SciPy have the related `diric` function, and julia's Base.Math has the ordinary `sinc` in it.

DSP.jl seems the right place for it.  I can make a PR if the maintainers agree it is appropriate.

Note that the wikipedia version has an unnecessary extra 2\pi factor in it
https://en.wikipedia.org/wiki/Dirichlet_kernel

In contrast, Matlab and SciPy have a more natural 1/N factor:
https://www.mathworks.com/help/signal/gs/the-dirichlet-function.html
https://docs.scipy.org/doc/scipy/reference/generated/scipy.special.diric.html

Other treatments I have seen (like the 1974 Levy paper and the math encyclopedia above) do not include such factors:
http://doi.org/10.1111/j.2164-0947.1974.tb03023.x

I'd be inclined to match SciPy and Matlab in name (`diric`) and scaling to facilitate code translation.
Comments?

PR #412, opened 2021-03-20
add Dirichlet kernel diric

JeffFessler:
Addresses #400 
It is desirable to have this in a repo because naive implementations based on the Wikipedia definition alone will fail near multiples of 2\pi.


galenlynch:
Thanks for doing this! I'll review it soon.

JeffFessler:
Hi @galenlynch, just thought I'd ping you about this...

rob-luke:
@JeffFessler could you please add this to the docs.

src/diric.jl
```diff
@@ -0,0 +1,50 @@
+# diric.jl Dirichlet kernel
+# 2021-03-19, Jeff Fessler, University of Michigan
```
rob-luke, line 2: These comments are normally not included in the code. Please follow the format from the rest of the repository.

Your contribution will be visible in the git logs. And also in the contributors list generated on the GitHub website.

JeffFessler: done

JeffFessler:
@rob-luke and @galenlynch, do y'all need some assistance maintaining this package?
I teach DSP and I would really like students to switch from Matlab to Julia, but that really
means it needs to have an actively maintained DSP package.  How can I help?

rob-luke:
> it needs to have an actively maintained DSP package

I think its unfair to accuse this package of being unmaintained. It has had constant attention since 2012, almost 10 years. Please be understanding that the time to review a PR varies based on the complexity of the change and also external factors (I assume most of us are academics and have less time to review during grant season for example, we also have personal lives). The considered approach to merging changes is what has maintained the high quality of this repository.

> How can I help?

You already are by submitting a PR! Other things that can help are reviewing other PRs, we usually try and get two approvals before merging. Similarly, triaging the issues and suggesting to close items that have already been addressed is helpful too. Also as you've already done, a gentle reminder via an @ ping can help to remind people to review can help too 😄

src/diric.jl
```diff
@@ -0,0 +1,47 @@
+export diric
```
rob-luke, line 1: Just a comment on code organisation for @martinholters to decide on.

Given that this function is documented in `utils` should the code go in `src/utils.jl`? Similarly, for the tests?

JeffFessler:
Sorry, I meant this to be an expression of concern, not an accusation.  I realize that people have a lot going on in their lives especially right now.  So from my perspective when I have a PR that has been there for almost 3 months it leaves the impression that the maintainer's attention is elsewhere (quite understandably) so perhaps it would be a relief to have more hands on deck.  I'll try a simpler @ ping next time.  (I read somewhere else that 2 weeks considered reasonable for sending another ping.)  Thanks for the quick reply.

src/diric.jl
rob-luke: Ops I meant @galenlynch sorry

rob-luke:
>  more hands on deck.

Absolutely! Like I said, reviewing PRs and addressing issues is a great help.

src/diric.jl
```diff
@@ -0,0 +1,47 @@
...
+the return type matches that of `Ω`.  Otherwise the return type is `Float64`.
+
+As of 2021-03-19, the Wikipedia definition has different factors.
+The definition here is consistent with scipy and other software frameworks.
+This implementation treats inputs near multiples of 2π fairly carefully.
+"""
+function diric(Ω::T, n::Int) where T <: AbstractFloat
+    n <= 0 && throw(ArgumentError("n=$n is non-positive"))
+    denom = sin(Ω / 2)
+    if isapprox(1 + denom, 1) # denom ≈ 0 ?
```
martinholters, line 28: This is a rather indirect condition. I'd propose `isapprox(denom, 0; atol=...)` instead, where I'm not sure what to use for the `...` here to be equivalent. (Which is part of the problem I'm having with this as it is now). Actually, there is no value that is exactly equivalent as
```julia
julia> denom = 1.4901161342859267e-8
1.4901161342859267e-8

julia> isapprox(1 + denom, 1)
true

julia> denom = -denom
-1.4901161342859267e-8

julia> isapprox(1 + denom, 1)
false
```
Whereas the proposed alternative would be symmetric around zero (which also seems to make more sense.)

src/diric.jl
```diff
@@ -0,0 +1,47 @@
...
+        if isodd(n)
+            ω = mod2pi(Ω + π) - π # (-π,π)
+        else
+            π2 = T == BigFloat ? 2T(π) : T(2π) # for type stability
+            π4 = T == BigFloat ? 4T(π) : T(4π)
```
martinholters, line 33: Doesn't `2T(π)` (or `4T(π)`) work here irrespective of `T`?

martinholters (changes requested):
Haven't checked the math, but superficially looks good to me except for a few minor points.

src/diric.jl
JeffFessler: Yes it does; not sure why I though I needed to do that.  Fixed - thanks!

src/diric.jl
JeffFessler: Very good points.  I agree that `atol` is preferable here for symmetry and clarity.
Because `isapprox` has a lot of unnecessary stuff in it here, the updated PR uses an `atol` that is provided as a keyword option with the most sensible default (that stems from `isapprox` code).
It is pretty unlikely that any user will ever use anything other than that default,
but I prefer not to bury constants inside functions...  See what you think about the new version.
Thanks!

src/diric.jl
JeffFessler: It's up to you of course, but personally I prefer to keep unrelated methods in different files.

src/diric.jl
```diff
@@ -0,0 +1,55 @@
...
+"""
+    kernel = diric(Ω::Real, n::Int; atol)
+
+Dirichlet kernel, also known as periodic sinc function,
+where `n` should be a positive integer.
+Returns `sin(Ω * n/2) / (n * sin(Ω / 2))` typically,
+but `±1` when `Ω` is a multiple of 2π.
+
+In the usual case where 'n' is odd, the output is equivalent to
+``1/n \\sum_{k=-(n+1)/2}^{(n+1)/2} e^{i k Ω}``,
```
martinholters, line 12:
```suggestion
``1/n \\sum_{k=-(n-1)/2}^{(n-1)/2} e^{i k Ω}``,
```
Alternatively, `e^{i (n-1) Ω/2}/n \\sum_{k=0}^{n-1} e^{i k Ω}` could be used, which would also hold for even `n`.

martinholters:
I've just tried this and it looks like the accuracy is very good around zero, but deteriorates at higher multiples of 2π. I'd therefore propose to unconditionally first do the wrapping of Ω to within [-π, π] for odd `n` and [-2π, 2π] for even `n`.

src/diric.jl
```diff
@@ -0,0 +1,55 @@
...
+
+This implementation treats inputs near multiples of 2π fairly carefully.
+If the denominator `sin(Ω/2)` is close to 0, then this function uses
+an accurate 2nd-order Taylor expansion rather than simple division. 
+
+The optional argument `atol` defaults to `sqrt(eps(T))`
+and is used to assess whether `sin(Ω/2) ≈ 0`.
+Here `T` is the type of the floating-point argument `Ω`.
+This default tolerance originates from [`rtoldefault`](@ref)
+used in [`isapprox`](@ref).
```
martinholters, line 32: Do these references actually work in some context? Building the docs locally, I get warnings about them during build and dead links in the output.

src/diric.jl
martinholters: Actually, letting this sink in for a bit, I don't think `atol` should be a parameter. The function should automatically pick a value based on the input type. This is not a parameter one might want to change based on context. Ideally, it should be chosen at the cross-over point where the naive formula becomes less exact than the Taylor approximation.

So I tried to figure out where that is. Around zero, it seems the error of the naive formula should be bounded by `2eps(T)`. (And some numerical experiments seem to verify this.) Given that the correct result is close to one, just the numerical error of evaluating the Taylor approximation might be comparable, so even in a range where the error due to the truncation of the series is small, the Taylor approximation might still not be beneficial. I'm not sure what happens when subnormals come into play, though. But in that case, probably `sin(x)==x`, so the naive formula would give `1.0` which would also probably the closest floating point value. So that indicates that only the case where `Ω/2==0` would require special treatment. A small but non-zero `atol` would feel safer though; `atol=eps(T)` comes to mind. Or we estimate the error of the series truncation by the first omitted term and see where that exceeds `eps(T)`, but that might be overly complicated.

For odd `n`, if we first map Ω to within (-π,π) (as already suggested below), the behavior around zero would all we have to think about. For even `n`, there is also 2π to consider...

martinholters:
Ah, of course, for even `n`, we can also always map to Ω [-π, π]; if that requires shifting by an odd multiple of 2π, we just have to negate the result.

src/diric.jl
```diff
@@ -0,0 +1,55 @@
...
+            ω = mod2pi(Ω + π) - π # (-π,π)
+        else
+            π2 = 2T(π)
+            ω = mod(Ω + π2, 4T(π)) - π2 # (-2π,2π)
+            if abs(ω) ≈ π2
+                # 2nd-order Taylor expansion near ±2π
+                return (abs(ω)-π2)^2 * (n^2 - 1) / 24 - one(T)
+            end
+        end
+        return one(T) - ω^2 * (n^2 - 1) / 24 # 2nd-order Taylor near 0
```
martinholters, line 48: If `ω^2 * (n^2 - 1) / 24 < 0.25*eps(T)` (i.e. `ω < sqrt(6eps(T)/(n^2-1))`), this will evaluate to `one(T)`. So if we fix `atol` at a rather low value as proposed above, it seems we could actually simplify this to just `one(T)` without loosing anything except for really large `n`.

src/diric.jl
JeffFessler: Oops, I was following some examples in the base julia docs without fully understanding them and now I realize that probably such refs only work within one package, not between packages.  I will remove them.

src/diric.jl
JeffFessler: Great points!  I agree that any decisions about tolerances should be in the design of the function rather than leaving it to the caller.

src/diric.jl
JeffFessler: Very insightful - in my enthusiasm about the Taylor expansion I overlooked that the square will make the 2nd-order term negligible in the precision in hand.  I agree we can simplify to `one`.

src/diric.jl
JeffFessler: Doh!  Good catch.

JeffFessler: BTW I started making other edits before noticing this suggestion.  I did incorporate the suggestion but I'm not enough of a git wizard to know how to "commit" your version.  Sorry about that.

JeffFessler:
@martinholters hopefully I've incorporated all your great suggestions.  I will be traveling the next 2 weeks so I might be slow to reply if you have further refinements.  Thanks!

src/diric.jl
martinholters: No worries.

martinholters:
We're definitely making progress here! I still see problematic numerical errors around 2π for larger even n (e.g. `n=400`). Guess we should either do the "map to [-π, π] and maybe flip the output sign" thing or expand the range where we do the polynomial approximation in that case, where I think I'd prefer the former.

BTW, thanks for your patience here.

src/diric.jl
```diff
@@ -0,0 +1,63 @@
...
+ 1.0  -0.2  0.2  -0.2  1.0  -0.2  0.2  -0.2  1.0
+
+julia> diric(0, 4)
+1.0
+```
+"""
+function diric(Ω::T, n::Int) where T <: AbstractFloat
+    n > 0 || throw(ArgumentError("n=$n is non-positive"))
+    if isodd(n)
+        Ω = mod2pi(Ω + π) - π # [-π,π)
```
martinholters, line 44:
```suggestion
        Ω = rem2pi(Ω, RoundNearest) # [-π,π)
```

src/diric.jl
```diff
@@ -0,0 +1,63 @@
...
+    else
+        Ω = 2 * (mod2pi(Ω/2 + π) - π) # [-2π,2π)
```
martinholters, line 46:
```suggestion
        Ω = 2 * rem2pi(Ω/2, RoundNearest) # [-2π,2π)
```

JeffFessler:
@martinholters, I still had trouble getting good precision near 2\pi for even `n` even after modifying the wrapping in the way I think you were suggesting.  I ended up just doubling the precision for intermediate calculations for the case of even `n` and that seemed to solve it for the tests I was using.  Odd `n` is the most useful case, so the extra "work" of doubling the precision for the even `n` case seems reasonable to me in the interest of providing an accurate result.  Do you have any ideas for a simpler way?

src/diric.jl
```diff
@@ -0,0 +1,78 @@
...
+```
+"""
+function diric(Ω::T, n::Int) where T <: AbstractFloat
+    n > 0 || throw(ArgumentError("n=$n is non-positive"))
+    sign = 1
+    if isodd(n)
+        Ω = rem2pi(Ω, RoundNearest) # [-π,π)
+    else
+        Ω = BigFloat(Ω ; precision = 2*precision(T))
+        Ω = 2 * rem2pi(abs(Ω)/2, RoundNearest) # [0,2π)
```
martinholters, line 48:
```suggestion
        Ω = 2 * rem2pi(abs(Ω)/2, RoundNearest) # [-2π,2π)
```

src/diric.jl
```diff
@@ -0,0 +1,78 @@
...
+        if Ω > π
+            sign = -1
+            Ω -= 2π # (-π,π]
```
martinholters, line 51:
```suggestion
            Ω -= 2π # (-π,0]
```

src/diric.jl
```diff
@@ -0,0 +1,78 @@
...
+        end
```
martinholters, line 52:
```suggestion
        else if Ω < π
            sign = -1
            Ω += 2π # (0,π]
        end
```

martinholters (commented):
This was almost what I had in mind, but missed half of it :smile:

src/diric.jl
JeffFessler: Note the `abs` above.  I did this so that I would not need the `else if` suggested below.
And I am pretty sure I tested accuracy near -2\pi to verify.

JeffFessler:
So the Julia 1.0 tests are failing because `BigFloat` in days of old does not support the `precision` kw arg.  Ugh.
That was added in 1.1
https://docs.julialang.org/en/v1/base/numbers/#BigFloats-and-BigInts
The latest commit intends to address it - I didn't not test it locally so someone please run CI...

src/diric.jl
martinholters: But e.g.
```julia
julia> 2*rem2pi(4.0, RoundNearest)
-4.566370614359173
```
even though `4.0 > 0`.

martinholters:
> So the Julia 1.0 tests are failing because BigFloat in days of old does not support the precision kw arg. Ugh.

I'd still hope we can get away without this conversion once the "avoid 2π" thing is done correctly.

JeffFessler:
Yep, after fixing up 2\pi cases properly we no longer need BigFloat.
Hopefully it is ready now.  Thanks for your patience!

test/diric.jl
```diff
@@ -0,0 +1,19 @@
...
+
+@testset "diric" begin
+    @test_throws ArgumentError diric(0, -2)
+    @test @inferred diric(0, 4) ≈ 1
+    @test @inferred diric(0 // 1, 5) == 1
+    @test @inferred diric(4π, 4) ≈ 1
+    @test @inferred diric(2π, 4) ≈ -1
+    @test @inferred diric(0, 5) ≈ 1
+    @test @inferred diric(2π, 5) ≈ 1
+    @test @inferred diric(π, 5) == 1 // 5
```
martinholters, line 11:
```suggestion
    @test @inferred(diric(0, 4)) ≈ 1
    @test @inferred(diric(0 // 1, 5)) == 1
    @test @inferred(diric(4π, 4)) ≈ 1
    @test @inferred(diric(2π, 4)) ≈ -1
    @test @inferred(diric(0, 5)) ≈ 1
    @test @inferred(diric(2π, 5)) ≈ 1
    @test @inferred(diric(π, 5)) == 1 // 5
```
Otherwise, this verifies that the comparison can be inferred.

src/diric.jl
```diff
@@ -0,0 +1,80 @@
...
+    return T(sign * sin(Ω * n/2) / (n * denom)) # typical case
+end
+
+# handle non AbstractFloat types (e.g., Int, Rational)
+diric(Ω::Real, n::Int) = diric(Float64(Ω), n::Int)
+
+# handle π exactly
+function diric(::Irrational{:π}, n::Int)
+    n > 0 || throw(ArgumentError("n=$n is non-positive"))
+    return 1 // n
```
martinholters, line 79: This is not generally correct? For even `n`, the result should be zero, for `mod(n,4)==3`, it should be `-1 // n`.

test/diric.jl
martinholters: Also, can we test a few arbitrary values which are not multiples of 2π?

martinholters (commented):
Alright, I think we're getting somewhere.

src/diric.jl
```diff
@@ -0,0 +1,80 @@
...
+=#
+    #   return one(T) - abs2(Ω) * (n*n - 1) / 24 # 2nd-order Taylor near 0
+        return sign * one(T)
+    end
+
+    return T(sign * sin(Ω * n/2) / (n * denom)) # typical case
+end
+
+# handle non AbstractFloat types (e.g., Int, Rational)
+diric(Ω::Real, n::Int) = diric(Float64(Ω), n::Int)
```
martinholters, line 74:
```suggestion
diric(Ω::Real, n::Int) = diric(float(Ω), n::Int)
```
Seems to be more typical, though I don't know what this entails exactly.

JeffFessler: Makes sense - probably gives Float32 on a 32-bit cpu...

src/diric.jl
JeffFessler: Doh!  I am just going to change it to let `float` take care of it.

src/diric.jl
```diff
@@ -0,0 +1,80 @@
...
+
+When `Ω` is an `AbstractFloat` (e.g., `Float32` or `BigFloat`),
+the return type matches that of `Ω`.  Otherwise the return type is `Float64`.
+
+As of 2021-03-19, the Wikipedia definition has different factors.
+The definition here is consistent with scipy and other software frameworks.
+
+This implementation treats inputs near multiples of 2π fairly carefully.
+If the denominator `sin(Ω/2)` is close to 0, then this function uses
+an accurate 2nd-order Taylor expansion rather than simple division.
```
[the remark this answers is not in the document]
JeffFessler: Oops, agreed.

src/diric.jl
```diff
@@ -0,0 +1,80 @@
...
+``1/n \\sum_{k=-(n-1)/2}^{(n-1)/2} e^{i k Ω}``,
+which is the discrete-time Fourier transform (DTFT)
+of a `n`-point moving average filter.
+
+When `n` is odd or even, the function is 2π or 4π periodic, respectively.
+The formula for general `n` is
+`diric(Ω,n) = ``e^{-i (n-1) Ω/2}/n \\sum_{k=0}^{n-1} e^{i k Ω}``.
+
+When `Ω` is an `AbstractFloat` (e.g., `Float32` or `BigFloat`),
+the return type matches that of `Ω`.  Otherwise the return type is `Float64`.
```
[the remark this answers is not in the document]
JeffFessler: Agreed.

JeffFessler:
Hopefully I covered all your suggestions.  In addition, I added a test for accuracy near 2\pi both to verify some more non multiple of pi values and so that if someone comes around later and tries to "simplify" the code to make it look like the matlab or scipy version they will see the accuracy issue.

martinholters (approved):
Generally looks good to me now. Some minor nits (I feel like I discover something new every time I look), but those are not mandatory in my eyes. We could also merge as is and only iterate more on this if the need actually arises.

src/diric.jl
```diff
@@ -0,0 +1,64 @@
...
+        end
+    end
+
+    denom = sin(Ω / 2)
+    atol = eps(T)
+    if abs(denom) ≤ atol # denom ≈ 0 ?
+        return sign
+    end
+
+    return sign * sin(Ω * n/2) / (n * denom) # typical case
```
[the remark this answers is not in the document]
JeffFessler: Yeah, this function is unlikely to be used in any tight inner loop so a bit of corner case type instability seems fine.

JeffFessler:
I went ahead and used `Integer`.  It's fine with me to fully refine it now while it's fresh on our minds...

src/diric.jl
```diff
@@ -0,0 +1,64 @@
...
+end
+
+# handle non AbstractFloat types (e.g., Int, Rational)
+diric(Ω::Real, n::Int) = diric(float(Ω), n::Int)
+
+# handle π
+diric(::Irrational{:π}, n::Int) = diric(float(π), n)
```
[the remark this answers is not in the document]
JeffFessler: Actually I don't think a specialized method is worth it now.  pi is special, but 2pi becomes a Float64 and I don't see any reason to give one integer multiple of pi special treatment.  `sin(pi)` does not return 0 in base...

src/diric.jl
martinholters: As I've commented a lot on this PR but not on this particular question it might be obvious, but just so it be said: I don't care.

martinholters (approved):
Good to merge from my side, assuming CI comes out green. I'd like to give it another 24 hours or so in case I spot something else with fresh eyes or someone else wants to chime in, but please bump if I forget to merge then.

src/diric.jl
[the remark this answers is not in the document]
martinholters: Yeah, there would be `sinpi` for that. Speaking of which... nah, let's save `diricpi` for another day :smile:

PR #413, opened 2021-03-22
replace 10^(x) with exp10(x) for db2pow and db2amp

N5N3:
replace 10^(x) with exp10(x) in db2pow and db2amp (about 4 times faster on my PC)

### DIFF
--- a/src/util.jl
+++ b/src/util.jl
@@ -135,7 +135,7 @@ const dBa = dBaconvert()
 Convert dB to a power ratio. This function call also be called
 using `a*dB`, i.e. `3dB == db2pow(3)`. The inverse of `pow2db`.
 """
-db2pow(a::Real) = 10^(a/10)
+db2pow(a::Real) = exp10(a/10)
 
 """
     db2amp(a)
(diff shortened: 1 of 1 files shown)


galenlynch:
Thanks for doing this!

rob-luke:
On my laptop I get...

```julia
julia> @benchmark 10^(a/10)  setup=(a=rand())
BenchmarkTools.Trial:
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     29.003 ns (0.00% GC)
  median time:      32.326 ns (0.00% GC)
  mean time:        35.036 ns (0.00% GC)
  maximum time:     247.130 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     993

julia> @benchmark exp10(a/10)  setup=(a=rand())
BenchmarkTools.Trial:
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     5.300 ns (0.00% GC)
  median time:      5.800 ns (0.00% GC)
  mean time:        6.347 ns (0.00% GC)
  maximum time:     64.900 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     1000

```

so looks good to me. If a second reviewer agrees they should hit merge.

rob-luke (approved):

PR #417, opened 2021-04-01
add license text to LICENSE.md

ericphanson:
The General registry requires an OSI-approved license (like MIT), which I think means the full text of the license. At least, that's what the new [automerge license check](https://discourse.julialang.org/t/fyi-the-general-registry-now-checks-for-an-osi-approved-software-license-when-registering-new-packages-and-new-versions/57018) will look for.

I copied the text from https://opensource.org/licenses/MIT; I wasn't sure who to put as the copyright holder so I just wrote "DSP.jl contributers".

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,6 +1,6 @@
 The DSP module is distributed under the MIT license.
 
-Copyright 2021 DSP.jl contributors
+Copyright 2012-2021 DSP.jl contributors
 
 Permission is hereby granted, free of charge, to any person obtaining a copy of this software and associated documentation files (the "Software"), to deal in the Software without restriction, including without limitation the rights to use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the Software, and to permit persons to whom the Software is furnished to do so, subject to the following conditions:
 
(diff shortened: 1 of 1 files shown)


LICENSE.md
```diff
@@ -1,2 +1,10 @@
 The DSP module is distributed under the MIT license.
 
+Copyright 2021 DSP.jl contributers
```
ericphanson, line 3:
```suggestion
Copyright 2021 DSP.jl contributors
```
whoops!

LICENSE.md
```diff
@@ -1,2 +1,9 @@
 The DSP module is distributed under the MIT license.
 
+Copyright 2021 DSP.jl contributors
```
ararslan, line 3: The license file was initially introduced in 2012. Should this state the current year or the year of initial copyright?

palday: initial copyright is what matters for when it passes into the public domain, although it's common to include a range

ericphanson, line 3:
```suggestion
Copyright 2012-2021 DSP.jl contributors
```
I'll put a range since that's what base Julia does

ericphanson:
thanks @ararslan!

Issue #341, opened 2020-01-22
analogfilter design seems to be broken


kumarbalachandran:
Please see https://discourse.julialang.org/t/dsp-jl-analog-filter-design/33646/5?u=kumarbalachandran for details.

kumarbalachandran:
Here is an attempt to reproduce the above post
```
# Test script to design an analog highpass elliptic filter with 8 poles and a cutoff 
# frequency of 4000 Hz. Various pressing questions arise:
#           1. What is the scaling of parameters for cutoff frequencies?
#           2. Why does an analog filter need a sampling frequency at all?
#           3. Why does the freqs function need a sampling frequency at all? I can
#              understand this for freqz, but 's'?

using DSP
using Plots

fs=4*5000e6 #Hz
# 4 GHz cutoff and a 20 GHz sampling freqency! Why 'fs" at all?
# Should there be a HighpassAnalog function?
responsetype = Highpass(4000e6; fs=fs)
# This at least is okay
designmethod = Elliptic(8, 1.0, 40.0)
# We want ANALOG, not digital
fil=analogfilter(responsetype, designmethod)

# Lets try to plot the response
rng=range(100e6, stop=4200e6, length=500)
resp=freqs(fil, rng, fs)
# The x-axis is in GHz
plot(collect(rng)/1e9, 20*log10.((abs.(resp))))
```
This gives the following plot
![image|690x461](upload://lKzAjFufs1XbUjVxWDEfknvrNAC.png) 

while if I scale the x-axis as 
```
# The x-axis is in GHz scaled by π
plot(collect(rng)/1e9 * π, 20*log10.((abs.(resp))))
```
I get (voila!)
![ellip2|690x465](upload://s8MeKwC1myjwcAWByLuaBjOZCUb.png) 

which does not make sense at all. However, there is clearly some factor of π missing in the DSP documentation as the cutoff is right where I want it to be in the plot. However I have no idea yet whether the filter is designed correctly.

I think the whole analogfilter design is broken. Please see the comments at the top of the code.

galenlynch:
This certainly seems like a bug, and I think you might be right that current interface to design analog filters is problematic.

kumarbalachandran:
One option is to create new ftypes AnalogHighPass, AnalogLowPass etc., that take only the cutoff frequency and no sampling frequency.

Alternatively, you can overload LowPass(Float64,String) as Lowpass(fc, "analog").  Numerical stability is a concern.

[...]

PR #418, opened 2021-04-11
Enable Analog Filter Design

Firionus:
This is only a quick and dirty fix (edit: for #341). The real solution is to move the `fs` argument from the response types to `digitalfilter`. 

This would enable something like:

```julia
responsetype = Lowpass(1e3) # 1kHz Low Pass
designmethod = Butterworth(2)
analog = analogfilter(responsetype, designmethod)
digital = digitalfilter(responsetype, designmethod, fs = 48e3)
```
The resulting filters `digital` and `analog` would then be quite similar. The validation that the lowpass cutoff is below the Nyquist frequency would stay in the `digitalfilter` function. 

With the current commit 083f, you'd need to create two different responsetypes for analog and digital:

```julia
analog_responsetype = Lowpass(1e3) # 1kHz Low Pass
designmethod = Butterworth(2)
analog = analogfilter(analog_responsetype, designmethod)

digital_responsetype = Lowpass(1e3, fs=48e3)
digital = digitalfilter(digital_responsetype, designmethod)
```

which kind of defeats the purpose of having a responsetype that can be shared between analog and digital filter design. 

Of course, if you only work on digital filters in normalized frequency, we can keep the current behavior by having a default of `fs=2`:
```julia
responsetype = Lowpass(.5)
designmethod = Butterworth(2)
digital = digitalfilter(responsetype, designmethod) # Implied fs=2, so it interprets the responsetype to have normalized frequency. 
```

What do you think of that proposed behavior? Should I implement it?


martinholters:
Superseded by #458.

PR #421, opened 2021-04-26
Bump Documenter version and Julia version for docs on CI

martinholters:
And update the doctests for changes in printed output.

Ref. https://github.com/JuliaDSP/DSP.jl/pull/392#discussion_r620218108 CC @rob-luke

### DIFF
--- a/.github/workflows/Documentation.yml
+++ b/.github/workflows/Documentation.yml
@@ -14,7 +14,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: julia-actions/setup-julia@v1
         with:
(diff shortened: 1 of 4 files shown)


rob-luke:
Thanks @martinholters

While on this topic, I find it helpful in other repos to be able to view the generated documentation for a PR. This can be done by enabling ‘ push_preview = true’ in docs/make.jl. However, it doesn’t seem to clean up when a PR is merged, so you end up with lots of old docs in the gh_pages branch. In my repos I manually delete these when I remember.

do you think this functionality would be helpful and do you know a better way to achieve it?

Issue #432, opened 2021-06-15
Elliptic Filter Design Broken in v"0.7.0"


mattcbro:
A recent update to DSP broke my code which used an Elliptic filter to design a bandpass filter.
Here is the code.

```
using DSP
using DSP.Filters

# create elliptic filter
bp1 = 0.75
bp2 = 10.0
Fsamp = 180
responsetype = Bandpass(bp1, bp2; fs = Fsamp)
designmethod = Elliptic(11, 0.25, 40)
#designmethod = Chebyshev2(11,  40)
bpass = digitalfilter(responsetype, designmethod)

# some random data
z = randn(300,2)
xnotch1 = filt(bpass, z)
```

This worked in v"0.6.10".  Furthermore it appears to run if I replace my filter type with Chebyshev2.

The Error I get is this:
```
ERROR: LoadError: BoundsError: attempt to access 1-element Vector{ComplexF64} at index [2]
Stacktrace:
  [1] getindex
    @ ./array.jl:801 [inlined]
  [2] splice!(a::Vector{ComplexF64}, i::Int64, ins::Vector{Any}) (repeats 2 times)
    @ Base ./array.jl:1481
  [3] groupzp(z::Vector{ComplexF64}, p::Vector{ComplexF64})
    @ DSP.Filters ~/.julia/packages/DSP/JvLS1/src/Filters/coefficients.jl:325
  [4] SecondOrderSections{:z, Float64, Float64}(f::ZeroPoleGain{:z, ComplexF64, ComplexF64, Float64})
    @ DSP.Filters ~/.julia/packages/DSP/JvLS1/src/Filters/coefficients.jl:389
  [5] SecondOrderSections
    @ ~/.julia/packages/DSP/JvLS1/src/Filters/coefficients.jl:427 [inlined]
  [6] SecondOrderSections
    @ ~/.julia/packages/DSP/JvLS1/src/Filters/coefficients.jl:269 [inlined]
  [7] convert
    @ ~/.julia/packages/DSP/JvLS1/src/Filters/coefficients.jl:5 [inlined]
  [8] filt(f::ZeroPoleGain{:z, ComplexF64, ComplexF64, Float64}, x::Matrix{Float64})
    @ DSP.Filters ~/.julia/packages/DSP/JvLS1/src/Filters/filt.jl:119
  [9] top-level scope
    @ /data/projects/Maestro/PulseLib/testsession.jl:20
 [10] include(fname::String)
[...]

PR #433, opened 2021-06-16
Avoid `BoundsError` in `groupzp`

martinholters:
The whole zero/pole grouping could use some love, I guess, but this is the easiest band-aid to fix #432.

I've also bumped the version for an immediate bugfix release.

PR #435, opened 2021-06-17
Make `SecondOrderSections(::ZeroPoleGain)`more robust

martinholters:
The helper function `groupzp` assumes that both zeros and poles are ordered such that complex entries appear in conjugate pairs consecutively and further, that these pairs are ordered consistently: Either always the positive imaginary part first, or always the negative imaginary part first. The output of `split_real_complex` fulfills this. However, for the poles, there is an intermediate sorting step which apparently can interfere with this ordering under certain circumstances. (And I'm not sure which circumstances, hence no additional tests.) This can lead to a `BoundsError` (#432), but also to zeros which are not conjugate pairs being grouped together, thereby silently producing wrong results.

While #433 was a band-aid preventing the `BoundsError`, this PR should fix the underlying issue by integrating the sorting into `split_real_complex` in such a way that the required ordering is ensured.

### DIFF
--- a/src/Filters/coefficients.jl
+++ b/src/Filters/coefficients.jl
@@ -322,9 +322,6 @@ function groupzp(z, p)
         groupedz[i] = splice!(z, closest_zero_idx)
         if !isreal(groupedz[i])
             i += 1
-            if closest_zero_idx > length(z)
-                closest_zero_idx = length(z)
-            end
             groupedz[i] = splice!(z, closest_zero_idx)
         end
         i += 1
@@ -337,7 +334,7 @@ end
 # Sort zeros or poles lexicographically (so that poles are adjacent to
 # their conjugates). Handle repeated values. Split real and complex
 # values into separate vectors. Ensure that each value has a conjugate.
-function split_real_complex(x::Vector{T}) where T
+function split_real_complex(x::Vector{T}; sortby=nothing) where T
     # Get counts and store in a Dict
     d = Dict{T,Int}()
     for v in x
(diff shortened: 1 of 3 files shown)


src/Filters/coefficients.jl
```diff
@@ -322,9 +322,6 @@ function groupzp(z, p)
         groupedz[i] = splice!(z, closest_zero_idx)
         if !isreal(groupedz[i])
             i += 1
-            if closest_zero_idx > length(z)
-                closest_zero_idx = length(z)
-            end
```
martinholters, line 327: This reverts the fix from #433, but the accompanying test still passes, corroborating that this new approach also fixes #432.

src/Filters/coefficients.jl
```diff
@@ -381,13 +382,9 @@ function SecondOrderSections{D,T,G}(f::ZeroPoleGain{D,Z,P}) where {D,T,G,Z,P}
     # Split real and complex poles
     (complexz, realz, matched) = split_real_complex(z)
     matched || throw(ArgumentError("complex zeros could not be matched to their conjugates"))
-    (complexp, realp, matched) = split_real_complex(p)
+    (complexp, realp, matched) = split_real_complex(p; sortby=x->abs(abs(x) - 1))
```
martinholters, line 385: For analog filters, `sortby=x->abs(real(x))` would probably make more sense, but that's an orthogonal issue.

martinholters:
Turns out the test case of #432 actually produces wrong results without this PR, so I modified the test to verify proper round-trippability. I've also included a version bump do craft the next bugfix release soonish...

PR #414, opened 2021-03-24
support for matrices in resample

baggepinnen:
The PR simply adds the support of providing a matrix of signals to `resample` and a keyword argument `dims` to specify which dimension to resample along.

PR #437, opened 2021-06-22
Make `resample` more generic

jonschumacher:
I found that `resample` is currently restricted to `AbstractMatrix`, which is an alias for `AbstractArray{T,2}`. Since `mapslices` works well with `dims > 2`, the changes from #414 can easily be extended to `AbstractArray` without restrictions on dimensionality, which I did in this PR.

### DIFF
--- a/src/Filters/stream_filt.jl
+++ b/src/Filters/stream_filt.jl
@@ -734,11 +734,11 @@ function resample(x::AbstractVector, rate::Real)
 end
 
 """
-    resample(x::AbstractMatrix, rate::Real, h::Vector = resample_filter(rate); dims)
+    resample(x::AbstractArray, rate::Real, h::Vector = resample_filter(rate); dims)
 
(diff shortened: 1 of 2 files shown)


martinholters:
Looks good on first glance, but I wonder why CI didn't run and I have no option of letting it. Can you maybe rebase and force-push?

jonschumacher:
Thanks for the feedback. I get the following message within this PR after merging the latest changes and thereby triggering the CI again, which should be similar to a rebase and force-push without the involved risks:
grafik

The link leads to https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks, which should give the instructions to running the CI.

rob-luke:
I've kicked the CI for you, feel free to ping if it gets stuck again.

martinholters (approved):
LGTM. I don't see why we shouldn't allow this.

rob-luke:
Thanks @jonschumacher

PR #438, opened 2021-06-22
Fix `DF2TFilter{<:PolynomialRatio}` with unequal numerator/denominator orders

### LINKED ISSUE #436: Filter coefficients error
Hi,
Is the below behaviour intentional?
Or have I misunderstood something....

<pre><font color="#4E9A06"><b>julia&gt; </b></font>cc=DSP.PolynomialRatio([1.0], [1.0,1.0])
PolynomialRatio{:z, Float64}(Polynomials.LaurentPolynomial(1.0), Polynomials.LaurentPolynomial(1.0*z⁻¹ + 1.0))

<font color="#4E9A06"><b>julia&gt; </b></font>ff=DSP.DF2TFilter(cc)
<font color="#EF2929"><b>ERROR: </b></font>ArgumentError: length of state vector must match filter order
Stacktrace:
 [1] <b>DF2TFilter</b>
<font color="#555753">   @ ~/.julia/packages/DSP/C3OhM/src/Filters/</font><font color="#555753"><u style="text-decoration-style:single">filt.jl:137</u></font><font color="#555753"> [inlined]</font>
 [2] <b>DF2TFilter</b><font color="#555753"> (repeats 2 times)</font>
<font color="#555753">   @ ~/.julia/packages/DSP/C3OhM/src/Filters/</font><font color="#555753"><u style="text-decoration-style:single">filt.jl:154</u></font><font color="#555753"> [inlined]</font>
 [3] top-level scope
<font color="#555753">   @ </font><font color="#555753"><u style="text-decoration-style:single">REPL[10]:1</u></font>
</pre>

This was with Julia 1.6.1 and DSP v0.7.2.
I briefly looked at the code and looked to be that state vector is initialized to the longer coefficient length, but then the error comes from a check that the input coefficients and state vector are all the same length. (lines 136-140 in filt.jl)
```
function DF2TFilter{Ti,Si}(coef::PolynomialRatio{:z}, state::Vector) where {Ti,Si}
        length(state) == length(coef.a)-1 == length(coef.b)-1 ||
            throw(ArgumentError("length of state vector must match filter order"))
        new{Ti,Si}(coef, state)
    end
```
Same error happens if the coefficients are padded with zeros to have same length, but not if coefficients are the same length and have non-zero first and last values.

martinholters:
Fixes #436.

### DIFF
--- a/src/Filters/filt.jl
+++ b/src/Filters/filt.jl
@@ -134,7 +134,7 @@ struct DF2TFilter{T<:FilterCoefficients{:z},S<:Array}
     state::S
 
     function DF2TFilter{Ti,Si}(coef::PolynomialRatio{:z}, state::Vector) where {Ti,Si}
-        length(state) == length(coef.a)-1 == length(coef.b)-1 ||
+        length(state) == max(length(coefa(coef)), length(coefb(coef)))-1 ||
             throw(ArgumentError("length of state vector must match filter order"))
         new{Ti,Si}(coef, state)
     end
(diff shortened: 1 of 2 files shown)


src/Filters/filt.jl
```diff
@@ -151,16 +151,22 @@ end
...
+    n = length(si) + 1
     b = coefb(f.coef)
+    if length(b) < n
+        append!(b, zeros(n-length(b)))
+    end
     a = coefa(f.coef)
-    n = length(b)
+    if length(a) < n
+        append!(a, zeros(n-length(a)))
+    end
```
martinholters, line 169: Note that `coefb` and `coefa` allocate already, so I aimed at the simple bugfix here, even if it may mean an additional allocation.

If we want to reduce the overhead here, we would need to access `coef.b` and `coef.a` directly, but that will likely make the actual filtering loop slower because indexing them is not as straight-forward (but on the plus-side, includes the zero-padding). Alternative, we could store the `b` and `a` in `DF2TFilter`, but that would require the type to be reorganized a bit.

martinholters:
Bump. I believe this a relatively straight-forward bugfix, but  a second pair of eyes would be appreciated.

martinholters:
Barring objections, I'll merge in a day or so.

rob-luke (approved):

PR #443, opened 2021-08-03
Update badges in README

martinholters:
Replacing the Travis one with the Github actions one should be uncontroversial, given that the last Travis run was for v0.7.0 in May.
Adding the docs badges seems in line with how most (?) packages do it and I like it, too. I also don't expect anyone to be against them; speak up in case I am mistaken.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 DSP.jl
 ======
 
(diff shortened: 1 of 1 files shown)


rob-luke:
Looks great! Thanks @martinholters

PR #448, opened 2021-11-12
Add `hann` export to be equal to `hanning`

staticfloat:
Since the original author's name was "Julius Von Hann", and referring to it as "hanning" is done only to draw similarities to the similar "hamming" window, let's export it under both common names.


src/windows.jl
```diff
@@ -159,7 +160,7 @@ The window is defined by sampling the continuous function:
 
 in the range `[-0.5, 0.5]`
 
-The `hanning` window satisfies the Constant Overlap-Add (COLA) property with an
+The `hanning` window satisfies the Constant Overlap-Add (COLA) property with a
```
rob-luke, line 163: Perhaps explain in the documentation why it is exported as hann too (similar to the explanation in the PR). 1) it's interesting piece of knowledge, and 2) so users can find it.

src/windows.jl
```diff
@@ -176,6 +177,7 @@ function hanning(n::Integer; padding::Integer=0, zerophase::Bool=false)
         0.5*(1+cos(2pi*x))
     end
 end
+hann = hanning
```
giordano, line 180: Shouldn't it be a const?
```suggestion
const hann = hanning
```

rob-luke:
Sorry for string reviewing. But could you please add a small test?

src/windows.jl
```diff
@@ -166,6 +166,11 @@ hop of 0.5, which means that adding together a sequence of delayed windows with
 a signal from a number of overlapping frames (each with a roughly rectangular
 window), to eliminate windowing amplitude modulation.
 
+While the this windowing function was originally named after its inventor
```
martinholters, line 169:
```suggestion
While this windowing function was originally named after its inventor
```

rob-luke (approved):

ViralBShah:
Good to merge?

Issue #429, opened 2021-05-18
Filter order


vitorn:
Is there a function to compute the required order for a Butterworth, Chebyshev or Kauer filter?  How about for min-max (Remez, Parks-McClellan) FIR filters?  I only found kaiserord, for Kaiser window FIR filters.

jsmithnh09:
The Butterworth filter approximation is something I'll give a shot at! I _think_ I have some suitable code for the lowpass and highpass cases, but I'll need to work out the bandpass and bandstop cases since those are a little trickier. I'm putting together test cases for these as well, based on [the examples on Mathworks' site.](https://www.mathworks.com/help/signal/ref/buttord.html)

jsmithnh09:
I believe the PR in #453 addresses the majority of the filter design cases here, I think the only outstanding one is parameter estimation for Parks-McClellan FIR filters.

jsmithnh09:
@vitorn We've added the filter order estimation schemes you requested, hopefully this closes out your issue!

martinholters:
Indeed, this seems fixed now. Comment if you disagree.

PR #451, opened 2021-12-06
Butterworth Filter Order Estimation (#429)

jsmithnh09:
This is a realization of #429 . I had utilized the Optim.jl package for bounded minimization when calculating the Bandstop filter case. I'm still very much a Julia newbie, so I apologize if the dependency on Optim isn't encoded in the Project.toml file, (still learning the developer ecosystem.) 

My open questions would be (1) is the Optim.jl dependency ok or would the reviewers suggest a more up-to-date package, and (2) should I investigate further on the minimization solution difference for the Bandstop filter case? The results are _slightly_ different than the solution provided by MATLAB, I had given some leniency in the buttord.jl test. I hope this is a worthy addition to DSP.jl!

### DIFF
--- a/src/Filters/buttord.jl
+++ b/src/Filters/buttord.jl
@@ -119,7 +119,7 @@ natfreq_estimate(warp::Real, Rs::Real, order::Integer) = warp / (db2pow(Rs) - 1)
 
 
 """
-    (N, ωn) = buttord(Wp, Ws, Rp, Rs)
+    (N, ωn) = buttord(Wp::Tuple{Real,Real}, Ws::Tuple{Real,Real}, Rp::Real, Rs::Real; domain=:z)
 
 Butterworth order estimate for bandpass and bandstop filter types. 
 `Wp` and `Ws` are 2-element pass and stopband frequency edges, with 
@@ -127,9 +127,13 @@ no more than `Rp` dB passband ripple and at least `Rs` dB stopband
 attenuation. Based on the ordering of passband and bandstop edges, 
 the Bandstop or Bandpass filter type is inferred. `N` is an integer 
 indicating the lowest estimated filter order, with `ωn` specifying
-the cutoff or "-3 dB" frequencies.
+the cutoff or "-3 dB" frequencies. If a domain of `:s` is specified,
+the passband and stopband frequencies are interpretted as radians/second,
+giving an order and natural frequencies for an analog filter. The default
+domain is `:z`, interpretting the input frequencies as normalized from 0 to 1,
+where 1 corresponds to π radians/sample.
 """
-function buttord(Wp::Tuple{Real,Real}, Ws::Tuple{Real,Real}, Rp::Real, Rs::Real)
+function buttord(Wp::Tuple{Real,Real}, Ws::Tuple{Real,Real}, Rp::Real, Rs::Real; domain::Symbol=:z)
 
     # make sure the band edges are in increasing order.
     Wps = (Wp[1] > Wp[2]) ? tuple(Wp[2], Wp[1]) : tuple(Wp[1], Wp[2])
@@ -138,9 +142,14 @@ function buttord(Wp::Tuple{Real,Real}, Ws::Tuple{Real,Real}, Rp::Real, Rs::Real)
     # infer filter type based on ordering of edges.
     ftype = (Wps[1] < Wss[1]) ? Bandstop : Bandpass
 
-    # pre-warp both components
-    wp = tan.(π/2 .* Wps)
-    ws = tan.(π/2 .* Wss)
+    # pre-warp both components, (if Z-domain specified)
+    if (domain == :z)
+        wp = tan.(π/2 .* Wps)
+        ws = tan.(π/2 .* Wss)
+    else
+        wp = Wps
+        ws = Wss
+    end
 
     if (ftype == Bandstop)
         # optimizer modifies passband frequencies.
@@ -155,12 +164,16 @@ function buttord(Wp::Tuple{Real,Real}, Ws::Tuple{Real,Real}, Rp::Real, Rs::Real)
     N = ceil(Int, order_estimate(Rp, Rs, wa))
 
     wscale = natfreq_estimate(wa, Rs, N)
-    ωn = (2/π).*atan.(fromprototype(wpadj, wscale, ftype))
+    if (domain == :z)
+        ωn = (2/π).*atan.(fromprototype(wpadj, wscale, ftype))
+    else
+        ωn = fromprototype(wpadj, wscale, ftype)
+    end
     N, ωn
 end
 
 """
-    (N, ωn) = buttord(Wp, Ws, Rp, Rs)
+    (N, ωn) = buttord(Wp::Real, Ws::Real, Rp::Real, Rs::Real; domain=:z)
 
 LPF/HPF Butterworth filter order and -3 dB frequency approximation. `Wp`
 and `Ws` are the passband and stopband frequencies, whereas Rp and Rs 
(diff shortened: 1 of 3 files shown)


jsmithnh09:
I think one other addition would be to include a keyword arg that accepts either `:z` or `:s` domain symbols that would indicate whether or not to apply frequency warping! I forgot to include this.

src/Filters/buttord.jl
```diff
@@ -0,0 +1,187 @@
...
+function fromprototype(Wp::AbstractArray{<:Real}, Wscale::Real, ftype::Type{Bandpass})
+    Wsc = [-Wscale, Wscale]
+    Wa = -Wsc .* (Wp[2]-Wp[1])./2 .+ sqrt.( Wsc.^2/4*(Wp[2]-Wp[1]).^2 .+ Wp[1]*Wp[2])
+    sort!(abs.(Wa))
+end
+
+order_estimate(Rp::Real, Rs::Real, warp::Real) = (log10(^(10, 0.1*Rs) - 1) - log10(^(10, 0.1*Rp) - 1)) / (2*log10(warp))
+natfreq_estimate(warp::Real, Rs::Real, order::Integer) = warp / (^(10, 0.1*Rs) - 1)^(1/(2*order))
+
+function buttord(Wp::AbstractArray{<:Real}, Ws::AbstractArray{<:Real}, Rp::Real, Rs::Real)
```
martinholters, line 122: I'd suggest using a tuple for the frequencies, as they have to hold exactly two elements. (Also in all the relevant helper functions.)
```suggestion
function buttord(Wp::Tuple{Real,Real}, Ws::Tuple{Real,Real}, Rp::Real, Rs::Real)
```
That might require some adjustments, like a replacements for `sort!(Wp)`, but should still be conceptually cleaner.

src/Filters/buttord.jl
```diff
@@ -0,0 +1,187 @@
...
+    """
+        buttord(Wp, Ws, Rp, Rs)
+
+    Butterworth order estimate for bandpass and bandstop filter types. 
+    `Wp` and `Ws` are 2-element pass/stopband frequency edges, with filttype specifying
+    the `Bandpass` or `Bandstop` filter types.
+    """
```
martinholters, line 129: Docstrings need to go before the function definition. Also, this should mention/explain the return value(s), which is not just the estimated order.

src/Filters/buttord.jl
```diff
@@ -0,0 +1,187 @@
...
+    ws = tan.(π/2 .* Ws)
+    if (ftype == Bandstop)
+        # optimizer modifies passband frequencies: this is intended behavior.
+        wa = toprototype!(wp, ws, Rp, Rs, ftype)
+    else
+        wa = toprototype(wp, ws, ftype)
+    end
+
+    # get the integer order estimate.
+    N = ceil(order_estimate(Rp, Rs, wa))
```
martinholters, line 152:
```suggestion
    N = ceil(Int, order_estimate(Rp, Rs, wa))
```
(likewise for the lowpass/highpass case below).

src/Filters/buttord.jl
```diff
@@ -0,0 +1,187 @@
...
+    sort!(abs.(Wa))
+end
+
+function fromprototype(Wp::AbstractArray{<:Real}, Wscale::Real, ftype::Type{Bandpass})
+    Wsc = [-Wscale, Wscale]
+    Wa = -Wsc .* (Wp[2]-Wp[1])./2 .+ sqrt.( Wsc.^2/4*(Wp[2]-Wp[1]).^2 .+ Wp[1]*Wp[2])
+    sort!(abs.(Wa))
+end
+
+order_estimate(Rp::Real, Rs::Real, warp::Real) = (log10(^(10, 0.1*Rs) - 1) - log10(^(10, 0.1*Rp) - 1)) / (2*log10(warp))
```
martinholters, line 119:
```suggestion
order_estimate(Rp::Real, Rs::Real, warp::Real) = (log10(db2pow(Rs) - 1) - log10(db2pow(Rp) - 1)) / (2*log10(warp))
```
The base 10 in the `log10`s used here is arbitrary, right? Not sure whether just using `log` has any pros or cons other than slightly shorter code.

martinholters (commented):
Not a thorough review, just some remarks from a quick glance.

I'm not too happy to add a dependency on Optim, but have no idea whether shipping a our own, taylored optimizer just for this task is any better.

src/Filters/buttord.jl
jsmithnh09: Functionally the `db2pow` gives the same result, so that's good! But I think I'm unsure of what you mean by using `log`, we aren't using natural log in the order estimate - I could be missing something in your description

src/Filters/buttord.jl
jsmithnh09: Moved and updated the docstrings!

src/Filters/buttord.jl
jsmithnh09: I believe I've made the necessary changes in [this commit.](https://github.com/JuliaDSP/DSP.jl/pull/451/commits/571df05a3fe54f1ae1d86de487b2f565444d8da5)

jsmithnh09:
I understand the hesitancy towards adding a new dependency - one thing to note though, the minimization will be needed if other filter type order estimation routines are added outside of just Butterworth, (Chebyshev types I and II, Elliptic, etc.) I'll investigate if a taylored optimizer can be used for these cases. What are your thoughts on the optional domain keyword suggestion, using `:z` or `:s`?

src/Filters/buttord.jl
```diff
@@ -0,0 +1,190 @@
...
+end
+
+function fromprototype(Wp::Tuple{Real,Real}, Wscale::Real, ftype::Type{Bandpass})
+    Wsc = [-Wscale, Wscale]
+    Wa = -Wsc .* (Wp[2]-Wp[1])./2 .+ sqrt.( Wsc.^2/4*(Wp[2]-Wp[1]).^2 .+ Wp[1]*Wp[2])
+    sort!(abs.(Wa))
+end
+
+order_estimate(Rp::Real, Rs::Real, warp::Real) = (log10(db2pow(Rs) - 1) - log10(db2pow(Rp) - 1)) / (2*log10(warp))
+natfreq_estimate(warp::Real, Rs::Real, order::Integer) = warp / (^(10, 0.1*Rs) - 1)^(1/(2*order))
```
martinholters, line 118:
```suggestion
natfreq_estimate(warp::Real, Rs::Real, order::Integer) = warp / (db2pow(Rs) - 1)^(1/(2*order))
```

src/Filters/buttord.jl
```diff
@@ -0,0 +1,190 @@
...
+    else
+        # no optimization in BPF case, use original warped passband edges.
+        wa = toprototype(wp, ws, ftype)
+        wpadj = wp
+    end
+
+    # get the integer order estimate.
+    N = ceil(Int, order_estimate(Rp, Rs, wa))
+
+    wscale = natfreq_estimate(wa, Rs, Integer(N))
```
martinholters, line 157: `N` is now an  `Int`, so
```suggestion
    wscale = natfreq_estimate(wa, Rs, N)
```

src/Filters/buttord.jl
```diff
@@ -0,0 +1,190 @@
...
+    # need to pre-warp since we want to use formulae for analog case.
+    wp = tan(π/2 * Wp)
+    ws = tan(π/2 * Ws)
+    wa = toprototype(wp, ws, ftype)
+
+    # rounding up fractional order. Using differences of logs instead of division. 
+    N = ceil(Int, order_estimate(Rp, Rs, wa))
+    
+    # specifications for the stopband ripple are met precisely.
+    wscale = natfreq_estimate(wa, Rs, Integer(N))
```
martinholters, line 185:
```suggestion
    wscale = natfreq_estimate(wa, Rs, N)
```

martinholters:
> I understand the hesitancy towards adding a new dependency - one thing to note though, the minimization will be needed if other filter type order estimation routines are added outside of just Butterworth, (Chebyshev types I and II, Elliptic, etc.) I'll investigate if a taylored optimizer can be used for these cases.

I'd also like to hear what others think about this.

> What are your thoughts on the optional domain keyword suggestion, using `:z` or `:s`?

That would just make the `tan`/`atan` frequency warping optional, right? Sounds like a good idea to include it then.

src/Filters/buttord.jl
martinholters: What I mean to say is: You could just replace `log10` with `log` in this line and the result would be the same up to rounding issues. But I don't know whether that is any better (or worse). It's just that the choice of base 10 seems arbitrary.

jsmithnh09:
I've added the domain keyword, current default domain is `:z`. I've added s-domain cases to the test script as well!

src/Filters/buttord.jl
jsmithnh09: You're right, the ratio is still the same - I had misinterpreted the base 10 comment as _inside_ the log10 function instead of the literal `log10` call. You're correct though, we could replace that with `log`, and the actual equation dictates using `log` as well in equation 10.3.48 of Proakis & Manolakis.

jsmithnh09:
I think I can probably continue adding the other filter type order estimation functions, if this is going to stay open for some time!

jsmithnh09:
> I think I can probably continue adding the other filter type order estimation functions, if this is going to stay open for some time!

On second thought, I'll wait for approval on this addition before adding the Elliptic/Chebyshev I and II order estimation routines.

jsmithnh09:
What are the thoughts on the changes? I've finished adding support for Elliptic/Chebyshev I and II order estimation routines, so it's no longer limited to just butterworth filter types.

PR #453, opened 2022-01-11
Butterworth, Elliptic, and Chebyshev I/II Filter Order Estimation

jsmithnh09:
This is a new PR for all filter types instead of solely Butterworth, (original PR was in #451.) I'm hoping this will address the issue highlighted by #429! Within the original PR, there was concern over adding a dependency on `Optim.jl`, which is needed for band-stop filter order estimation for all of the filter types. 

I'm interested to hear if others would recommend writing a tailored minimization function as Martin had pointed out or if the package dependency is ok, but I'm not sure how active this project is nowadays. Hopefully this is useful! I've written tests as well in `filt_order.jl` for checking solutions in both discrete and analog design cases against Scipy and the MATLAB use cases highlighted on their reference site.


rob-luke:
I am also concerned about adding a large dependency like optim (which itself has a dozen dependencies). Can you take a stab at writing a tailored optimisation routine so we can evaluate if its worth the tradeoff?

jsmithnh09:
@rob-luke I busted out the Numerical Recipes and made changes based on what's in the Optim.jl package for Brent's method. I think I've successfully removed the dependency on `Optim.jl`, but let me know if there ought to be any root finding tests for confirming `brent` works fine down the line!

jsmithnh09:
@rob-luke @martinholters I'm going back and revisiting previous PR's, thankfully the tests pass with the Brent method replacement; no more `Optim.jl` dependency. Is anything else required to continue with the merge? Thanks again for reviewing this!

src/Filters/filt_order.jl
```diff
@@ -0,0 +1,458 @@
...
+ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDERS OR CONTRIBUTORS
+BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY,
+OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF
+THE POSSIBILITY OF SUCH DAMAGE.
+
+--- end of scipy license
```
martinholters, line 40: Is this compatible with our MIT license? I guess so, but don't knoiw.

martinholters: Yes, if your code is based on the SciPy, I don't argue against including this license, but rather wonder if we may include the code here and publish it under the project's MIT license. (And yes, if that is a problem, then we have to worry about remez_fir, too.) There's probably a site somewhere that collects this license compatibility stuff, but from a quick search, I couldn't find anything applicable.

jsmithnh09: From what I've gathered, the only outstanding stipulation is the third clause of the BSD license, indicating the need for explicit written permission if the Julia code is derived from SciPy. The only portion that I had looked to SciPy for, (outside of the references listed in the `filter_order.jl` file), was how to optimize the passband edges in the bandpass/bandstop filter cases. Let me know if there's anything else I can do or provide @martinholters !

src/Filters/filt_order.jl
```diff
@@ -0,0 +1,458 @@
...
+"""
+function buttord(Wp::Tuple{Real,Real}, Ws::Tuple{Real,Real}, Rp::Real, Rs::Real; domain::Symbol=:z)
+
+    # make sure the band edges are in increasing order.
+    Wps = (Wp[1] > Wp[2]) ? tuple(Wp[2], Wp[1]) : tuple(Wp[1], Wp[2])
+    Wss = (Ws[1] > Ws[2]) ? tuple(Ws[2], Ws[1]) : tuple(Ws[1], Ws[2])
+
+    # infer filter type based on ordering of edges.
+    ftype = (Wps[1] < Wss[1]) ? Bandstop : Bandpass
+
```
martinholters, line 246: Should we verify that `(Wps[1] < Wss[1]) == (Wps[2] > Wss[2])` and throw an informative error otherwise?

src/Filters/filt_order.jl
```diff
@@ -0,0 +1,458 @@
...
+end
+
+function fromprototype(Wp::Tuple{Real,Real}, Wscale::Real, ftype::Type{Bandpass})
+    Wsc = [-Wscale, Wscale]
+    Wa = -Wsc .* (Wp[2]-Wp[1])./2 .+ sqrt.( Wsc.^2/4*(Wp[2]-Wp[1]).^2 .+ Wp[1]*Wp[2])
+    sort!(abs.(Wa))
+end
+
+butterworth_order_estimate(Rp::Real, Rs::Real, warp::Real) = (log(db2pow(Rs) - 1) - log(db2pow(Rp) - 1)) / (2*log(warp))
+natfreq_estimate(warp::Real, Rs::Real, order::Integer) = warp / (db2pow(Rs) - 1)^(1/(2*order))
```
martinholters, line 88: Is this only used by the `buttord` code? Then it could also use a `butterworth_` prefix.

src/Filters/filt_order.jl
```diff
@@ -0,0 +1,458 @@
...
+function chebyshev_order_estimate(Rp::Real, Rs::Real, Wa::Real)
+    # Chebyshev1/2 order estimate. Requires inverse hyperbolic cosine.
+    ϵs, ϵp = db2pow(Rs) - 1, db2pow(Rp) - 1    
+    acosh(√(ϵs / ϵp)) / acosh(Wa) # Eq. (10.3.63) in [1]
+end
+
+function brent(f, x1::T, x2::T) where T <: AbstractFloat
+    a, b = x1, x2
+    fa, fb = f(a), f(b)
+    c = copy(a) # midpoint of [b, a] bracket
```
martinholters, line 113: `c` seems unused before the next assignment below also is not accessed outside the `while` loop, so what's this line for? Can we just get rid of it?
```suggestion
```

src/Filters/filt_order.jl
```diff
@@ -0,0 +1,458 @@
...
+    g = 1/2 * (3 - √(5)) # golden ratio
+    k = zero(T)
+    k1 = zero(T)
+    
+    # first interval calculation.
+    m = a + g*(b - a)
+    m1, m2 = m, m
+    fm = f(m)
+    fm1, fm2 = fm, fm
+    iter = zero(T)
```
martinholters, line 124:
```suggestion
    iter = 0
```

src/Filters/filt_order.jl
```diff
@@ -0,0 +1,458 @@
...
+        and bandstop edges, the Bandstop or Bandpass filter type is inferred. `N` is an integer indicating the 
+        lowest estimated filter order, with `ωn` specifying the cutoff or "-3 dB" frequencies. If a domain of 
+        `:s` is specified, the passband and stopband frequencies are interpretted as radians/second, giving an 
+        order and natural frequencies for an analog filter. The default domain is `:z`, interpretting the input 
+        frequencies as normalized from 0 to 1, where 1 corresponds to π radians/sample.
+        """
+        function $fcn(Wp::Tuple{Real, Real}, Ws::Tuple{Real, Real}, Rp::Real, Rs::Real; domain::Symbol=:z)
+            Wps = (Wp[1] > Wp[2]) ? tuple(Wp[2], Wp[1]) : tuple(Wp[1], Wp[2])
+            Wss = (Ws[1] > Ws[2]) ? tuple(Ws[2], Ws[1]) : tuple(Ws[1], Ws[2])
+            ftype = (Wps[1] < Wss[1]) ? Bandstop : Bandpass
```
martinholters, line 377: Like above: Should we verify that `(Wps[1] < Wss[1]) == (Wps[2] > Wss[2])` and throw an informative error otherwise?

src/Filters/filt_order.jl
```diff
@@ -0,0 +1,458 @@
...
+and bandstop edges, the Bandstop or Bandpass filter type is inferred. `N` is an integer indicating the 
+lowest estimated filter order, with `ωn` specifying the cutoff or "-3 dB" frequencies. If a domain of 
+`:s` is specified, the passband and stopband frequencies are interpretted as radians/second, giving an 
+order and natural frequencies for an analog filter. The default domain is `:z`, interpretting the input 
+frequencies as normalized from 0 to 1, where 1 corresponds to π radians/sample.
+"""
+function cheb2ord(Wp::Tuple{Real, Real}, Ws::Tuple{Real, Real}, Rp::Real, Rs::Real; domain::Symbol=:z)
+    Wps = (Wp[1] > Wp[2]) ? tuple(Wp[2], Wp[1]) : tuple(Wp[1], Wp[2])
+    Wss = (Ws[1] > Ws[2]) ? tuple(Ws[2], Ws[1]) : tuple(Ws[1], Ws[2])
+    ftype = (Wps[1] < Wss[1]) ? Bandstop : Bandpass
```
martinholters, line 435: And here, too: Should we verify that `(Wps[1] < Wss[1]) == (Wps[2] > Wss[2])` and throw an informative error otherwise?

martinholters (commented):
I haven't gone through all the math, but mostly looks good to me, except for a few minor nitbits.

src/Filters/filt_order.jl
jsmithnh09:
```suggestion
    (Wps[1] < Wss[1]) != (Wps[2] > Wss[2]) && error("Pass and stopband edges must be ordered for Bandpass/Bandstop cases.")

    # infer filter type based on ordering of edges.
    ftype = (Wps[1] < Wss[1]) ? Bandstop : Bandpass
```
Would this be a suitable enough error message?

jsmithnh09:
```julia
(Wps[1] < Wss[1]) != (Wps[2] > Wss[2]) && error("Pass and stopband edges must be ordered for Bandpass/Bandstop filters.")
```
Would this be a contextual enough error message?

src/Filters/filt_order.jl
jsmithnh09: Adding!

src/Filters/filt_order.jl
martinholters: I don't worry too much about the exact wording. But should probably be an `ArgumentError`.

jsmithnh09: Added! I also included some test cases for `brent` so that we're covered.

src/Filters/filt_order.jl
jsmithnh09: Added!

martinholters (approved):
LGTM now. I'll leave this open for another one or two days in case anyone else wants to chime in, otherwise I'll go ahead merge.

martinholters:
Thanks for not giving up on this! It took quite a while, mostly due to lack of feed...

jsmithnh09:
No worries at all! Thank you for the reviews/feedback/effort in maintaining the project. Happy to see this is a good addition!

PR #457, opened 2022-01-18
Include Julia 1.6 in versions to run CI against

martinholters:


martinholters:
The total number of test runs here might be a bit excessive. We could consider only running all OSes/architectures with the oldest supported Julia version (1.0 at present), Julia LTS (1.6), Julia latest (1.8), and Julia nightly, and running the remaining Julia version only on a single OS/architecture (e.g. Ubuntu x64). Opinions?

martinholters:
Failures on nightly should be fixed by #456; tests pass on newly added Julia 1.6,

rob-luke:
> consider only running all OSes/architectures with the oldest supported Julia version (1.0 at present), Julia LTS (1.6), Julia latest (1.8), and Julia nightly, and running the remaining Julia version only on a single OS/architecture (e.g. Ubuntu x64). Opinions?

I am a -1 vote on this.
Given this is a core package for many of us I don't see a problem with running many tests. They run quite quickly and in parallel, plus the package doesn't have excessive activity. I'd only vote to decrease test/version coverage if we are hitting CI quota limits. While in theory, the oldest and latest versions should be sufficient, bugs can creep in and out.

martinholters:
Alright, then let's go with this as is.

PR #458, opened 2022-01-25
Enable Analog Filter Design

gruberchr:
This is a proposal for fixing analog filter design (see bug #341). For this frequency normalization has been shifted to function `digitalfilter()`. And function `analogfilter()` doesn't require any sampling frequency now. All affected tests have been adapted by simply providing sampling frequency to function `digitalfilter()` instead of `FilterType` construction.

Documentation has not been changed yet, I'll do this next. Additionally some new tests should be appended for analog filter design, I guess.

Any remarks?


gruberchr:
> Are we otherwise happy with the filter design API or should we take the opportunity and do more redesign work?

I was also thinking about introducing domain specifiers `:s` and `:z` for the `FilterType` as well as it is already the case for the `FilterCoefficients` type. This way it could be distinguished between "unnormalized" (`:s` domain) and "normalized" (`:z` domain) `FilterType`s. This would extend the flexibility for the user. One could decide to do frequency normalization during filter type specification already, as it is currently implemented, or doing frequency normalization during filter design, as proposed in this PR. Analog filter design would still be possible by using an unnormalized `FilterType`.

gruberchr:
I had a look on the test cases in *test/filter_design.jl* now. I think the test cases should be extended as follows:

1. Test cases for function `analogfilter()` with non-normalized frequency values in the filter response definition.
1. Test cases for function `digitalfilter()` with non-normalized frequency values in the filter response definition and a sampling frequency specification.

I can find the following test cases for analog filter design:
- `freq. scaling` (By the way, is this a typo? Should be `lowpass` I guess)
- `highpass`
- `bandpass`
- `bandstop`

Should these test cases be changed to use arbitrary (non-normalized) frequency values or should we add new test cases?

All test cases are based on a MATLAB reference implementation. I don't have MATLAB, but could use Octave instead. Accuracy seems to be sufficient as well. Is Octave a problem?

For digital filter design I can find the following test case:
- `digital IIR`

We could use the existing reference implementation and simply add some code lines to design the same digital filter using non-normalized frequency values. I've provided a proposal (0d05b11). Or should we add new test cases?

What do you think?

gruberchr:
Off topic: On my PC the test case `TF to ZPK` fails. There seems to be an accuracy problem. If I set `relerr=4` in the function call `zpkfilter_accuracy()` the test runs successfully.

Does anybody know, what could be the problem? I don't really understand function `zpkfilter_accuracy()` at the moment.

gruberchr:
@martinholters Did you have the time to take a look at the code and the tests?

martinholters:
Makes sense to me and from an admittedly not-yet-too-close look at the code and tests, everything looks good, but we should probably add tests for what this is solving: analog filters with frequencies outside the "normalized" range.

gruberchr:
I updated analog filter design tests now using non-normalized frequency values in the filter response definition.

Unfortunately Octave's elliptic filter design function `ellip()` generates different filter coefficients than Julia when designing a 20th order analog filter. Maybe this is a bug in Octave's `signal` package. When I plot a frequency response from Octave's filter coefficients, the stop band attenuation does not seem to be as expected (20 dB). Therefore I changed the test to design a Chebyshev type II filter instead of an elliptic filter.

Additionally I fixed the organization and naming of the analog filter design test set.

gruberchr:
@martinholters Any progress on this MR?

ViralBShah:
I know this was reviewed a long time ago - but wondering if it is good to pull in and merge?

martinholters (approved):
Yes, let's go ahead with this.
As it is a breaking change, we might want to bump the version number to `0.8.0-dev` as part of this PR so that we don't forget. Opinions? @ViralBShah?

ViralBShah:
Yes, makes sense to go to 0.8-dev.

martinholters:
Now that #532 bumped the version to 0.8, let's try this.

martinholters:
Thanks @gruberchr and sorry for the long delay.

PR #468, opened 2022-03-30
Add complex valued bandpass filter

zsoerenm:
This pull request allows to create ComplexBandpass filters for one sided pass bands (different pass bands for positive or negative frequencies)
I haven't implemented `scalefactor(coefs::Vector, ftype::ComplexBandpass)` yet, as I'm not really sure how to implement it for this case. So this currently only works for `scale = false`.
Also I haven't implemented tests, yet. Where do all the filter coefficients in the tests for the different filters come from? Maybe someone can help me out with this.
Here is an example:
```
using DSP, FFTW, Plots
signal = randn(ComplexF64, 10000)
filtered_signal = filt(digitalfilter(ComplexBandpass(-5e6, -3e6; fs = 10e6), FIRWindow(hanning(512), scale = false)), signal)
p = periodogram(filtered_signal, fs = 10e6)
Plots.plot(fftshift(freq(p)), 10 * log10.(fftshift(power(p))))
```
ComplexBandpass


src/Filters/design.jl
```diff
@@ -241,6 +241,11 @@ function normalize_freq(w::Real, fs::Real)
     f >= 1 && error("frequencies must be less than the Nyquist frequency $(fs/2)")
     f
 end
+function normalize_complex_freq(w::Real, fs::Real)
+    f = 2*w/fs
+    f >= 2 && error("frequencies must be less than the Nyquist frequency $(fs)")
```
martinholters, line 246: Should we also put a lower bound on `f`? It seems that `0 <= f < 2` and `-1 <= f < 1` would be reasonable limits, but depending on context, one or the other may be more useful. So require `-1 <= f < 2`? Looks rather arbitrary. Or maybe we put no restriction on `f` at all? Really, I don't know.

martinholters:
Regarding tests, I think most of the reference data has been produced using comparable Matlab functions. Personally, I'm not too huge a fan of this. For the case at hand, I could imagine determining the frequency responses of a couple of designed filters and verifying that they fulfill the specifications (under some assumptions on what passband ripple and stopband gain can be reasonably expected).

src/Filters/design.jl
```diff
@@ -234,61 +234,84 @@ Elliptic(n::Integer, rp::Real, rs::Real) = Elliptic(Float64, n, rp, rs)
...
     Lowpass(Wn)
 
 Low pass filter with cutoff frequency `Wn`.
 """
-Lowpass(w::Real) = Lowpass{typeof(w/1)}(w)
-
-struct Highpass{T} <: FilterType
+struct Lowpass{T<:Real} <: FilterType
     w::T
+    Lowpass{T}(w::Real) where T <: Real = new{typeof(one(T) / 1)}(w)
```
martinholters, line 254: This means that `Lowpass{Int}(1)` will return a `Lowpass{Float64}`, right? While not forbidden, I don't think that's a good idea. I'd really expect the `T(x) isa T` for any type `T` (although there are exceptions to this, I know).

martinholters (changes requested):
I think constructor should return values of the type they're asked to construct, which appears not to be the case here.

src/Filters/design.jl
```diff
@@ -234,61 +234,84 @@ Elliptic(n::Integer, rp::Real, rs::Real) = Elliptic(Float64, n, rp, rs)
...
 Low pass filter with cutoff frequency `Wn`.
 """
-Lowpass(w::Real) = Lowpass{typeof(w/1)}(w)
-
-struct Highpass{T} <: FilterType
+struct Lowpass{T<:Real} <: FilterType
     w::T
+    Lowpass{T}(w::Real) where {T<:Real} = new{T}(w)
+    Lowpass(w::Real) = Lowpass{typeof(w / 1)}(w)
 end
```
martinholters, line 256: The diff is a bit messy, so maybe I'm missing something, but why can't this just be
```suggestion
struct Lowpass{T<:Real} <: FilterType
    w::T
end
Lowpass(w::Real) = Lowpass{typeof(w / 1)}(w)
```
(which, if I read correctly, doesn't change much, just adds the `<: Real` restriction on `T`)

martinholters: Ah, right, without any explicit inner constructors, a `Lowpass(w::Real)` method is auto-generated. (Before, a `Lowpass(w::Any)` was auto-generated which didn't conflict with the explicit `Lowpass(w::Real)`  being defined.) Somewhat unfortunate code-churn, but seems unavoidable then.

But while at it, what do think about:
```suggestion
struct Lowpass{T<:Real} <: FilterType
    w::T
    Lowpass{T}(w::Real) where {T<:Real} = new{T}(w)
    Lowpass(w::T) where {T<:Real} = Lowpass{float(T)}(w)
end
```
Might a bit more efficient for non-trivial types like `BigInt` or `BigFloat`, but I don't know whether it might make a semantic difference.

martinholters: Admitted, I haven't looked where we use instances of these types. What do we expect of the type of `w`?

martinholters: Or, to put it differently, what goes wrong with just:
```suggestion
struct Lowpass{T<:Real} <: FilterType
    w::T
end
```

martinholters: Then let's just leave this as `typeof(w / 1)` for now; mostly unrelated to the intention of the PR anyway.

src/Filters/design.jl
```diff
@@ -472,8 +495,10 @@ function bilinear(f::ZeroPoleGain{:s,Z,P,K}, fs::Real) where {Z,P,K}
...
 
 # Pre-warp filter frequencies for digital filtering
-prewarp(ftype::Union{Lowpass, Highpass}, fs::Real) = (typeof(ftype))(prewarp(normalize_freq(ftype.w, fs)))
-prewarp(ftype::Union{Bandpass, Bandstop}, fs::Real) = (typeof(ftype))(prewarp(normalize_freq(ftype.w1, fs)), prewarp(normalize_freq(ftype.w2, fs)))
+prewarp(ftype::Lowpass, fs::Real) = Lowpass(prewarp(normalize_freq(ftype.w, fs)))
+prewarp(ftype::Highpass, fs::Real) = Highpass(prewarp(normalize_freq(ftype.w, fs)))
```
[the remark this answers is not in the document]
martinholters: LGTM as is.

martinholters:
LGTM modulo the minor docstring fix. Good to go then or are there any objections?

martinholters:
Ah, good point. I think this should work:
```julia
function scalefactor(coefs::Vector{T}, ftype::ComplexBandpass, fs::Real) where T<:Number
    n = length(coefs)
    freq = normalize_complex_freq(middle(ftype.w1, ftype.w2), fs)
    c = zero(T)
    for k = 1:n
        c = muladd(coefs[k], cispi(-freq * (k - (n + 1) / 2)), c)
    end
    return abs(c)
end
```

src/Filters/design.jl
```diff
@@ -569,19 +594,31 @@ FIRWindow(; transitionwidth::Real=throw(ArgumentError("must specify transitionwi
...
     [w2*sinc(w2*(k-(n+1)/2)) - w1*sinc(w1*(k-(n+1)/2)) for k = 1:n]
 end
 
+function firprototype(n::Integer, ftype::ComplexBandpass, fs::T) where {T<:Real}
+    w1 = normalize_complex_freq(ftype.w1, fs)
+    w2 = normalize_complex_freq(ftype.w2, fs)
+    w_center = (w2 + w1) / 2
+    w_cutoff = (w2 - w1) / 2
+    lp = Lowpass(w_cutoff)
+    _firprototype(n, lp, fs, Complex{T}) .*= cispi.(w_center * (0:(n-1)))
```
martinholters, line 619: As `w_cutoff` is a normalized frequency, we have to use `fs=2` for it:
```suggestion
    _firprototype(n, lp, 2, Complex{T}) .*= cispi.(w_center * (0:(n-1)))
```
But I suspect this will make the test fail.

test/filter_design.jl
```diff
@@ -1012,6 +1012,13 @@ end
     winfirtaps_scipy = readdlm(joinpath(dirname(@__FILE__), "data", "digitalfilter_hamming_129_bandpass_fc0.1_0.2_fs1.0.txt"),'\t')
     @test winfirtaps_jl ≈ winfirtaps_scipy
 
+    @test_throws ArgumentError ComplexBandpass(2, 1)
+
+    # TODO: write actual tests for ComplexBandpass
+    winfirtaps_jl = digitalfilter(ComplexBandpass(0.1, 0.2), FIRWindow(hamming(128), scale=false); fs=1)
+    winfirtaps_scipy = readdlm(joinpath(dirname(@__FILE__), "data", "digitalfilter_hamming_128_complexbandpass_fc1.txt"), '\t', ComplexF64)
+    @test winfirtaps_jl ≈ winfirtaps_scipy  # replace with real test data
```
martinholters, line 1020: Can we add a comment how the reference data was generated here? Also, I think we should check with another `fs`, just to make sure the scaling works as expected, e.g.
```julia
digitalfilter(ComplexBandpass(100, 200), FIRWindow(hamming(128), scale=false); fs=1000)
```
should give the same result.

martinholters (changes requested):
Trying to re-create the example from the OP, I got different results which lead me to what I suspect to be a bug in how `fs` is used.

test/filter_design.jl
martinholters: Ah, I wasn't aware of this circulus vitiosus. (Even noting the comment, the `_scipy` suffix in the variable name lead me to wrong conclusions.)

I'll try to cook up a more effective test.

martinholters (approved):
LGTM.

Issue #292, opened 2019-04-28
Direct colvolution


galenlynch:
It would be great if we had a direct convolution kernel, which would probably be faster for small convolutions.

ssfrr:
We have a [time-domain implementation](https://github.com/JuliaDSP/DSP.jl/blob/0339fe2035be3ab718e35e101db798005d89d8a3/src/Filters/filt.jl#L431) as part of `filt`, as well as a heuristic for when to do time-domain or FFT. Seems like we should unify `filt` and `conv`, and then combine the heuristics to pick from the 3 methods (time-domain, full FFT, and overlap-add FFT).

galenlynch:
Yeah I 100% agree, my main motivation for adding overlp-save to conv was to pave the way to combine `filt` and `conv`, at least for FIR filtering. However, the devils is in the details: extrapolating the 1D direct convolution in `filt` to ND with abstract arrays seems like it would be just as fiddly as overlap save was.

martinholters:
I'd really love to have a `conv` fallback that is a generic as possible, being usable with any type combination that supports `*` and `+`, i.e. letting `collect` figure out the result element type. I had made a start somewhere, but hadn't really pursued it. Maybe I can give it another shot. But we might still want an implementation for a limited set of types that is potentially faster.

martinholters:
So here is a potential fallback implementation:
```julia
julia> function directconv(A::AbstractArray{<:Any,M}, B::AbstractArray{<:Any,N}) where {M,N}
           axes_A = ntuple(i -> axes(A, i), Val(max(M,N)))
           axes_B = ntuple(i -> axes(B, i), Val(max(M,N)))
           krange(i) = CartesianIndices((:).(max.(first.(axes_B), Tuple(i) .- last.(axes_A)), min.(last.(axes_B), Tuple(i) .- first.(axes_A))))
           return [sum(A[i-k]*B[k] for k in krange(i)) for i in CartesianIndices((:).(first.(axes_A).+first.(axes_B), last.(axes_A).+last.(axes_B)))]
       end
directconv (generic function with 1 method)

julia> A = rand(3,4); B = rand(5,6,7);

julia> @btime conv($A,$B);
  112.307 μs (178 allocations: 72.63 KiB)

julia> @btime directconv($A,$B); 
  20.673 μs (885 allocations: 52.00 KiB)
```
[...]

PR #477, opened 2022-07-20
exchange BLAS.BlasFloat for Union{AbstractFloat, Complex{<:AbstractFl…

MikaelSlevinsky:
…oat}}

There is an upcoming package for FFTs on generic number types https://github.com/JuliaApproximation/GenericFFT.jl. This will enable, in particular, convolutions with `BigFloat` and `Complex{BigFloat}` arrays. To make this Just Work<sup>TM</sup>, we may simply remove the type restrictions referring to FFTW number types (quoted as `BLAS.BlasFloat`s). Then when both packages are loaded, `using DSP, GenericFFT`, `conv(rand(BigFloat, 10), rand(BigFloat, 20))` works analogously to finding BigFloat eigenvalues through `GenericLinearAlgebra`.

### DIFF
--- a/src/dspbase.jl
+++ b/src/dspbase.jl
@@ -677,7 +677,7 @@ function _conv(u, v, su, sv)
 end
 
 # We use this type definition for clarity
-const AbstractFloats = Union{<:AbstractFloat,Complex{T} where T<:AbstractFloat}
+const RealOrComplexFloat = Union{AbstractFloat, Complex{T} where T<:AbstractFloat}
 
 # May switch argument order
 """
@@ -688,7 +688,7 @@ depending on the size of the input. `u` and `v` can be  N-dimensional arrays,
 with arbitrary indexing offsets, but their axes must be a `UnitRange`.
 """
 function conv(u::AbstractArray{T, N},
-              v::AbstractArray{T, N}) where {T<:AbstractFloats, N}
+              v::AbstractArray{T, N}) where {T<:RealOrComplexFloat, N}
     su = size(u)
     sv = size(v)
     if prod(su) >= prod(sv)
@@ -698,8 +698,8 @@ function conv(u::AbstractArray{T, N},
     end
 end
 
-function conv(u::AbstractArray{<:AbstractFloats, N},
-              v::AbstractArray{<:AbstractFloats, N}) where N
+function conv(u::AbstractArray{<:RealOrComplexFloat, N},
+              v::AbstractArray{<:RealOrComplexFloat, N}) where N
     fu, fv = promote(u, v)
     conv(fu, fv)
 end
@@ -711,11 +711,11 @@ conv(u::AbstractArray{<:Number, N}, v::AbstractArray{<:Number, N}) where {N} =
     conv(float(u), float(v))
 
 function conv(u::AbstractArray{<:Number, N},
-              v::AbstractArray{<:AbstractFloats, N}) where N
+              v::AbstractArray{<:RealOrComplexFloat, N}) where N
     conv(float(u), v)
 end
 
-function conv(u::AbstractArray{<:AbstractFloats, N},
+function conv(u::AbstractArray{<:RealOrComplexFloat, N},
               v::AbstractArray{<:Number, N}) where N
     conv(u, float(v))
 end
--- a/src/dspbase.jl
+++ b/src/dspbase.jl
@@ -677,7 +677,7 @@ function _conv(u, v, su, sv)
 end
 
 # We use this type definition for clarity
-const AbstractFloats = Union{AbstractFloat,Complex{T} where T<:AbstractFloat}
+const AbstractFloats = Union{<:AbstractFloat,Complex{T} where T<:AbstractFloat}
 
 # May switch argument order
 """
--- a/src/dspbase.jl
+++ b/src/dspbase.jl
@@ -1,7 +1,6 @@
 # This file was formerly a part of Julia. License is MIT: https://julialang.org/license
 
 import Base: trailingsize
-import LinearAlgebra.BLAS
 
 const SMALL_FILT_CUTOFF = 58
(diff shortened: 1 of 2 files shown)


MikaelSlevinsky:
Hi there, the tests pass locally for me and the change seems rather benign overall. Could a maintainer please take a look and approve running workflows?

MikaelSlevinsky:
92/93! Yahoo! (I think that coverage is actually cumulative, so I'm not sure it's true that total coverage decreased.)

MikaelSlevinsky:
Ping

MikaelSlevinsky:
Hello, could a maintainer please take a look at this? Thank you

galenlynch:
Yes, I'll look soon, sorry!

src/dspbase.jl
```diff
@@ -677,6 +676,9 @@ function _conv(u, v, su, sv)
     _conv!(out, u, v, su, sv, outsize)
 end
 
+# We use this type definition for clarity
+const AbstractFloats = Union{AbstractFloat,Complex{T} where T<:AbstractFloat}
```
dehann, line 680:
```suggestion
const AbstractFloats = Union{<:AbstractFloat,Complex{T} where T<:AbstractFloat}
```

MikaelSlevinsky: Can you tell me the difference between the two?

dehann: hi, when you allow inheritance from the non-concrete type then the type inference system can better resolve potentially down to concrete types.  If you force non-concrete types then the type inference will stop at the non-concrete and require dynamic dispatch at run time.  I was a bit confused by this in the past myself, where the best examples go something like this:
```julia
ff(x::AbstractVector) = # allows inference to concrete types
ff(x::Union{AbstractVector,Real}) = # restricts to non-concrete types
ff(x::Union{<:AbstractVector, <:Real) = # allows inference down to concrete types
```

More pronounced in types (different example):
```julia
struct MyType1
  # restricts to non-concrete type
  dict::Dict{Symbol, AbstractSting}
end

struct MyType2
  # allows type inference try find anonymous concrete type
  dict::Dict{Symbol, <:AbstractSting}
end
```

Maybe newer versions of Julia cleaned up the syntax on this, but I'm just in the habit of adding the `<:` for performance and noticed it on the PR here.

MikaelSlevinsky: Thanks!

MikaelSlevinsky:
Any other changes requested?

src/dspbase.jl
martinholters: @dehann could you give a concrete example where adding the `<:` inside a `Union` like here is beneficial? (Note that `Union{<:AbstractFloat,Complex{T} where T<:AbstractFloat} == Union{AbstractFloat,Complex{T} where T<:AbstractFloat}` while `Dict{Symbol, <:AbstractString} != Dict{Symbol, AbstractString}`.) Otherwise I'd be in favor of removing the `<:`.

src/dspbase.jl
```diff
@@ -677,6 +676,9 @@ function _conv(u, v, su, sv)
     _conv!(out, u, v, su, sv, outsize)
 end
 
+# We use this type definition for clarity
+const AbstractFloats = Union{<:AbstractFloat,Complex{T} where T<:AbstractFloat}
```
martinholters, line 680: I'm too happy with the name `AbstractFloats`. Not that I have a good idea for an alternative, though. Maybe `RealOrComplexFloat`?

martinholters:
This should probably be ok. But once we get to addressing #292, it would be unclear which types to use frequency domain or time domain convolution for if the applicability of `fft` depends on whether some other package (GenericFFT) has been loaded. Given how long #292 has been open, that shouldn't stop this PR offering an immediate improvement for now, though.

src/dspbase.jl
MikaelSlevinsky: Hmm, these two functions shows that they both allow inference to concrete subtypes of one of the partners in the `Union`.
```julia
julia> ff(x::Union{AbstractVector, Real}) = norm(x)
ff (generic function with 1 method)

julia> @code_warntype ff([1,2,3])
MethodInstance for ff(::Vector{Int64})
  from ff(x::Union{Real, AbstractVector}) in Main at REPL[35]:1
Arguments
  #self#::Core.Const(ff)
  x::Vector{Int64}
Body::Float64
1 ─ %1 = Main.norm(x)::Float64
└──      return %1


julia> gg(x::Union{<:AbstractVector, <:Real}) = norm(x)
gg (generic function with 1 method)

julia> @code_warntype gg([1,2,3])
MethodInstance for gg(::Vector{Int64})
  from gg(x::Union{Real, AbstractVector}) in Main at REPL[38]:1
Arguments
  #self#::Core.Const(gg)
  x::Vector{Int64}
Body::Float64
1 ─ %1 = Main.norm(x)::Float64
└──      return %1
```

Presumably it didn't matter in the case at hand because the type union was always used as a type parameter, which should in and of itself allow inference.

There's also nothing in the documentation about type unions on using `<:` https://docs.julialang.org/en/v1/manual/types/#Type-Unions

src/dspbase.jl
MikaelSlevinsky: Your name is more descriptive, so I changed it.

martinholters (approved):
I think this fine. Any comments, @galenlynch? Merge and tag?

MikaelSlevinsky:
This can probably be merged, right? FYI, this benign change reduces my package load time by about 40%.

galenlynch:
Sorry I missed this, @martinholters. I'll try to finish #292. I got stuck figuring out how to best select between the many conv methods after implementing a bunch of different variants that work better in different situations. I think for now it may be better to accept occasionally choosing the wrong method in the name of being faster and more general, on average.

PR #481, opened 2022-09-15
Add Remez Filter Order Estimation (#429)

jsmithnh09:
This PR adds `remezord`, using the same passband/stopband ripple tolerances and normalized frequency edge parameters as the IIR order estimation routines. In this case, we're estimating the minimum order of an FIR filter. For the test cases, I added a few based on the results from [1] and cross-referencing the results from MATLAB's `firpmord`. 

[1] Rabiner, L. R., McClellan, J. H., & Parks, T. W. (1975). FIR digital filter design techniques using weighted Chebyshev approximation. Proceedings of the IEEE, 63(4), 595-610.

Hopefully this addresses everything in #429 !

### DIFF
--- a/src/Filters/filt_order.jl
+++ b/src/Filters/filt_order.jl
@@ -481,6 +481,5 @@ function remezord(Wp::Real, Ws::Real, Rp::Real, Rs::Real)
     B = (2.66e-3 * L1^2) + (0.5941 * L1) + 0.4278
     Kf = 0.51244 * (L1 - L2) + 11.01217
     D = A * L2 - B
-    L = ((D - Kf * df^2) / df) + 1
-    return ceil(Int, L) - 1
+    return ceil(Int, ((D - Kf * df^2) / df))
 end
--- a/src/Filters/filt_order.jl
+++ b/src/Filters/filt_order.jl
@@ -460,28 +460,27 @@ function cheb2ord(Wp::Tuple{Real,Real}, Ws::Tuple{Real,Real}, Rp::Real, Rs::Real
 end
 
 """
-    N = remezord(F₁, F₂, δ₁, δ₂)
+    N = remezord(Wp, Ws, Rp, Rs)
 
 Order estimation for lowpass digital filter cases based
-on the equations and coefficients in [^Rabiner]. Source equation
-returns the minimum filterlength, this returns the order, N=L-1. 
-Results are valid within 1.3% error. `F₁` and `F₂` are the normalized 
-passband and stopband frequencies, with `δ₁` indicating the passband ripple
-and `δ₂` is the stopband attenuation (linear.) 
+on the equations and coefficients in [^Rabiner]. The original
+equation returned the minimum filterlength, whereas this implementation 
+returns the order (N=L-1). Results are valid within 1.3% error. `Wp` and 
+`Ws` are the normalized passband and stopband frequencies, with `Rp` indicating 
+the passband ripple and `Rs` is the stopband attenuation (linear.) 
 
 [^Rabiner]: Herrmann, O., Lawrence R. Rabiner, and D. S. K. Chan. "Practical 
 design rules for optimum finite impulse response lowpass digital filters." Bell System 
 Technical Journal 52.6 (1973): 769-799.
 """
-function remezord(F₁::Real, F₂::Real, δ₁::Real, δ₂::Real)
-    (0 > F₁ > 0.5) || (0 > F₂ > 0.5) && throw(ArgumentError("Pass and stopband edges must be greater than DC and less than Nyquist."))
-    K = δ₁ / δ₂
-    l1, l2 = log10(δ₁), log10(δ₂)
-    df = abs(F₂ - F₁) # works in HPF case if passband/stopband edges are flipped
-    A = (5.309e-3 * l1^2) + (7.114e-2 * l1) - 0.4761
-    B = (2.66e-3 * l1^2) + (0.5941 * l1) + 0.42768
-    Kf = 0.51244 * log10(K) + 11.01217
-    D = A * l2 - B
+function remezord(Wp::Real, Ws::Real, Rp::Real, Rs::Real)
+    (0 > Wp > 0.5) || (0 > Ws > 0.5) && throw(ArgumentError("Pass and stopband edges must be greater than DC and less than Nyquist."))
+    L1, L2 = log10(Rp), log10(Rs)
+    df = abs(Ws - Wp) # works in HPF case if passband/stopband edges are flipped
+    A = (5.309e-3 * L1^2) + (7.114e-2 * L1) - 0.4761
+    B = (2.66e-3 * L1^2) + (0.5941 * L1) + 0.4278
+    Kf = 0.51244 * (L1 - L2) + 11.01217
+    D = A * L2 - B
     L = ((D - Kf * df^2) / df) + 1
     return ceil(Int, L) - 1
 end
--- a/src/Filters/filt_order.jl
+++ b/src/Filters/filt_order.jl
@@ -465,9 +465,12 @@ end
 Order estimation for lowpass digital filter cases based
 on the equations and coefficients in [^Rabiner]. The original
 equation returned the minimum filterlength, whereas this implementation 
-returns the order (N=L-1). Results are valid within 1.3% error. `Wp` and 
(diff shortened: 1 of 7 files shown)


jsmithnh09:
Interesting - it looks like the x64 ubuntu and macOS failed. I see the test in particular is for TF to ZPK. Did the test fail because of the use of `BigFloat`? I'm curious as to why the other tests passed, except for those two...

jsmithnh09:
@martinholters would you happen to have insight on this/the discrepancy?

src/Filters/filt_order.jl
```diff
@@ -430,31 +430,56 @@ lowest estimated filter order, with `ωn` specifying the cutoff or "-3 dB" frequ
...
     N, ωn
+end
+
+"""
+    N = remezord(Wp, Ws, Rp, Rs)
+
+Order estimation for lowpass digital filter cases based
+on the equations and coefficients in [^Rabiner]. The original
+equation returned the minimum filterlength, whereas this implementation 
+returns the order (N=L-1). Results are valid within 1.3% error. `Wp` and
```
martinholters, line 468: It's a bit unfortunate that this returns the order, but `remez` then takes the length. But I think `remez` is the one being inconsistent with taking `numtaps` instead of filter order. (Matlab's `remez` also takes the order.) But that ship has sailed for the moment...

But given that hardly anyone will look up the equation, I guess we can remove that sentence and replace it with a note warning about `remez` needing `L=N+1`.

test/filt_order.jl
```diff
@@ -208,8 +208,23 @@ end
...
+    # Using the test-cases highlighted in [^Parks] Figures 8 and 15.
+    #
+    # [^Parks]: Rabiner, L. R., McClellan, J. H., & Parks, T. W. (1975). 
+    #   FIR digital filter design techniques using weighted Chebyshev 
+    #   approximation. Proceedings of the IEEE, 63(4), 595-610.
+    #
+    @test remezord(0.41665, 0.49417, 0.0116, 0.0001) == 39
+    @test remezord(0.135, 0.203, 0.1, 0.1) == 10
+    @test remezord(0.203, 0.135, 0.1, 0.1) == 10 # flipped HPF case.
+    @test remezord(0.00963, 0.13271, 0.0116, 0.0001) == 24
```
martinholters, line 229: I've tried actually designing these filters with `remez` and found that I needed minimum filter lengths 36, 14, 14, and 29 to fulfill the specs. That's more than the "1.3% error" promised in the docstring. Any idea why that is?

Also, the docstring only talks about the lowpass case, but here, we test for highpass, too.

martinholters (commented):
All the unrelated formatting changes make this a bit hard to review as one has to search for the relevant bits...

jsmithnh09:
> All the unrelated formatting changes make this a bit hard to review as one has to search for the relevant bits...

I had opened this in VSCode, the Julia Language Server likely had auto-formatted this on save, my mistake!

src/Filters/filt_order.jl
jsmithnh09: If we're looking to be consistent, we could keep this in terms of `numtaps` as remez requires! I'm not tied to the current commit, I think having a consistent API is more important. Whichever direction we decide on, I think you make a good point about noting the order vs. length values in the function header.

test/filt_order.jl
jsmithnh09: The function implementation I'm providing in this commit is based on Equation(s) 3.142 through 3.145 in Rabiner & Gold Theory and application of Digital Signal Processing. Rabiner/Gold cite the original paper in the Appendix section, namely the paper in the function header, which uses the same coefficients. The original paper claimed "%1.3 percent error", but I think the textbook makes a better statement on the estimation:

> The initial value of N along with the specified values Fp, Fs and d1/d2 are the initial guess as to the appropriate design parameters. Depending on whether the actual values of d1 and d2 are too high or too low, the value of N should be raised or lowered accordingly until specifications are just met or exceeded.

"d1" and "d2" in this case refer to the passband and stopband attenuation. Could you provide what calls you made to remez, and have you done a comparison with `firpmord` in MATLAB to see if you received the same minimum filter orders?

jsmithnh09: As for the lowpass vs. highpass case, the design equation is based on the absolute frequency delta of the pass/stopbands, so I guess this is just a dangling test case we could eliminate. 

I appreciate the feedback you've provided so far, I don't have a full understanding or the context on what the roadmap for the DSP.jl repository should be.  I'm not looking to make the package more complicated. I hope you see that's the case as well, feel free to make suggestions if you think I'm not providing adequate code or good enough reasoning.

src/Filters/filt_order.jl
martinholters: Given that the IIR filter design functions all take order (what else should they), I think we should stay with order here, too, and live with `remez` being the odd one out for the moment.

Going forward, maybe we can change `remez` to take (mutually exclusive) keyword arguments `order` and `numtaps` and deprecate the positional `numtaps` argument. But that's clearly not in the scope of this PR.

test/filt_order.jl
martinholters: Ok, maybe replace the 1.3% claim then with a more general warning that this is just an estimate?

This is what I did (and yes, I have to correct myself, for the case, even length 35 is sufficient):
```julia
julia> function testfilt(L, Wp, Ws, Rp, Rs)
           h = remez(L, [0, Wp, Ws, 0.5], [1, 0]; weight = 1 ./ [Rp, Rs], grid_density = 32)
           ω = range(0, π, length = 65536)
           H = freqresp(PolynomialRatio(h, [1]), ω)
           ip = findlast(<=(2π*Wp), ω)
           is = findfirst(>=(2π*Ws), ω)
           return maximum(x -> abs(abs(x) - 1), H[1:ip]) <= Rp && maximum(abs, H[is:end]) <= Rs
       end
testfilt (generic function with 1 method)

julia> testfilt(34, 0.41665, 0.49417, 0.0116, 0.0001)
false

julia> testfilt(35, 0.41665, 0.49417, 0.0116, 0.0001)
true

julia> testfilt(13, 0.135, 0.203, 0.1, 0.1)
false

julia> testfilt(14, 0.135, 0.203, 0.1, 0.1)
true

julia> testfilt(28, 0.00963, 0.13271, 0.0116, 0.0001)
false

julia> testfilt(29, 0.00963, 0.13271, 0.0116, 0.0001)
true

julia> function testfilt_hp(L, Wp, Ws, Rp, Rs)
           h = remez(L, [0, Ws, Wp, 0.5], [0, 1]; weight = 1 ./ [Rp, Rs], grid_density = 32, filter_type = isodd(L) ? filter_type_bandpass : filter_type_hilbert)
           ω = range(0, π, length = 65536)
           H = freqresp(PolynomialRatio(h, [1]), ω)
           ip = findlast(<=(2π*Ws), ω)
           is = findfirst(>=(2π*Wp), ω)
           return maximum(abs, H[1:ip]) <= Rp && maximum(x -> abs(abs(x) - 1), H[is:end]) <= Rs
       end
testfilt_hp (generic function with 1 method)

julia> testfilt_hp(13, 0.203, 0.135, 0.1, 0.1)
false

julia> testfilt_hp(14, 0.203, 0.135, 0.1, 0.1)
true
```
> what the roadmap for the DSP.jl repository should be

Yea, as if there was one...

martinholters: FWIW, I stumbled upon two formulas attributed to Bellanger and Kaiser, respectively, which give lengths 43, 10, 27 and 42, 9, 27 for these test cases. So that underlines that these are all estimates and ave to be treated as such.

So for me the take-away is that we should just reformulate the docstring to be more defensive.

martinholters (commented):
Put something in `docs/src/filters.md` to make this discoverable and it should be ready to go.

martinholters (approved):

martinholters:
And now, test failure in "TF to ZPK" is back despite #483, this time hitting Windows only. I'll prod at that once more...

martinholters:
Given that 3b6889f2683e90ba7cad7c7283184e463f48dfb3 had passed tests and there were only doc changes since then, I'm going to merge.

jsmithnh09:
Sorry to see that happening again - Martin, just wanted to say thank you for the feedback and continuing to support this package. Always constructive and happy to see there's (some) community still active!

PR #483, opened 2022-10-04
Rework "TF to ZPK" test to cope with OpenBLAS update

martinholters:
OpenBLAS v0.3.21 gives slightly different eigen values, leading to different polynomial roots, which made the old test fail. In particular, the DC gain of the filter is now `0.9999999999999846` (instead of `1.0000000000000122` with the previous OpenBLAS version), while the Matlab based coefficients yield `0.9999999999999951`. Thus the error is more than twice that of the Matlab version, making the test fail. But of course, both results are sufficiently close to 1 to be acceptable.

This replaces the test with an error across a larger frequency range where indeed, the Matlab based version is still worse.

@jsmithnh09 this is the reason for the failures seen over at #481.

### DIFF
--- a/test/filter_design.jl
+++ b/test/filter_design.jl
@@ -1,5 +1,6 @@
 !(dirname(@__FILE__) in LOAD_PATH) && push!(LOAD_PATH, dirname(@__FILE__))
 using DSP, Test, FilterTestHelpers
+using LinearAlgebra: norm
 using DelimitedFiles: readdlm
 
 #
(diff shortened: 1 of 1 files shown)


test/filter_design.jl
```diff
@@ -364,8 +363,19 @@ end
...
+    H_ref = freqresp(Butterworth(BigFloat, 20), freq)
+    H_1 = freqresp(f, freq)
+    H_2 = freqresp(ZeroPoleGain{:s}(Float64[], m_p, 1), freq)
+    # both should be good
+    @test H_1 ./ H_ref ≈ ones(length(freq))
+    @test H_2 ./ H_ref ≈ ones(length(freq))
+    # ... but ours (H1) should be the better one
+    delta_1 = H_1 ./ H_ref .- 1
+    delta_2 = H_2 ./ H_ref .- 1
+    @test norm(delta_1) < norm(delta_2)
```
martinholters, line 378: I'm not sure comparing to the coefficients obtained with a 10(?)-year old Matlab version makes much sense. Isn't it sufficient to verify that ours are good?
```suggestion
    H_1 = freqresp(f, freq)
    @test H_1 ./ H_ref ≈ ones(length(freq))
```

PR #492, opened 2022-12-06
Standardize the CI setup

ararslan:
Currently this package is using a rather comical amount of CI resources, as it runs tests on Julia versions 1.0, 1.1, 1.2, 1.3, 1.4, 1.5, 1.6, stable, and nightly, all on Linux, macOS, and Windows, both x86 and x86-64. The more standard approach is to run on the earliest supported Julia version, current stable, and nightly with Linux x86-64, then add other platforms as desired running only on current stable. Things have now been set up here accordingly, with Linux x86, macOS x86-64, Windows x86, and Windows x86-64 as additions to the CI matrix.

The test steps have also been standardized, using `v1` specifiers rather than manually adjusting action versions, using the standard set of Julia-specific actions, and swapping the (rather archaic) Coveralls service for Codecov.

Documentation has been included as a job in CI rather than a separate workflow, as its required setup is identical that for CI and most packages have things set up this way.

The CompatHelper setup has been adjusted to be better synced with the official template (though the template wasn't copied wholesale as it's unnecessarily complex at present).

TagBot has been synced with the official template.

Standardizing the CI setup will provide a more consistent experience for contributors and should generally ease maintenance burden.


martinholters:
Would we want to set `permissions` also in `CI.yml`?

ararslan:
What do we need permissions for in CI?

martinholters:
Is the question "which permissions do we need?" or "why should we bother setting permissions?"?

For the former: I guess `contents: write`. For the `test`  job, `contents: read` why be sufficient, but the `docs` job needs write access for deployment. Not sure if anything else is needed.

For the latter: If not specifying any permissions, everything is allowed, so if any of the used actions goes haywire, it can do all sorts of bad stuff. And with the lower-bound-only version specifications, we allow using updated versions of actions without checking them first.

ararslan:
> Is the question "which permissions do we need?" or "why should we bother setting permissions?"?

I suppose I meant the latter. I don't know of any Julia packages that explicitly set permissions for tests and documentation, and it isn't part of the recommended templates aside from TagBot. We're primarily using `julia-actions` actions so I'm not particularly worried, honestly.

martinholters:
Fair point.

PR #500, opened 2023-04-18
Support `dims` kwarg in `rms`

### LINKED ISSUE #499: Make `rms` work on dimension of arrays
it would be convenient to do `rms` on multiple channel of signals like `mean`

martinholters:
Closes #499.

### DIFF
--- a/src/util.jl
+++ b/src/util.jl
@@ -4,6 +4,7 @@ import Base: *
 import LinearAlgebra.BLAS
 using LinearAlgebra: mul!
 using FFTW
+using Statistics: mean
 
 export  hilbert,
         fftintype,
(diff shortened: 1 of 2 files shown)


src/util.jl
```diff
@@ -162,11 +163,18 @@ Convert an amplitude ratio to dB (decibel), or ``20
...
-Return the root mean square of signal `s`.
+Return the root mean square (rms) of signal `s`. Optional keyword parameter
+`dims` can be used to specify the dimensions along which to compue the rms.
 """
-rms(s::AbstractArray{T}) where {T<:Number} = sqrt(sum(abs2, s)/length(s))
+function rms(s::AbstractArray{T}; dims=:) where {T<:Number}
+    if dims === (:)
+        return sqrt(mean(abs2, s))
+    else
+        return sqrt.(mean(abs2, s; dims=dims))
```
martinholters, line 175: This would also work if `dims === (:)`, but that assumes that the result `Number` of `mean` broadcasts as a singleton, which I didn't feel 100% comfortable relying on, therefore the branch to avoid the broadcast when `dims === (:)`.

martinholters:
```
   Expression: rms(x; dims=1) ≈ sqrt.(mean(abs.(x) .^ 2; dims=1))
  MethodError: no method matching mean(::typeof(abs2), ::Array{Float64,2}; dims=1)
```
Right, I had forgotten we still support Julia 1.0 where this method of `mean` is not supported, yet.

src/util.jl
martinholters: Well, also `mean(f, X; dims)` doesn't exist on Julia 1.0, so this needs a special case anyway.

martinholters:
Anyone willing to give this another pair of eyes? Otherwise, I'll go ahead and merge in a few days.

martinholters:
> If compat is going to be raised to 1.6, wouldn't it be fine to return to `mean(abs2, s; dims=dims)`?

Absolutely. If #532 merges before this one, I'll revert to `mean`. (Otherwise, the present workaround should not be worse except for  its verboseness.)

PR #502, opened 2023-05-24
Create WelchConfig object

haberdashPI:
This allows `welch_pgram` to use a predefined set of configuration parameters in an object named `WelchConfig`, rather than passing these configuration values by keyword argument. FFT plans and intermediate buffers are preconfigured in `WelchConfig` making repeated calls that use the same config object faster.

This also sets `MTConfig` and `WelchConfig` to be children of the abstract type `AbstractPGramConfig`

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1204607356620460

### DIFF
--- a/src/periodograms.jl
+++ b/src/periodograms.jl
@@ -39,7 +39,7 @@ struct ArraySplit{T<:AbstractVector,S,W} <: AbstractVector{Vector{S}}
     end
 
 end
-ArraySplit(s::AbstractVector, n, noverlap, nfft, window; buffer) =
+ArraySplit(s::AbstractVector, n, noverlap, nfft, window; buffer=nothing) =
     ArraySplit{typeof(s),fftintype(eltype(s)),typeof(window)}(s, n, noverlap, nfft, window; buffer)
 
 function Base.getindex(x::ArraySplit{T,S,Nothing}, i::Int) where {T,S}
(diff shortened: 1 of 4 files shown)


src/periodograms.jl
```diff
@@ -354,6 +354,65 @@ forward_plan(X::AbstractArray{T}, Y::AbstractArray{Complex{T}}) where {T<:Union{
...
+"""
+    WelchConfig(data; n=length(signal)>>3, noverlap=n>>1,
+             onesided=eltype(signal)<:Real, nfft=nextfastfft(n),
+             fs=1, window=nothing)
+
+    WelchConfig(nsamples, eltype; n=nsamples>>3, noverlap=n>>1,
+             onesided=eltype<:Real, nfft=nextfastfft(n),
+             fs=1, window=nothing)
+
+Captures all configuraiton options for the [welch_pgram](@ref) in a single struct. (Akin to
```
martinholters, line 388:
```suggestion
Captures all configuration options for the [welch_pgram](@ref) in a single struct. (Akin to
```

martinholters:
Should we introduce
```julia
periodogram(s::AbstractVector{T}, config::WelchConfig) where T<:Number = welch_pgram(s, config)
periodogram(signal::AbstractVector, config::MTConfig) = mt_pgram(signal, config)
```
? I'm not sufficiently familiar with these function to tell whether this makes sense, but if it doesn't, the common supertype `AbstractPGramConfig` would be questionable as well.

martinholters (commented):
Testing `welch_pgram!` would be good to get it covered as well.

martinholters:
> On the topic of abstract supertypes, `Filter` doesn't seem to be doing anything

True. I wouldn't mind it being deleted if it bothers you.

PR #503, opened 2023-06-28
Add frequency estimators by Jacobsen and Quinn

mbaz:
This commit adds two new frequency estimation algorithms:

    * `jacobsen`, a very fast algorithm that performs a three-point
      curve fit around the DFT peak. Not super accurate.
    * `quinn`, a very accurate iterative algorithm based on ARMA(2,2).
      It requires an initial frequency estimate (which may be
      provided by `jacobsen`).

    Both algorithms work with real and complex signals.


src/estimation.jl
```diff
@@ -38,4 +40,166 @@ function esprit(x::AbstractArray, M::Integer, p::Integer, Fs::Real=1.0)
...
+        δ = -real((X[k+1] - X[k-1]) / (2X[k] - X[k-1] - X[k+1]) )
+    else
+        δ = 0.0
+    end
+
+    f + δ*Fs/N
+end
+
+"""
+    qf(x::Vector, f0::Real = 0.0, Fs::Real = 1.0 ; tol = 1e-6, maxiters = 20)
```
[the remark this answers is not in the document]
mbaz: Thank you for the review!

I've pushed a new commit fixing this; in addition, I added more tests and fixed a bug in `quinn` for complex inputs.

ViralBShah:
Is this good to merge?

mbaz:
Added a couple of tests to include the two lines that were missing coverage.

ViralBShah:
Perhaps good to wait a few days for @martinholters to take a look.

src/estimation.jl
```diff
@@ -38,4 +40,166 @@ function esprit(x::AbstractArray, M::Integer, p::Integer, Fs::Real=1.0)
...
+complex, the algorithm is [^Quinn2009].
+
+[^Quinn1991]: B Quinn and J Fernandes, "A fast efficient technique for the
+estimation of frequency", Biometrika, Vol. 78 (1991).
+
+[^Quinn2009]: B Quinn, "Recent advances in rapid frequency estimation", Digital
+Signal Processing, Vol. 19 (2009), Elsevier.
+
+"""
+quinn(x ; kwargs...) = quinn(x, 0.0, 1.0 ; kwargs...)
```
martinholters, line 117:
```suggestion
quinn(x ; kwargs...) = quinn(x, jacobsen(x, Fs), 1.0 ; kwargs...)
```

src/estimation.jl
```diff
@@ -38,4 +40,166 @@ function esprit(x::AbstractArray, M::Integer, p::Integer, Fs::Real=1.0)
...
+
+quinn(x, Fs ; kwargs...) = quinn(x, 0.0, Fs ; kwargs...)
```
martinholters, line 119:
```suggestion
quinn(x, Fs ; kwargs...) = quinn(x, jacobsen(x, Fs), Fs ; kwargs...)
```

src/estimation.jl
```diff
@@ -38,4 +40,166 @@ function esprit(x::AbstractArray, M::Integer, p::Integer, Fs::Real=1.0)
...
+    fₙ = Fs/2
+    T = length(x)
+
+    # Run a quick estimate of largest sinusoid in x
+    if f0 == 0.0
+        f_est = jacobsen(x, Fs)
+        ω̂ = π*f_est/fₙ  # estimate ranges from 0 to π
+    else
+        ω̂ = π*f0/fₙ
+    end
```
martinholters, line 131:
```suggestion
    ω̂ = π*f0/fₙ
```
(with above changes to the other methods)

src/estimation.jl
```diff
@@ -38,4 +40,166 @@ function esprit(x::AbstractArray, M::Integer, p::Integer, Fs::Real=1.0)
...
+    fₙ = Fs/2
+    T = length(x)
+
+    # Run a quick estimate of largest sinusoid in x
+    if f0 == 0.0
+        f_est = jacobsen(x, Fs)
+        ω̂ = π*f_est/fₙ
+    else
+        ω̂ = π*f0/fₙ
+    end
```
martinholters, line 172:
```suggestion
    ω̂ = π*f0/fₙ
```
(same here)

src/estimation.jl
```diff
@@ -38,4 +40,166 @@ function esprit(x::AbstractArray, M::Integer, p::Integer, Fs::Real=1.0)
...
+            ξ[t] = x[t] + exp(complex(0,ω̂))*ξ[t-1]
+        end
+        # step 3
+        S = let s = 0.0
+                for t=2:T
+                    s += x[t]*conj(ξ[t-1])
+                end
+                s
+            end
+        num = imag(S*exp(complex(0,-ω̂)))
```
martinholters, line 194:
```suggestion
        num = imag(S*cis(-ω̂)))
```

src/estimation.jl
```diff
@@ -38,4 +40,166 @@ function esprit(x::AbstractArray, M::Integer, p::Integer, Fs::Real=1.0)
...
+    N = length(x)
+    X = rfft(x)
+    k = argmax(abs.(X))  # index of DFT peak
+    if (k+1 <= N) && (k-1 >= 1)
+        δ = -real((X[k+1] - X[k-1]) / (2X[k] - X[k-1] - X[k+1]) )
+    else
+        δ = 0.0
+    end
+
+    (k + δ)*Fs/N
```
martinholters, line 62: This seems to be off by one:
```suggestion
    return (k - 1 + δ)*Fs/N
```

src/estimation.jl
```diff
@@ -38,4 +40,166 @@ function esprit(x::AbstractArray, M::Integer, p::Integer, Fs::Real=1.0)
...
+    quinn(x::Vector, Fs::Real = 1.0 ; kwargs...)
+
+    quinn(x::Vector ; kwargs...)
+
+Algorithms by Quinn and Quinn & Fernandes for frequency estimation. Given a
+signal `x` and an initial guess `f0`, estimate and return the frequency of the
+largest sinusoid in `x`. `Fs` is the sampling frequency. All frequencies are
+expressed in Hz.
+
+If the initial guess `f0` is not provided or if it is equal to zero, then a guess
```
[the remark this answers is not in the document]
mbaz: Agreed -- it's cleaner not to special-case `f0 = 0`.

src/estimation.jl
mbaz: Yes, `k` here is off by one from the DFT definition. I will add some related tests...

mbaz:
@martinholters Thank you for the review!

src/estimation.jl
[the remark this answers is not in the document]
martinholters: Yes, better, but the docs need to be updated.

src/estimation.jl
```diff
@@ -38,4 +40,155 @@ function esprit(x::AbstractArray, M::Integer, p::Integer, Fs::Real=1.0)
...
+            ξ[t] = x[t] + exp(complex(0,ω̂))*ξ[t-1]
+        end
+        # step 3
+        S = let s = 0.0
+                for t=2:T
+                    s += x[t]*conj(ξ[t-1])
+                end
+                s
+            end
+        num = imag(S*cis(-ω̂)))
```
martinholters, line 183:
```suggestion
        num = imag(S*cis(-ω̂))
```
(my bad)

src/estimation.jl
```diff
@@ -38,4 +40,155 @@ function esprit(x::AbstractArray, M::Integer, p::Integer, Fs::Real=1.0)
...
+end
+
+function quinn(x::Vector{<:Complex}, f0::Real, Fs::Real ; tol = 1e-6, maxiters = 20)
+    fₙ = Fs/2
+    T = length(x)
+
+    ω̂ = π*f0/fₙ
+
+    # Remove any DC term in x
+    x .= x .- complex(mean(real.(x)), mean(imag.(x)))
```
martinholters, line 164: Why not
```suggestion
    x .= x .- mean(x)
```
?

test/estimation.jl
```diff
@@ -19,3 +19,54 @@ using DSP, Test
...
 end
+
+@testset "jacobsen" begin
+    # real input
+    fs = 100
+    t = range(0, 5, step = 1/fs)
+    fr = 28.3
+    sr = cos.(2π*fr*t .+ π/4.2)
+    f_est_real = jacobsen(sr, fs)
+    @test isapprox(f_est_real, fr, atol = 0.5)
```
martinholters, line 30: In the complex case we have a tighter bound:
```suggestion
    @test isapprox(f_est_real, fr, atol = 1e-2)
```

But if I'm not mistaken, the frequency resolution of the underlying FFT is 0.2 here and hence simple peak-picking should be within 0.1, so `atol` should definitely be lower than that, but I'm not sure what one can expect.

test/estimation.jl
```diff
@@ -19,3 +19,54 @@ using DSP, Test
...
+    @test jacobsen(zeros(10)) == 0.1
```
martinholters, line 31:
```suggestion
```
I don't think the frequency estimate of an all-zero signal is something we should make any guarantees about. (But if we do, 0.1 here seems rather surprising.)

test/estimation.jl
```diff
@@ -19,3 +19,54 @@ using DSP, Test
...
+    # complex input
+    fc = -40.3
+    sc = cis.(2π*fc*t .+ π/1.4)
+    f_est_complex = jacobsen(sc, fs)
+    @test isapprox(f_est_complex, fc, atol = 1e-2)
+    @test jacobsen(zeros(ComplexF64, 10)) == -0.5
```
martinholters, line 37:
```suggestion
```
Why should the frequency of an all-zero signal depend on the type (wheher complex or not)?

src/estimation.jl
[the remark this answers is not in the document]
mbaz: Done -- the description of optional arguments has been improved.

src/estimation.jl
mbaz: Fixed

src/estimation.jl
mbaz: Fixed -- I either missed this method of `mean`, or it didn't exist when I first wrote this code.

test/estimation.jl
mbaz: This was a side effect of the off-by-one error in `jacobsen`. I've re-written these tests and added some more. For `jacobsen`, now `atol = 1e-5` is used.

test/estimation.jl
mbaz: Again, an effect of the off-by-one bug. This has been fixed.

test/estimation.jl
mbaz: For all-zero (both real and complex signals), the DFT is constant. In this case, `argmax` returns the first array element. For real signals, the first frequency is zero (since we use `rfft`), but for complex signals it is negative Nyquist.

The purpose of these tests was to make sure all code in `jacobsen` is covered by the tests (codecov identified a branch that was not tested). This branch is taken when the DFT peak occurs in one of the two extreme frequencies. I've now added better tests for this scenario, covering both extremes of the spectrum.

src/estimation.jl
```diff
@@ -38,4 +40,157 @@ function esprit(x::AbstractArray, M::Integer, p::Integer, Fs::Real=1.0)
...
+Signal Processing, Vol. 19 (2009), Elsevier.
+
+"""
+quinn(x ; kwargs...) = quinn(x, jacobsen(x, 1.0), 1.0 ; kwargs...)
+
+quinn(x, Fs ; kwargs...) = quinn(x, jacobsen(x, Fs), Fs ; kwargs...)
+
+function quinn(x::Vector{<:Real}, f0::Real, Fs::Real ; tol = 1e-6, maxiters = 20)
+    fₙ = Fs/2
+    T = length(x)
```
martinholters, line 125: Nitpick, but for `jacobsen`, this is called `N`, which I'd prefer here, too, not only for consistency, but also because `T` suggests either a time/duration in seconds or a -- in the julia context -- a type.

test/estimation.jl
```diff
@@ -19,3 +19,77 @@ using DSP, Test
...
+    sr = sin.(2π*fr*t .+ 3π/2.2)
+    f_est_real = jacobsen(sr, fs)
+    @test isapprox(f_est_real, fr, atol = 1e-5)
+    # test at higher extreme of DFT
+    fr = 49.9002
+    sr = cos.(2π*fr*t)
+    f_est_real = jacobsen(sr, fs)
+    @test isapprox(f_est_real, fr, atol = 1e-5)
+    # test at lower extreme of DFT
+    @test jacobsen(zeros(10)) == 0.0
```
martinholters, line 41:
```suggestion
    @test jacobsen(ones(10)) == 0.0
```
to have a well-defined correct result?

martinholters (commented):
Some more nit-picking...

src/estimation.jl
mbaz: The reason `T` is used here is to follow the paper's nomenclature. The code mimics all variables in the paper (for example `ξ`, `β`, etcetera). I think this makes it easier to study the code and the paper side-by-side.

Having said that, I agree that in this case `T` may be too confusing. I'll make this change.

ViralBShah:
@mbaz In preparation for a 0.8 release, would it be possible for you to get this PR ready?

mbaz:
@ViralBShah I'm still actively working on it. Two things happened: (1) I realized applying Jacobsen to real signals requires a bit of care, and (2) my work load peaked the past two weeks.

My plan at the moment is to have a single Jacobsen estimator based on the DFT. If/when I work out the issues with an RFFT-based estimator, I'll make another PR. How does that sound?

What is the timeline for 0.8? I will try to finish up this PR in the next couple of days.

ViralBShah:
Thank you. There is no tight deadline, and we can always add new features in new releases. I am just thinking out aloud, but if we can get as much of the backlog of PRs in as we can in the next 2 weeks or so - that will make for a nice 0.8 release.

@martinholters I would of course defer to you on that.

ViralBShah:
@mbaz Just a gentle nudge to see if possible to finish this PR. 0.8 may be in a week or two. But of course, if you are busy, there's always 0.9!

mbaz:
The test failure seems to be related to codecov.

test/estimation.jl
```diff
@@ -21,25 +21,9 @@ using DSP, Test
...
-    @test isapprox(f_est_real, fr, atol = 1e-5)
-    fr = 12.45
-    sr = sin.(2π*fr*t .+ 3π/2.2)
-    f_est_real = jacobsen(sr, fs)
-    @test isapprox(f_est_real, fr, atol = 1e-5)
-    # test at higher extreme of DFT
-    fr = 49.9002
-    sr = cos.(2π*fr*t)
-    f_est_real = jacobsen(sr, fs)
-    @test isapprox(f_est_real, fr, atol = 1e-5)
```
martinholters: These tests seem useful, assuming they pass. If they don't pass we should probably try to make them. Or is there a fundamental reason they cannot? Then the documentation of `jacobsen` should probably mention this limitation.

src/estimation.jl
```diff
@@ -44,29 +44,19 @@ function esprit(x::AbstractArray, M::Integer, p::Integer, Fs::Real=1.0)
...
-    end
-
-    return (k - 1 + δ)*Fs/N
-end
-
-function jacobsen(x::Vector{<:Complex}, Fs::Real = 1.0)
+function jacobsen(x::AbstractVector, Fs::Real = 1.0)
     N = length(x)
     X = fftshift(fft(x))
     k = argmax(abs.(X))  # index of DFT peak
```
martinholters, line 62: For real `x`, it appears to be a matter of chance which of the two maxima (due to the symmetry) we get. Should we limit the search range to 0 ... Fs/2 in that case?

test/estimation.jl
mbaz: These tests all pass, and I can add back if you'd prefer. The reason to remove them is that, in general, Jacobsen (like other estimators designed for complex signals) shouldn't be used with real signals, since the results can be quite inaccurate.

Would it be better to enforce that the signal is complex? I didn't do that because I wanted the argument types to be consistent with our ESPRIT implementation.

src/estimation.jl
mbaz: Well, since the algorithm (as mentioned in my comment above) is designed for complex signals, if someone uses it with real signals, caveat emptor applies...

I'll be happy to implement this in the way that you prefer. I can add a warning to the docs, or I can force the return value to be positive, or even restrict the signal type to complex.

mbaz: I decided to make these changes:

* Add explanation/warning about using Jacobsen with real signals to Jacobsen's docstring.
* Also document that, if input is real, the frequency estimate will be positive.
* Add tests for Jacobsen with real signals. The signal's frequency  in these tests is close to Fs/4, where the estimation error is at its lowest.

src/estimation.jl
```diff
@@ -38,4 +40,157 @@ function esprit(x::AbstractArray, M::Integer, p::Integer, Fs::Real=1.0)
...
+    N = length(x)
+    X = fftshift(fft(x))
+    k = argmax(abs.(X))  # index of DFT peak
+    fpeak = fftshift(fftfreq(N, Fs))[k]  # peak frequency
+    if ((k != N) && (k != 1))  # avoid out-of-bounds indexing
+        # jacobsen's formula
+        δ = -real((X[k+1] - X[k-1]) / (2X[k] - X[k-1] - X[k+1]) )
+    else
+        δ = 0.0
+    end
```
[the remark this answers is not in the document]
martinholters: This still deserves an answer.

src/estimation.jl
martinholters: Sounds good.

src/estimation.jl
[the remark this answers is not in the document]
mbaz: Sorry -- I didn't mean to leave you hanging, I just thought this was made moot by my decision not to special-case Jacobsen for real inputs.

Your suggestion here is definitely correct if Jacobsen is applied to a real signal. The issue is that Jacobsen assumes a complex sinuosoid, which means that the frequency estimate will be inaccurate, especially near the extremes of the spectrum.

I considered the following options:
* Offer a version of Jacobsen specialized for real signals, even though we know it is suboptimal; the main benefit is using `rfft` instead of `fft`, which halves the number of calculations.
* Offer a generic, `fft`-based version of Jacobsen that still works for real signals, with a warning in the docstring.

The current PR implements the second option. I don't think maintaining a sub-optimal version of Jacobsen is worth it, especially when this PR also adds an estimator for real signals (Quinn).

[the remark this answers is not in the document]
martinholters: We might be talking past each other. I specifically mean the code suggestion for the complex case. I think it both simplifies the code and gives better results for `k==1` and `k==N` by exploiting that theoretically `X[0]==X[N]` and `X[N+1]==X[1]` due to the periodicity of the FT. (If the input is real-valued, due to the symmetry, I think the computation will give `δ==0.0`, so it just won't help then, but won't be wrong either.)

And if we do this, I think we can eliminate the `fftshift`s for a slight performance improvement, though that might not be relevant.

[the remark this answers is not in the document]
mbaz: Thanks for your patience. I actually implemented this but, by accident, left the code out of the PR... :facepalm: 

I've pushed a commit which is more verbose, but (I think) easier to read, and also gets rid of `fftshift`.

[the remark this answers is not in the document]
mbaz: Just pushed a commit to improve code coverage.

martinholters (approved):
Thanks for your patience. I think it was worth it.

mbaz:
Thank you too, @martinholters -- the code is much better now than it was in the original PR.

PR #514, opened 2023-11-29
update tukey docs

haakon-e:
fix docs typos in definition of tukey window


ViralBShah:
Can we get another set of eyes here to review and possibly merge?

PR #520, opened 2023-12-28
Add doctest to Xcorr

ErikBuer:
Hi.
  
As this is my first PR for the project, I figured I'd make it small, just to get through the process.
  
The PR adds a doctest to the xcorr function. 
When running the doctest, I noticed some other tests failing, so I corrected the output such that all tests passed.
There are also some small linter corrections in the markdown file.
  
  
I could not find any developer instructions for the project?
It could also be nice with some defined scope, what is the goal, and so on.

I prefer doctests, because these add examples that let the users play around with a working example.
If this is something that resonates, I may go over and make some examples, as I see a lot of functions lacking examples.


ViralBShah:
Good to merge - perhaps after the docs 1.0 PR?

ViralBShah:
@ErikBuer Generally many Julia packages do not have well-defined scope, but it tends to be defined by the current set of maintainers (a list which does change and it's always great to bring new people in to drive a project forward). Major deviations or change in directions are often discussed on discourse first (for example, should funcitonality be in multiple packages).

ViralBShah:
@wheeheee This seems good to merge.

PR #521, opened 2023-12-29
Change documentation for xcorr and conv to be true to the actual implementation.

ErikBuer:
It looks like there was an intention to implement a clever selection between time domain and frequency domain convolution, but this was never implemented.

The documentation currently says that one of the two domain methods are used, where in reality only FFT convolution is implemented. As it is about 5 years since the functions was last changed, I **feel** it is time to let go, and just describe what the function does **currencty**.

If you have any thoughts, I'd love to hear them :)

### DIFF
--- a/src/dspbase.jl
+++ b/src/dspbase.jl
@@ -683,8 +683,8 @@ const RealOrComplexFloat = Union{AbstractFloat, Complex{T} where T<:AbstractFloa
 """
     conv(u,v)
 
-Convolution of two arrays. Uses either FFT convolution or overlap-save,
-depending on the size of the input. `u` and `v` can be  N-dimensional arrays,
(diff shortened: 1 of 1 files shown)


ErikBuer:
Hmm. I may have mixed this up a bit.

Yes it appears it does "overlap-save?" in ´unsafe_conv_kern_os!´.

I mistok the overlap-save for a time-domain convolution, that was listed as "TODO" in `_conv!`

```julia
# Does convolution, will not switch argument order
function _conv!(out, u, v, su, sv, outsize)
    # TODO: Add spatial / time domain algorithm
    _conv_fft!(out, u, v, su, sv, outsize)
end
```

Thank you for the correction @wheeheee  :)

PR #524, opened 2024-02-05
Update CI.yml

ViralBShah:
Run CI on LTS and stable.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -12,25 +12,20 @@ jobs:
       fail-fast: false
       matrix:
         version:
-          - '1.0'  # Oldest version supported by this package
+          - '1.6'  # Oldest version supported by this package
           - '1'    # Current stable version
-          - nightly
         os:
           - ubuntu-latest
+          - macos-latest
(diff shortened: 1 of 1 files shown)


ViralBShah:
The codecov app is installed. I have bumped you to be an owner of the organization for now so that we can get this all set up correctly.

.github/workflows/CI.yml
```diff
@@ -12,26 +12,21 @@ jobs:
       fail-fast: false
       matrix:
         version:
-          - '1.0'  # Oldest version supported by this package
+          - '1.6'  # Oldest version supported by this package
```
martinholters, line 15: This doesn't match
https://github.com/JuliaDSP/DSP.jl/blob/d85a9ebbc6bf5fdc58486955f8bdac7405fda215/Project.toml#L25
and I think that if we claim to support Julia down to 1.0, we should test against it.
But I guess it's quite reasonable to actually bump the compat to `1.6`.

ViralBShah: I think 1.6 should be reasonable.

PR #525, opened 2024-02-05
add codecov token secret to CI.yml

wheeheee:
also added CODECOV_TOKEN Dependabot repository secret

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -40,6 +40,7 @@ jobs:
       - uses: codecov/codecov-action@v4
         with:
           files: lcov.info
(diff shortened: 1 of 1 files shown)


ViralBShah:
I don't think we want CI to fail in case codecov upload errors. Let's see how it works out for the foreseeable future, and if we need to be more strict.

ViralBShah (approved):

ViralBShah:
It seems that the general consensus is that we should use `fail_ci_if_error: true`. @wheeheee would you be able to add that in a new PR? Sorry for setting you on the wrong path here.

PR #526, opened 2024-02-05
Bump Documenter.jl to v1, fix typos

wheeheee:
- index.md points to the contents page now, as it's the default main page
- updated Documenter compat to "1"
- indented references so they show up in the footnotes, nicely formatted
- fixed typos


ViralBShah (approved):

ViralBShah:
I'm ok with merging this quickly so that it can unlock other doc PRs. I think we can fix anything that is not perfect in a follow-up PR.

Issue #518, opened 2023-12-05
No documentation for DSP.Filters.bilinear()


anchal-physics:
It felt weird that a bilinear transform would not be provided in DSP.jl. I searched everywhere in the documentation and was almost about to write my own when I found a [discourse](https://discourse.julialang.org/t/not-sure-how-to-apply-a-filter/50265/6) where someone had used this function. Please add bilinear to the documentation.

ViralBShah:
PR appreciated!

PR #527, opened 2024-02-06
Adding documentation for bilinear function

anchal-physics:
This addresses https://github.com/JuliaDSP/DSP.jl/issues/518

This PR adds documentation for the bilinear function in the `Filters` module. docstrings are added in source code for the two methods of bilinear function with the top level method explained in detail to providde a clear explanation of what the transform function is doing.

The bilinear function is also added to the export list in `Filters.jl` so that it is available for Documenter and also for users to use.

The documentation is added to the `docs/src/filters.md` file in the `Filters` module, Filter design section where the bilinear function is added as a general digital filter design method.

No code has been changed in this PR, only documentation has been added and an extra function has been added to the export list.


ViralBShah:
I wonder if we should tag a 0.8 release in the next couple of weeks or so. 

If someone can review this and merge, that will be great to get into the new release (but of course it is a doc change and can also happen any time).

martinholters (approved):
LGTM, thanks.

PR #528, opened 2024-02-06
fail ci with codecov error + workflow_dispatch

wheeheee:
- `fail_ci_if_error: true` => if the codecov upload errors, CI fails (probably due to secrets / codecov token)
- manually triggering CI is enabled with `workflow_dispatch`

Following up on #525.
On a side note, Windows on v1.6 is taking up a lot of time on `julia-buildpkg`.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -4,6 +4,7 @@ on:
     branches: [master]
     tags: ['*']
   pull_request:
+  workflow_dispatch:
 jobs:
   test:
     name: Julia ${{ matrix.version }} - ${{ matrix.os }} ${{ matrix.arch }}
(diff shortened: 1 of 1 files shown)


ViralBShah:
Yeah, not sure what to do about the 1.6 windows build being so slow. Hopefully we will have a new LTS very soon.

ViralBShah:
@dilumaluthge Any ideas why 1.6 on windows is so slow and is there any obvious quick thing we can do on the CI side?

DilumAluthge:
Hmmm. 6 minutes to download the cache, but then 6 more minutes in buildpkg? That's a bit odd. You could try commenting out the caching.

I'm also wondering why the General registry is getting cloned over Git; we should be using the Pkg server instead.

@ianbutterworth @omus any ideas on the caching?

IanButterworth:
I can't remember the behavior of using the pkgserver on 1.6 but it might not by default on windows? 

https://github.com/JuliaLang/Pkg.jl/pull/2377

PR #532, opened 2024-02-06
Bump the minimum required Julia version to 1.6

martinholters:
Ref. https://github.com/JuliaDSP/DSP.jl/pull/524#pullrequestreview-1862047815

I don't think we have to worry anymore about versions older than 1.6 (current LTS released almost three years ago).


ViralBShah:
Can we remove some of the old versions in OffsetArrays, Polynomials,  and SpecialFunctions?

ViralBShah (commented):
Can we remove some of the old versions in FFTW, OffsetArrays, Polynomials, Reexport, and SpecialFunctions?

I just pushed a commit to see if pushing aggressively will work with 1.6.

ViralBShah:
We may want to do this after #458 merges so that we can adopt these new deps from 0.8-dev.

ViralBShah:
Agree we do not need to wait for the other PR to merge for the version bump. This seems like it is good to merge now.

ViralBShah:
Merging this for now. We can always tweak as we go along.

PR #534, opened 2024-02-06
add type as another possible argument to `DF2TFilter`

### LINKED ISSUE #371: InexactError with stateful filter and complex data
Make a stateful filter object and use it on complex data:

`s  = rand(30) + im*rand(30)`
`df = digitalfilter(Lowpass(0.25), Butterworth(4))`
`f  = DF2TFilter(df)`
`filt(f,s)`

This errors out with `InexactError` at `complex.jl:37`. Real-valued data works just fine. Complex-valued data works fine with a stateless filter `filt(df,s)`.

Any advice on fixing? I'm new to Julia, so this might be a newbie-issue.

wheeheee:
Provides additional syntax that hopefully fixes #371

### DIFF
--- a/src/Filters/filt.jl
+++ b/src/Filters/filt.jl
@@ -215,9 +215,19 @@ function filt!(out::AbstractVector, f::DF2TFilter{<:Biquad,<:Vector}, x::Abstrac
     out
 end
 
-# Variant that allocates the output
+"""
+    filt(f::DF2TFilter{<:FilterCoefficients{:z},<:Array{T}}, x::AbstractVector{V}) where {T,V}
+
+Apply the [stateful filter](@ref stateful-filter-objects) `f` on `x`.
+
+!!! warning
+    The output array has eltype `promote_type(T, V)`, where
+    `T` is the eltype of the filter state.\n
+    For more control over the output type, provide a preallocated
+    output array `out` to `filt!(out, f, x)`.
+"""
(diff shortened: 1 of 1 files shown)


src/Filters/filt.jl
```diff
@@ -204,11 +216,13 @@ function filt!(out::AbstractVector, f::DF2TFilter{<:Biquad,<:Vector}, x::Abstrac
 end
 
 # Variant that allocates the output
-filt(f::DF2TFilter{<:FilterCoefficients{:z},<:Array{T}}, x::AbstractVector) where {T} =
-    filt!(Vector{T}(undef, length(x)), f, x)
+filt(f::DF2TFilter{<:FilterCoefficients{:z},<:Array{T}}, x::AbstractVector{V}) where {T,V} =
+    filt!(Vector{promote_type(T,V)}(undef, length(x)), f, x)
```
martinholters, line 220: Is there a reasonable use case for `promote_type(T,V) != T`? That would mean we cannot store the input samples in the state, at least not without loss of precision, right? Or am I missing something?

martinholters: Yes, but storing samples of a `Float64` signal as `Float32` is unusual to begin with. Whatever we do here, it may come as a surprise in one way or another, I guess. Would an identity filter convert the input to `Float32` and back to `Float64` or pass it through as is? Seems one would expect the latter. Is that true for the current implementations? But then a unit delay would suddenly cause loss of precision, which might also be a bit of a surprise.

So... I don't know. Your call unless anyone else wants to chime in.

PR #539, opened 2024-02-18
Fix bounds errors when resampling with some arbitrary ratios

### LINKED ISSUE #317: Got "BoundsError" while trying to use `resample` function.
```
using DSP

a_sig = sin.([1:1:35546; ])
resample(a_sig, 1/55.55)
```

And we will get:
```
ERROR: BoundsError: attempt to access 640-element Array{Float64,1} at index [641]
Stacktrace:
 [1] setindex! at .\array.jl:767 [inlined]
 [2] filt!(::Array{Float64,1}, ::FIRFilter{DSP.Filters.FIRArbitrary{Float64}}, ::Array{Float64,1}) at C:\Users\Losses\.juliapro\JuliaPro_v1.1.1.1\packages\DSP\wwKNu\src\Filters\stream_filt.jl:660
 [3] filt(::FIRFilter{DSP.Filters.FIRArbitrary{Float64}}, ::Array{Float64,1}) at C:\Users\Losses\.juliapro\JuliaPro_v1.1.1.1\packages\DSP\wwKNu\src\Filters\stream_filt.jl:673
 [4] resample(::Array{Float64,1}, ::Float64, ::Array{Float64,1}) at C:\Users\Losses\.juliapro\JuliaPro_v1.1.1.1\packages\DSP\wwKNu\src\Filters\stream_filt.jl:728
 [5] resample(::Array{Float64,1}, ::Float64) at C:\Users\Losses\.juliapro\JuliaPro_v1.1.1.1\packages\DSP\wwKNu\src\Filters\stream_filt.jl:733
 [6] top-level scope at none:0
```

anowacki:
In some cases when `filt(::FIRFilter{FIRArbitrary}, x)` is called
with certain values of `x`, the buffer is actually one sample too
short and a `BoundsError` is thrown in
`filt!(buffer, ::FIRFilter{FIRArbitrary}, x)`.  Add one extra sample
to the calculated buffer length to catch these exceptional cases, which
mostly arise when resampling with particular resampling ratios.

(This code is really a hack and simply works around the deeper problem
of ~calculating the correct output buffer length~ writing the correct number of points in `filt!`; this should instead be
addressed properly in a future change.)

Closes #317.

### DIFF
--- a/src/Filters/stream_filt.jl
+++ b/src/Filters/stream_filt.jl
@@ -630,6 +630,19 @@ end
 
 function filt(self::FIRFilter{Tk}, x::AbstractVector{Tx}) where {Th,Tx,Tk<:FIRKernel{Th}}
     bufLen         = outputlength(self, length(x))
+    # In some cases when `filt(::FIRFilter{FIRArbitrary}, x)` is called
+    # with certain values of `x`, `filt!(buffer, ::FIRFilter{FIRArbitrary}, x)`
+    # tries to write one sample too many to the buffer and a `BoundsError`
+    # is thrown.  Add one extra sample to catch these exceptional cases.
+    #
+    # See https://github.com/JuliaDSP/DSP.jl/issues/317
+    #
+    # FIXME: Remove this if and when the code in
+    #        `filt!(buffer, ::FIRFilter{FIRArbitrary}, x)`
+    #        is updated to properly account for pathological arbitrary rates.
+    if Tk <: FIRArbitrary
+        bufLen += 1
+    end
     buffer         = Vector{promote_type(Th,Tx)}(undef, bufLen)
     samplesWritten = filt!(buffer, self, x)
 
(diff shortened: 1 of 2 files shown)


martinholters (approved):
This is an ugly hack, but the situation is certainly better than before...

anowacki:
The following isn't quite correct:

> (This code is really a hack and simply works around the deeper problem
of calculating the correct output buffer length; this should instead be
addressed properly in a future change.)

The output buffer is the correct length; it is the writing to the buffer in the `filt!` function which is at fault, as it occasionally writes one point too many in the stream filtering routines.  So I have updated the comment and commit message to better reflect that.

ViralBShah:
@anowacki Would it be helpful for you to have commit access with the repo to help with maintenance? Not a whole lot changes workflow wise (PRs, reviews, etc), but we have more hands to keep things going and to improve the package.

ViralBShah:
Thanks. Happy to add you later if it makes sense. Just let me know.

ViralBShah:
Merge?

tknopp:
I ran into this issue and found that it is fixed on the main branch. @martinholters would it be possible to trigger a release including this fix?

PR #541, opened 2024-02-20
add inbounds to src/Filters/design.jl

Jafagervik:
Avoid unnecessary boundary checking


Jafagervik:
Yes. I can try to profile where the compiler turns of blundary checking, but this macro is indeed powerful, and since I'm already using my own version of DSP.jl for my master thesis on ai and signal processing in julia, I thought I'd might as well contribute! I will do some more benches as soon as possible, but I'm quite busy these days

ViralBShah:
Thank you for the contribution, and good to know that you are already using this in your code. 

A quick benchmark in this PR will help serve as documentation for the change. Generally, the preference is not to add `@inbounds` because people have used it incorrectly and aggressively in the past. But if the benefit is clear, it is definitely worthwhile to use it.

ViralBShah:
@wheeheee The windows failures seem codecov upload problems. I think the change to fail on codecov errors is perhaps aggressive. Would reverting it fix the error here (which is not in the package at all)?

martinholters:
FWIW, the `@inbounds` annotations here do look correct to me, but if the benchmarks don't support that there is a performance improvement, I don't think we should be doing this. It will increase the mental load and the risk of introducing bugs in the future.

Jafagervik:
> Actually, I think there's still a potential missed optimisation here. It would probably be fine to not zero out the `Vector`s in Butterworth etc., which could be helpful to @Jafagervik

This would help me a lot at least. I'm on julia 1.10.1 right now and I use Butterworth quite a lot

ViralBShah:
Close (if this is included in the cleanups)?

PR #542, opened 2024-02-20
use `std` to test lpc error

wheeheee:
The lpc error was previously compared to 1, but the `std` of the `randn` generated signal is a better choice.


martinholters (approved):
Sure. Of course, `std(s)` will be quite close to one, but why risk spurious test failures.

PR #551, opened 2024-02-28
Reduce CI load

ViralBShah:
Use includes rather than excludes for the CI script. 
Runs fewer combinations.


ViralBShah:
Apologies - I forgot squashing here.

Issue #548, opened 2024-02-27
Stream data to digital filter / live IIR filter


scls19fr:
Hello,

Following discussion started at https://discourse.julialang.org/t/scipy-signal-iirfilter-equivalent-in-dsp-jl-and-more/110783/4 I wonder if DSP.jl provides structure of digital filter which can be used for streaming input data ie processing live data and not a vector of precalculated values ?
Ie is there some kind of filter which preserve its state each time a new input is coming.

I did this [Pluto.jl](https://github.com/fonsp/Pluto.jl) notebook https://gist.github.com/scls19fr/0ae16d92ca39d3eb9c42cc0fc618c723 for experimenting this kind of problem (both for filtering a vector of input signal and for filtering streaming signal thanks to [PlutoUI Clock](https://featured.plutojl.org/basic/plutoui.jl#f31668c6-c768-11ea-1501-5f41afa7c83b)).

I end up to convert idea mentioned https://www.samproell.io/posts/yarppg/yarppg-live-digital-filter/ with Julia.

[OnlineStatsBase](https://github.com/joshday/OnlineStatsBase.jl) API looks interesting for this kind of work.

So I did

```julia
	abstract type AbstractLiveFilter{T} <: OnlineStat{T} end

	mutable struct LiveDigitalFilter{T} <: AbstractLiveFilter{T}
		value::T
    	n::Int
		
		b::Vector
		a::Vector
	
		_xs::CircularBuffer
		_ys::CircularBuffer

		function LiveDigitalFilter{T}(pr::PolynomialRatio) where {T}
			b = reverse(pr.b.coeffs)
			a = reverse(pr.a.coeffs)
			_xs = CircularBuffer{T}(length(b))
			fill!(_xs, zero(T))
			_ys = CircularBuffer{T}(length(a) - 1)
			fill!(_ys, zero(T))
			new{T}(0, 0, b, a, _xs, _ys)
		end
	end

	function reset!(f::LiveDigitalFilter)
		f.n = 0
		empty!(f._xs)
		empty!(f._ys)
		fill!(f._xs, zero(eltype(f.b)))
		fill!(f._ys, zero(eltype(f.a)))	
	end
	
	function OnlineStatsBase._fit!(f::LiveDigitalFilter, x)
	    f.n += 1
		pushfirst!(f._xs, x)
		y = sum(f.b .* f._xs) - sum(f.a[2:end] .* f._ys)
		y = y / f.a[1]
[...]

Issue #367, opened 2020-06-08
filtfilt with second order sections


gnadt:
Attempting to use second order sections to filter measurements. Getting a mul! error.

sos = SecondOrderSections{Float64,Array{Float64,1}}
   biquads –> Vector{Biquad{Float64}} with 21 elements
   g –> Vector{Float64} with 21 elements
meas = Vector{Float64} with 15500 elements

filtfilt(sos,meas)

ERROR: DimensionMismatch("result C has dimensions (2, 21), needs (2,1)")
Stacktrace:
 [1] _generic_matmatmul!(::Array{Any,2}, ::Char, ::Char, ::Array{Float64,2}, ::Array{Float64,1}, ::LinearAlgebra.MulAddMul{true,true,Bool,Bool}) at /Users/vagrant/worker/juliapro-release-osx1011-0_6/build/tmp_julia/Julia-1.4.app/Contents/Resources/julia/share/julia/stdlib/v1.4/LinearAlgebra/src/matmul.jl:739
 [2] generic_matmatmul!(::Array{Any,2}, ::Char, ::Char, ::Array{Float64,2}, ::Array{Float64,1}, ::LinearAlgebra.MulAddMul{true,true,Bool,Bool}) at /Users/vagrant/worker/juliapro-release-osx1011-0_6/build/tmp_julia/Julia-1.4.app/Contents/Resources/julia/share/julia/stdlib/v1.4/LinearAlgebra/src/matmul.jl:727
 [3] mul!(::Array{Any,2}, ::Array{Float64,2}, ::Array{Float64,1}, ::Bool, ::Bool) at /Users/vagrant/worker/juliapro-release-osx1011-0_6/build/tmp_julia/Julia-1.4.app/Contents/Resources/julia/share/julia/stdlib/v1.4/LinearAlgebra/src/matmul.jl:235
 [4] mul! at /Users/vagrant/worker/juliapro-release-osx1011-0_6/build/tmp_julia/Julia-1.4.app/Contents/Resources/julia/share/julia/stdlib/v1.4/LinearAlgebra/src/matmul.jl:208 [inlined]
 [5] filtfilt(::SecondOrderSections{Float64,Array{Float64,1}}, ::Array{Float64,1}) at /Users/gnadt/.juliapro/JuliaPro_v1.4.0-1/packages/DSP/KGj8O/src/Filters/filt.jl:325
 [6] top-level scope at none:0

martinholters:
[...]

PR #552, opened 2024-03-01
Annotate docstring arguments with types in Filters

wheeheee:
Addresses some concerns in #548 and implements the type restriction in #367.

### DIFF
--- a/src/dspbase.jl
+++ b/src/dspbase.jl
@@ -5,7 +5,10 @@ const SMALL_FILT_CUTOFF = 58
 _zerosi(b,a,T) = zeros(promote_type(eltype(b), eltype(a), T), max(length(a), length(b))-1)
 
 """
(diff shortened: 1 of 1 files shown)


src/dspbase.jl
```diff
@@ -5,10 +5,12 @@ const SMALL_FILT_CUTOFF = 58
 _zerosi(b,a,T) = zeros(promote_type(eltype(b), eltype(a), T), max(length(a), length(b))-1)
 
 """
-    filt(b, a, x, [si])
+    filt(b::Union{AbstractVector,Number}, a::Union{AbstractVector,Number}, x, [si])
```
scls19fr, line 8: x is AbstractVector

ViralBShah:
Merge?

ViralBShah:
For obvious PRs, there is the button on "Merge without waiting for requirements to be met". I would recommend using it for various docs and maintenance PRs.

PR #553, opened 2024-03-07
Don't fail if codecov commit fails

ViralBShah:


ViralBShah:
> i guess this should be fine, but is there some other way to get a warning if codecov keeps failing?

I was thinking that if it fails once every few PRs, maybe it's not worth worrying about?

Issue #416, opened 2021-03-31
[DOC] Inaccuracy in the Esprit Algorithm Documentation Code


RoyiAvital:
The current code and documentation for the [`esprit()` function](https://github.com/JuliaDSP/DSP.jl/blob/4cf717503e856a11b551ed3e463589dd5889f6d1/src/estimation.jl#L7-L30) has some inaccuracies.

While the documentation states it uses the model of `p` sines (See `Given length N signal "x" that is the sum of p sinusoids of unknown frequencies, estimate and return an array of the p frequencies.`) in reality the Esprit algorithm assumes `p` complex exponents.

The difference is crucial as in the case of estimating the frequencies of a set of `p` real sines the algorithm will yield `p` frequencies yet in duals of positive and negative frequencies (While is the data is complex exponents the `p` frequencies will match the `p` frequencies of the model).

So documentation should clarify the model is about complex exponent and in case of estimating real signal (sine / cosine) the input for frequencies should be `2p` and then filter the negative frequencies).

galenlynch:
PR welcome

PR #554, opened 2024-03-28
Update ESPRIT documentation.

kcin96:
This PR addresses issue #416 and adds examples to the documentation.

### DIFF
--- a/src/estimation.jl
+++ b/src/estimation.jl
@@ -12,26 +12,57 @@ export esprit, jacobsen, quinn
 ESPRIT [^Roy1986] algorithm for frequency estimation.
 Estimation of Signal Parameters via Rotational Invariance Techniques
 
-Given length N signal "x" that is the sum of p sinusoids of unknown frequencies,
-estimate and return an array of the p frequencies.
+Given length N signal `x` that is the sum of `p` complex exponentials (cisoids) of unknown frequencies,
+estimate and return an array of the `p` frequencies.
+
+```math
+x(t)=\\sum_{i=1}^p A_ie^{j2\\pi f_i t}
+```
+where ``A_i`` is the amplitude of the ``i``th cisoid, ``f_i`` is the ``i``th frequency.
 
 # Arguments
 - `x::AbstractArray`: complex length N signal array
 - `M::Integer`: size of correlation matrix, must be <= N.
       The signal subspace is computed from the SVD of an M x (N-M+1) signal matrix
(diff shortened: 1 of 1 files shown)


src/estimation.jl
```diff
@@ -12,26 +12,57 @@ export esprit, jacobsen, quinn
...
 
-Given length N signal "x" that is the sum of p sinusoids of unknown frequencies,
-estimate and return an array of the p frequencies.
+Given length N signal `x` that is the sum of `p` complex exponentials (cisoids) of unknown frequencies,
+estimate and return an array of the `p` frequencies.
+
+```math
+x(t)=\\sum_{i=1}^p A_ie^{j2\\pi f_i t}
+```
+where ``A_i`` is the amplitude of the ``i``th cisoid, ``f_i`` is the ``i``th frequency.
```
mbaz, line 21: Suggestion:

    where ``A_i`` and ``f_i`` are the amplitude and frequency of the ``i``th cisoid.

src/estimation.jl
```diff
@@ -12,26 +12,57 @@ export esprit, jacobsen, quinn
 ESPRIT [^Roy1986] algorithm for frequency estimation.
 Estimation of Signal Parameters via Rotational Invariance Techniques
 
-Given length N signal "x" that is the sum of p sinusoids of unknown frequencies,
-estimate and return an array of the p frequencies.
+Given length N signal `x` that is the sum of `p` complex exponentials (cisoids) of unknown frequencies,
```
mbaz, line 15: Does ESPRIT make any assumptions about the phase of the cisoids? If so, it may be worth it to mention them here.

src/estimation.jl
kcin96: I'll update that line. Thanks!

src/estimation.jl
kcin96: Yes, good point. How about
```suggestion
Given a signal `x` of length N that is the sum of `p` complex exponentials (cisoids) with unknown constant amplitudes and phases,
```

src/estimation.jl
kcin96, line 21:
```suggestion
where ``A_i`` and ``f_i`` are the amplitude and frequency of the ``i``th cisoid.
```

src/estimation.jl
mbaz: Sounds good!

kcin96:
Thanks for review @mbaz @wheeheee!

Issue #480, opened 2022-09-14
Power, or power spectral density of Periodogram?


KronosTheLate:
Currently, there is a function `power` that returns the field `power` of a Periodogram.

However, from what I can tell, the returned values are power spectral _densities_, which is of a different unit.
Am I correct in that understanding? If so, this should be reflected in the docstrings for `Periodogram`, `periodogram`, and 'power`.

kcin96:
Yes, I agree it is power spectral density. I'll update the documentation.

KronosTheLate:
Awesome, thanks for taking this issue seriously and fixing it for me ^_^

PR #557, opened 2024-04-12
Update periodogram documentation.

kcin96:
This PR 
1. Update docs to mention that power spectral densities are computed for periodograms as per #480 
2. Update periodogram docs with examples.
3. Fixes welch_pgram! so Floats can work as well as Ints. 
4. Add test to welch_pgram!


kcin96:
That is a good point! So something like this which checks that the type of input `s` matches that of `config.inbuf` 
```julia
function welch_pgram!(out::AbstractVector, s::AbstractVector, config::WelchConfig{T}) where T<:Number
     if length(out) != length(config.freq)
         throw(DimensionMismatch("""Expected `output` to be of length `length(config.freq)`;
             got `length(output)` = $(length(out)) and `length(config.freq)` = $(length(config.freq))"""))
    elseif eltype(out) != fftabs2type(T)
        throw(ArgumentError("Eltype of output ($(eltype(out))) doesn't match the expected "*
                            "type: $(fftabs2type(T))."))
    elseif eltype(s) != eltype(config.inbuf)
        throw(ArgumentError("Eltype of input `s` ($(eltype(s))) doesn't match the expected `config` "*
                            "type: $(eltype(config.inbuf))."))
end
```

kcin96:
I see, you're right, and it is much cleaner now.

PR #561, opened 2024-06-21
[DOCS] Add Small Note about Method Validations

TheCedarPrince:
Hey there!

After a discussion with @ericphanson , I learned that DSP.jl is actually validated against MATLAB and MNE-Python! In my eyes, that made using DSP.jl much more salient to me in knowing that, although its methods are made in Julia from scratch, they are compared with suites that are considered state of the art in many domains. I know it is a small edit, but I thought it was worthwhile enough to update the docs here because if I think that way, it may appeal to other folks thinking similarly.

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -1,6 +1,6 @@
 # Welcome to DSP.jl's documentation!
 
-DSP.jl provides a number of common [digital signal processing](https://en.wikipedia.org/wiki/Digital_signal_processing) routines in Julia. So far, the following submodules are implemented:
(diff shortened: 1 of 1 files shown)


martinholters:
> DSP.jl is actually validated against MATLAB and MNE-Python

I think that is only true for part of DSP.jl.

ericphanson:
Here is the context btw: https://discourse.julialang.org/t/decomposing-an-eeg-signal-from-scratch/115904/6

TheCedarPrince:
That's very reasonable @martinholters -- I updated my PR to reflect that point.

ViralBShah (approved):

ViralBShah:
I'm merging for now, but it can easily be fine tuned later.

PR #564, opened 2024-09-10
Update CI.yml to use pre-release builds

ViralBShah:
Using `pre` instead of `nightly`. Thoughts?


martinholters:
I'm on board with `nightly` -> `pre`. But the `1.6` -> `lts` change seems problematic to me, as that means we will eventually stop testing on anything older than 1.10, while still claiming compatibility with 1.6 onward. Better manually bump the lowest version in `Project.toml` and for CI in lockstep IMHO.

PR #565, opened 2024-09-13
[release-0.7] Update CI.yml to use pre-release builds

martinholters:
Same as #564 but for v0.7


martinholters:
Ah, apparently requires #555, so backport that one, too.

PR #516, opened 2023-12-05
`filt` perf optimizations

wheeheee:
- instead of allocating for every column, reuse `si` (different from input `si`)
- `filt` currently errors for `N>2` (edit: if the trailing dimensions aren't all of length 1); this PR doesn't change that

PR #566, opened 2024-09-13
Backport `filt` fix (#516 and #539)

martinholters:
Backports #539 and, as a prerequisite, #516. (A more targeted subset of the latter might be in order, but including it wholesale was simplest to do.)


martinholters:
Tests fail on nightly, but that's unrelated, pre-exisitng on v0.7.9, and should be "solved" with #565.

martinholters:
Ah, that's a bit unfortunate, but with 1.10 becoming LTS soonish (fingers crossed), I wouldn't worry too much. That is, if anyone steps up to backport the bugfix of #539 without the performance changes of #516, that would be welcome, otherwise I'll merge this one in a couple of days.

martinholters:
I was hoping for something which makes the backport simpler, not tack even more stuff on which hasn't even landed in master yet and might itself warrant some detailed review/discussion. If you open a PR, we discuss, backport (hoping it backports cleanly) ... that will take quite a while and the point is to do a timely release of v0.7.10 to get the bugfix out.

So, I'll go ahead with this as is. If you PR your tweaks we can always think about backporting them and doing a v0.7.11, but finally cutting v0.8.0 would also be quite something...

PR #569, opened 2024-09-30
Optimized `_filt_fir!` for `DF2TFilter`

wheeheee:
Also adjusts cutoff, uses `LazyString` for errors.

### DIFF
--- a/src/Filters/filt.jl
+++ b/src/Filters/filt.jl
@@ -185,7 +185,7 @@ function filt!(out::AbstractVector, f::DF2TFilter{<:PolynomialRatio,<:Vector}, x
             end
             _filt_iir!(out, b, a, x, si, 1)
         elseif n <= SMALL_FILT_CUTOFF
-            vtup = ntuple(j -> @inbounds(b[j]), Val(n))
+            vtup = ntuple(j -> b[j], Val(length(b)))
             si .= getfield.(_filt_fir!(out, vtup, x, si, Val(:DF2)), :value)
         else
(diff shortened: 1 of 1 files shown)


src/Filters/filt.jl
```diff
@@ -171,24 +171,24 @@ function filt!(out::AbstractVector, f::DF2TFilter{<:PolynomialRatio,<:Vector}, x
...
+            if as < n
+                append!(a, zero(eltype(a)) for _ in 1:(n-as))
+            elseif bs < n
+                append!(b, zero(eltype(b)) for _ in 1:(n-bs))
             end
-            si[n-1] = muladd(b[n], xi, -a[n] * val)
-            out[i] = val
+            _filt_iir!(out, b, a, x, si, 1)
+        elseif n <= SMALL_FILT_CUTOFF
+            vtup = ntuple(j -> @inbounds(b[j]), Val(n))
```
martinholters, line 188: `length(b) == n` by a structural invariant of `DF2TFilter`, right? Then
```suggestion
            vtup = ntuple(j -> b[j], Val(length(b))
```
should allow elision of the bounds check and looks a bit safer.

martinholters (approved):
I don't understand the `SMALL_FILT_VECT_CUTOFF` dance, but apart from that LGTM.

martinholters:
Ah, so the `@inbounds` is semantically correct for any `N`, but whether it helps or hinders performance depends on `N`? That's... ugh...

Anyway, from my point of view, this is good to go.

PR #570, opened 2024-10-09
Update macOS CI

wheeheee:
Since macOS-latest now points to macos-14.
Also, the macOS-latest x64 job currently emulates x64 on aarch64 (actions/runner-images#9741). Should this be macos-13 x64 instead?


martinholters:
Hm... using `macos-13` on `x64` and `macos-14` on `aarch64` does sound reasonable.

martinholters (approved):

PR #571, opened 2024-10-19
Fix `_polyprep`, `PolynomialRatio`, for `Number` args

wheeheee:
The current implementation uses `reverse(x)`, which is a method error for `Number` arguments. Thus `PolynomialRatio(::Number, ::Number)` doesn't work as currently documented. ~This commit also reduces runtime dispatch.~ edit: guess it doesn't because of all the inlining anyway
2nd edit: it wasn't, but now the type instabilities have been removed by manually inlining `_polyprep`

### DIFF
--- a/src/Filters/coefficients.jl
+++ b/src/Filters/coefficients.jl
@@ -56,9 +56,9 @@ end
 # Transfer function form
 #
 
-function shiftpoly(p::LaurentPolynomial{T}, i::Integer) where {T<:Number}
+function shiftpoly(p::LaurentPolynomial{T,D}, i::Integer) where {T<:Number,D}
     if !iszero(i)
-        return Compat.@inline p * LaurentPolynomial([one(T)], i, indeterminate(p))
+        return p * LaurentPolynomial{T,D}([one(T)], i)
     end
     return p
 end
@@ -129,7 +129,7 @@ PolynomialRatio{:s}(b::PolynomialRatioArgTs{T1}, a::PolynomialRatioArgTs{T2}) wh
     PolynomialRatio{:s,promote_type(T1, T2)}(b, a)
 
 """
-    _polyprep(D::Symbol, x::Union{T,Vector{T}}, ::Type=T) where {T<:Number}
+    _polyprep(D::Symbol, x::Union{T,Vector{T}}, ::Type) where {T<:Number}
 
 Converts `x` to polynomial form. If x is a `Number`, it has to be converted into
 a `Vector`, otherwise `LaurentPolynomial` dispatch goes into stack overflow
@@ -140,7 +140,7 @@ trying to collect a 0-d array into a `Vector`.
     The DSP convention for Laplace domain is highest power first.\n
     The Polynomials.jl convention is lowest power first.
 """
-@inline _polyprep(D::Symbol, x::Union{T,Vector{T}}, ::Type{V}=T) where {T<:Number,V} =
+@inline _polyprep(D::Symbol, x::Union{T,Vector{T}}, ::Type{V}) where {T<:Number,V} =
     LaurentPolynomial{V,D}(x isa Vector ? reverse(x) : [x], D === :z ? -length(x) + 1 : 0)
 
 function PolynomialRatio{:z,T}(b::Union{T1,Vector{<:T1}}, a::Union{T2,Vector{<:T2}}) where {T<:Number,T1<:Number,T2<:Number}
--- a/src/Filters/coefficients.jl
+++ b/src/Filters/coefficients.jl
@@ -143,7 +143,7 @@ trying to collect a 0-d array into a `Vector`.
 @inline _polyprep(D::Symbol, x::Union{T,Vector{T}}, ::Type{V}) where {T<:Number,V} =
     LaurentPolynomial{V,D}(x isa Vector ? reverse(x) : [x], D === :z ? -length(x) + 1 : 0)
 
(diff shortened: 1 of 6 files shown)


src/Filters/coefficients.jl
```diff
@@ -122,20 +122,43 @@ PolynomialRatio{:s, Int64}(LaurentPolynomial(3 + 2*s + s²), LaurentPolynomial(4
...
-    PolynomialRatio{D,T}(_polyprep(D, b, T), _polyprep(D, a, T))
-PolynomialRatio{D}(b::Union{Number,Vector{<:Number}}, a::Union{Number,Vector{<:Number}}) where {D} =
-    PolynomialRatio{D}(_polyprep(D, b), _polyprep(D, a))
+    The DSP convention for Laplace domain is highest power first.\n
+    The Polynomials.jl convention is lowest power first.
+"""
+@inline _polyprep(D::Symbol, x::Union{T,Vector{T}}, ::Type{V}) where {T<:Number,V} =
+    LaurentPolynomial{V,D}(x isa Vector ? reverse(x) : [x], D === :z ? -length(x) + 1 : 0)
+
+function PolynomialRatio{:z,T}(b::Union{T1,Vector{<:T1}}, a::Union{T2,Vector{<:T2}}) where {T<:Number,T1<:Number,T2<:Number}
```
martinholters, line 146:
```suggestion
function PolynomialRatio{:z,T}(b::Union{T1,Vector{T1}}, a::Union{T2,Vector{T2}}) where {T<:Number,T1<:Number,T2<:Number}
```
Or am I missing something?

src/Filters/coefficients.jl
```diff
@@ -122,20 +122,43 @@ PolynomialRatio{:s, Int64}(LaurentPolynomial(3 + 2*s + s²), LaurentPolynomial(4
...
+        throw(ArgumentError("filter must have non-zero leading denominator coefficient"))
+    end
+    Tn = typeof(one(T1) / one(T2))
+    if !isone(a[1])
+        bn = b / a[1]
+        an = a / convert(Tn, a[1])
+    else    # for type stability
+        bn = T1 === Tn ? b : convert.(Tn, b)
+        an = T2 === Tn ? a : convert.(Tn, a)
+    end
```
martinholters, line 157: How much does that gain us compared to
```suggestion
    bn = b / a[1]
    an = a / a[1]
```
? And if it is worthwhile, do we really need to go via `Tn` here, given that the result will be converted to `T` then anyway? And why is this better than leaving the normalization to the inner constructor?

martinholters (commented):
Mostly looks good to me, but I'd like to see some justification for why 
`PolynomialRatio{:z,T}(b::Union{T1,Vector{<:T1}}, a::Union{T2,Vector{<:T2}}) where {T<:Number,T1<:Number,T2<:Number}`has to be some complicated.

src/Filters/coefficients.jl
martinholters: Hm, if we want to reduce the number of allocations, the code should maybe be restructured even more. E.g. the `reverse` in `_polyprep` could often (but not always!) be a `reverse!`. It would probably be best to first convert to a `Vector{T}` and then operate in-place. But a) I'm not sure it's worth it -- is creation of `PolynomialRatio` ever performance critical? -- and b) that is probably more than a "Fix ..." PR should be doing.

src/Filters/coefficients.jl
```diff
@@ -122,20 +122,37 @@ PolynomialRatio{:s, Int64}(LaurentPolynomial(3 + 2*s + s²), LaurentPolynomial(4
...
-    PolynomialRatio{D}(_polyprep(D, b), _polyprep(D, a))
+!!! warning
+
+    The DSP convention for Laplace domain is highest power first.\n
+    The Polynomials.jl convention is lowest power first.
+"""
+@inline _polyprep(D::Symbol, x::Union{T,Vector{T}}, ::Type{V}) where {T<:Number,V} =
+    LaurentPolynomial{V,D}(x isa Vector ? reverse(x) : [x], D === :z ? -length(x) + 1 : 0)
+
+function PolynomialRatio{:z,T}(b::Union{T1,Vector{T1}}, a::Union{T2,Vector{T2}}) where {T<:Number,T1<:Number,T2<:Number}
```
martinholters, line 146:
```suggestion
function PolynomialRatio{:z,T}(b::Union{Number,Vector{<:Number}}, a::Union{Number,Vector{<:Number}}) where {T<:Number}
```

src/Filters/coefficients.jl
```diff
@@ -122,20 +122,37 @@ PolynomialRatio{:s, Int64}(LaurentPolynomial(3 + 2*s + s²), LaurentPolynomial(4
...
+    if isempty(a) || iszero(a[1])
+        throw(ArgumentError("filter must have non-zero leading denominator coefficient"))
+    end
+    bn = b / a[1]
+    an = a / a[1]
+    return PolynomialRatio{:z,T}(_polyprep(:z, bn, T), _polyprep(:z, an, T))
```
martinholters, line 152:
```suggestion
    return PolynomialRatio{:z,T}(_polyprep(:z, b / a[1], T), _polyprep(:z, a / a[1], T))
```

martinholters (approved):
LGTM modulo minor style nits. Thanks!

src/Filters/coefficients.jl
martinholters: Matter of taste, I guess.

> aside, it's weird that the repeated `a[1]`s aren't elided in code_llvm

True, but not that it matters much, here. Is the bounds check elided?

PR #574, opened 2024-10-27
Allow different argument eltypes for `filt_stepstate` (fix #573)

### LINKED ISSUE #573: Missing Documentation for the Function `filtfilt(b::AbstractVector, a::AbstractVector, x::AbstractArray)`
The documentation only expose `filtfilt(b::AbstractVector, x::AbstractArray)`.
Yet [`filtfilt(b::AbstractVector, a::AbstractVector, x::AbstractArray)` is available](https://github.com/JuliaDSP/DSP.jl/blob/c6663ab5ddd5586da63912548f2dbf3a164e4b49/src/Filters/filt.jl#L277C10-L277C55) but not documented.

Could it be documented?

OK, Tried using it and it generates errors:

```julia
ERROR: MethodError: no method matching filt_stepstate(::Vector{Float64}, ::Vector{Int64})
```

wheeheee:


martinholters (approved):
A test for the `isone(a[1]) && T !== V` case would be good. Or remove that optimization and do the scaling unconditionally. Run-time will be dwarfed by the linear solver for computation of `si` anyway, I guess.

But I don't think this fixes #573 as that issue is about missing documentation which this PR doesn't add.

martinholters:
> IIUC, it's just repeating the two-argument version for each column of an array?

I think the issue is that we only have `filtfilt(coef::FilterCoefficients, x::AbstractArray)` documented, not `filtfilt(b, a, x)`.

src/Filters/filt.jl
```diff
@@ -263,36 +263,39 @@ end
...
 
 """
     filtfilt(coef::FilterCoefficients, x::AbstractArray)
+    filtfilt(b::AbstractVector, x::AbstractArray)
+    filtfilt(b::AbstractVector, a::AbstractVector, x::AbstractArray)
 
-Filter `x` in the forward and reverse directions using filter
-coefficients `coef`. The initial state of the filter is computed so
+Filter `x` in the forward and reverse directions using either a
+`FilterCoefficients` object `coef`, or the filters `b` and / or `a`
```
martinholters, line 293:
```suggestion
`FilterCoefficients` object `coef`, or the coefficients `b` and optionally `a`
```
?

src/Filters/filt.jl
```diff
@@ -263,36 +263,39 @@ end
...
+as in [`filt`](@ref). The initial state of the filter is computed so
 that its response to a step function is steady state. Before
 filtering, the data is extrapolated at both ends with an
 odd-symmetric extension of length
-`3*(max(length(b), length(a))-1)`.
+`min(3*(max(length(b), length(a))-1), size(x, 1) - 1)`
```
martinholters, line 298:
```suggestion
`min(3*(max(length(b), length(a))-1), size(x, 1) - 1)`.
```

martinholters:
> Isn't that case covered in lines 381 to 383?
> edit: Oh, sorry, _that_ kind of test...

That needs a test where `a` and `b` have an eltype that change under division (e.g. both are `Vector{Int}`), while your latest test includes an `x::Vector{Int}` but still `b::Vector{Float64}`, so those lines are still not covered.

PR #575, opened 2024-10-28
Require Julia v1.10 or later

martinholters:
Not sure whether we want to do this now or wait a bit, but the DSP version bump to 0.8 feels like a good opportunity.
(Admittedly, my main motivation is to be able to use package extensions, ref #545.)


martinholters:
I'll leave this open for a few days in case anyone dares to object, but will merge if that's not the case.

Issue #224, opened 2018-08-25
Support conv(x,y,z)?


dlfivefifty:
I believe `conv` is associative: 
```julia
julia> x = randn(3); y= randn(4); z=randn(5); conv(conv(x,y),z) ≈ conv(x,conv(y,z))
true
```
so `conv(x,y,z)` is well-defined. Any reason not to support this? 

I have a small use for this in ApproxFun (determining the dimensions of polynomial degrees of tensor products of polynomial bases), so could add it as a PR.

rickhg12hs:
You could "extend" `DSP.conv` yourself with:

```julia
import DSP.conv
conv(x, y, z...) = conv(conv(x, y), z...)

```

dlfivefifty:
I know. That’s why it’s a “?”

rickhg12hs:
Ah, it's a policy question.  OK.

Also, technically speaking, if this were taken to the extreme, a more numerically stable and computationally efficient algorithm could be used than the convenience function I gave.

ssfrr:
One possible issue is that we really should get rid of `conv2` and instead just have `conv` do dispatch and work for N-D arrays (#202). In that case there might be a conflict between this proposal and [the 3-arg `conv2`](https://github.com/JuliaDSP/DSP.jl/blob/f3c9382ba995d6d5d2b06e517c77c9d5b99e0c82/src/dspbase.jl#L152-L164). I guess technically dispatch should be able to distinguish between these cases (`conv(::Vector, ::Vector, ::Vector)` vs `conv(::Vector, ::Vector, ::Matrix)`) but then you might get into a tricky spot where a Nx1 `Matrix` is treated substantially differently than a `Vector`.

I haven't thought about it much but wanted to raise a flag that this could be a potential issue.

martinholters:
The natural replacement for the 3-arg `conv2` would be a 3-arg `conv` taking a vector, a rowvector, and a matrix. That would also nicely disambiguate whether `u` or `v` of the 3-arg `conv2` is taken rowwise (which also isn't clarified by the docstring it seems).

ssfrr:
Yeah, I like that. Representing a rank-1 matrix by its column/row factors (and making that explicit) makes a lot of sense.

dlfivefifty:
Just to be clear: the syntax would be `conv(u, v’, A) `, right? And it’s consistent   because ` conv(u, v’) = u*v’ `

martinholters:
[...]

Issue #166, opened 2017-06-25
Extend `conv` to arbitrary lists of vectors 


KristofferC:
Moved from https://github.com/JuliaLang/julia/issues/16126

At this moment `conv` convolves only two vectors, `conv(u,v)`. If you have a list of vectors `lst` that you want to convolve, you can do it like this:

`reduce(conv, lst)`

But this is very inefficient because it computes the inverse Fourier transform too many times. A much more efficient way is to compute the FT of every vector once, multiply them element-wise, and then compute only one FT inversion of the product sequence. But for this you need a dedicated `conv(u, v, w, ...)` that accepts an arbitrary number of vectors.

martinholters:
Sounds like a good idea, but we have `conv(u::AbstractVector, v::AbstractVector, A::AbstractMatrix)` for a 2D-convolution with separable (into `u` and `v`) convolution kernel. And we support mixing arrays of different dimensions for the two-arg cases, so that would imply a subtle different meaning for that three-arg method if we support n-arg in general.
That said, we should probably deprecate the current three-arg method to `conv(u::AbstractVector, v::Transpose{<:AbstractVector}, A::AbstractMatrix)` which would just be a special case of the general n-arg method to be added in the future.

PR #577, opened 2024-11-05
Deprecate `conv(u, v::AbstractVector, A)` to `conv(u, v::Transpose, A)`

martinholters:
Prepare for a possible future multi-arg `conv` (cf. #338, #224, #166) by deprecating the current `conv(u::AbstractVector, v::AbstractVector, A::AbstractMatrix)` (which would have a conflicting meaning) to `conv(u::AbstractVector, v::Transpose{T,<:AbstractVector}, A::AbstractMatrix)` which is consistent with a future `conv(u, v, A) == conv(conv(u, v), A))`, just more effcient.

See https://github.com/JuliaDSP/DSP.jl/issues/224#issuecomment-416128217, https://github.com/JuliaDSP/DSP.jl/issues/166#issuecomment-2454896988.

Even without the general n-arg `conv`, I think this signature makes more sense as it makes it explicit which of `u` and `v` operates in which direction. (Which before wasn't even documented.)


src/dspbase.jl
```diff
@@ -738,19 +738,19 @@ end
...
 
 2-D convolution of the matrix `A` with the 2-D separable kernel generated by
-the vectors `u` and `v`.
+the vector `u` and row-vector `v`.
 Uses 2-D FFT algorithm.
 """
-function conv(u::AbstractVector{T}, v::AbstractVector{T}, A::AbstractMatrix{T}) where T
+function conv(u::AbstractVector{T}, v::Transpose{T,<:AbstractVector}, A::AbstractMatrix{T}) where T
     # Arbitrary indexing offsets not implemented
     @assert !Base.has_offset_axes(u, v, A)
```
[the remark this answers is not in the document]
martinholters: Unrelated, but in principle :+1:, although it should probably use `conv_with_offset` from #545 once that has landed.

src/dspbase.jl
```diff
@@ -787,19 +787,21 @@ end
     conv(u,v,A)
 
 2-D convolution of the matrix `A` with the 2-D separable kernel generated by
-the vectors `u` and `v`.
+the vector `u` and row-vector `v`.
 Uses 2-D FFT algorithm.
 """
-function conv(u::AbstractVector{T}, v::AbstractVector{T}, A::AbstractMatrix{T}) where T
+function conv(u::AbstractVector{T}, v::Transpose{T,<:AbstractVector}, A::AbstractMatrix{T}) where T
```
martinholters, line 793: This could also be `Adjoint` instead of `Transpose` (with no change to the implementation below). Would that make more sense? Would one rather separate a separable kernel `H` as `H=u*v'` or `H=u*transpose(v)`? (Anyone around working with complex-valued 2D-data and separable filters?)
Or should that even be `Union{Adjoint{T,<:AbstractVector}, Transpose{T,<:AbstractVector}}` (aka `LinearAlgebra.AdjOrTransAbsVec{T}`)?

martinholters: So I thought I'd ping the author of that method in case he has any insight on whether `Transpose` or `Adjoint` would be more idiomatic. Some `git blame`s later, I've arrived at https://github.com/JuliaLang/julia/commit/823cff9c0714be30a1afb63f412041a73087c08f where @JeffBezanson added this to `signal.j` almost 13 years ago. So Jeff, what do you say?

More seriously, the current code does `transpose`, so the deprecation for `conv(u, v, A)` could either be `conv(u, transpose(v), A)` or `conv(u, conj(v)', A)`, where I find the former more obvious. If we're going with that, it's either `Transpose` or `AdjOrTransAbsVec`, and I tend to prefer the simpler `Transpose` for now. We can always wider the signature later if the need arises.

Another issue that in light of #403, the argument order `conv(A, u, transpose(v))` might be more future-proof, as there, the first argument plays the role of the input and the second one plays the role of a convolution kernel (as inspired by MATLAB).

martinholters: Letting this sink in a bit more, I think these concerns are non-issues:
* As the old code `transpose`d, that's what the deprecation should do. If we want to similarly allow `Adjoint`, we can always do so later.
* If another argument order turns out to be more natural later, we can add support for it then. But I see no reason why we should not support the order now implemented -- convolution is commutative and associative, after all -- so keeping that order for the deprecation seems least surprising.

martinholters:
Another pair of eyes would be welcome, but absent objections, I plan to merge this soonish.

PR #578, opened 2024-11-05
Deprecate multi-arg and tuple `nextfastfft`

### LINKED ISSUE #291: Deprecate nonscalar `nextfastfft` methods?
If a user has a `dims` tuple we currently have 3 different ways to call `nextfastfft` defined [here](https://github.com/JuliaDSP/DSP.jl/blob/5d3a16bd6a69856e37296f4bd3a588076069660c/src/util.jl#L168-L170):

- `nextfastfft(dims)`
- `nextfastfft(dims...)`
- `nextfastfft.(dims)`

In general this sort of automatic collection handling has been getting deprecated in favor of having the caller use broadcasting (option 3 above). I propose we deprecate the first 2.

It's possible that the multi-arg version (`nextfastfft(x, y)`) is useful, but I don't think that we as a general policy expand other single-arg functions in this way.

martinholters:
Closes #291. I agree with the assessment therein and add that the remaining method is the only one documented.

### DIFF
--- a/src/deprecated.jl
+++ b/src/deprecated.jl
@@ -1,4 +1,9 @@
-# deprecations after 0.6
+# deprecations in 0.8
+import .Util.nextfastfft
+@deprecate nextfastfft(ns::Tuple) nextfastfft.(ns) false
+@deprecate nextfastfft(ns...) nextfastfft.(ns) false
+
+# deprecations in 0.7
 @deprecate freqz(filter::FilterCoefficients{:z}) freqresp(filter, range(0, stop=π, length=250))
 @deprecate freqz(filter::FilterCoefficients{:z}, w) freqresp(filter, w)
 @deprecate freqs(filter::FilterCoefficients{:s}, w) freqresp(filter, w)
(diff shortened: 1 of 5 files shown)


martinholters:
Ah, that's a valid point, it already shows for n=2:
```julia
julia> @btime fft($(rand(            91 , 2^16)));
  121.739 ms (8 allocations: 182.00 MiB)

julia> @btime fft($(rand(nextfastfft(91), 2^16)));
  359.419 ms (8 allocations: 192.00 MiB)
```
while
```julia
julia> @btime fft($(rand(            91 )));
  3.718 μs (6 allocations: 3.48 KiB)

julia> @btime fft($(rand(nextfastfft(91))));
  3.238 μs (6 allocations: 3.48 KiB)
```
(BTW, I have no clue why the 2d-case is that bad.)

But even if we decide that a proper `nextfastfft(ns::Tuple)` is worth doing (and I have no idea what it should look like), we don't know whether any current uses outside DSP.jl actually then do an nD-FFT or just multiple 1D-FFTs of different sizes -- unlikely, but possible -- after calling `nextfastfft`. So changing this might come as a surprise. Unless we go through a deprecation cycle first, that is, and then re-introduce it with proper docs.

martinholters:
Ok, so instead of deprecating, we could also document the mutli-arg and/or tuple method and point out that in the future, the result might be optimized towards nD-FFT and not just be determined point-wise. I'd be fine with both. I still tend to prefer the current PR, but just because the work's already been done. So if anyone opens a PR with the docs change, I'm happy to close this one, otherwise I'll go ahead and merge in a few days.

martinholters:
Closing in favor of #580.

PR #579, opened 2024-11-05
Update CI.yml to use 'min' instead of 1.10

ViralBShah:


martinholters:
I didn't know about `'min'` -- super useful!

PR #580, opened 2024-11-07
Deprecate only multi-arg `nextfastfft`

wheeheee:
Alternative to #578

### DIFF
--- a/src/deprecated.jl
+++ b/src/deprecated.jl
@@ -1,3 +1,7 @@
+# deprecations in 0.8
+import .Util.nextfastfft
+@deprecate nextfastfft(ns...) nextfastfft.(ns) false
+
 # deprecations after 0.6
 @deprecate freqz(filter::FilterCoefficients{:z}) freqresp(filter, range(0, stop=π, length=250))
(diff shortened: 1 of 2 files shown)


martinholters:
I'm good with this.
Why the `\n`s in the docstring?
Why still draft?

martinholters:
What benefit would you expect from that restriction? (Not that I can come up with a valid use-case where support for a heterogeneous tuple would be needed, either.)

martinholters (approved):

martinholters:
> Actually this got me thinking about the current `conv!` offset behavior. IIUC currently some output axes may be offset and the others not. Is this what we want?

Good question. I don't know of any array type that supports a mixture of offset-axes/non-offset-axes along its dimensions, so hard to guess what a use-case might be and what behavior one would wish for. Should we rather throw an error if such a mixture is provided as input? That would prevent us from getting locked into a behavior we might later find not to be what one would expect.

martinholters:
> But I think letting `conv_with_offset` also be defined on `axes(u)` etc, would be better either way.

Only if we are sure that heterogeneous input axes should not give heterogeneous output axes. Which at least I am not. But let me open a dedicated issue for this discussion.

Issue #152, opened 2017-02-16
welch_pgram broken for n != nfft


c42f:
When n != nfft, `getindex(::ArraySplit, i)` fills the fft buffer with `n` elements rather than `nfft`.  This can result in a lot of trailing zeros in the buffer, so periodograms on irregular length data are likely to be contaminated with artifacts from the frequency response of a boxcar window.

On a related note, it would be great if we could change the default windowing function to something better.  I'm aware that there can be subtle tradeoffs when choosing a window, but just about any choice will be a better default than what we have now: currently it's pretty much mandatory to set the window by hand, or the results will contain a lot of spurious ringing.  I don't know a lot about window choice, but matlab seems to default to the Hamming window for `pwelch()`.

simonster:
What would you expect the behavior to be if `n != nfft`?

c42f:
A very good question, I now suspect this issue is more of a misunderstanding on my part about what should be expected from the function than any real problem with the code.  A standard case of pebkec I suppose :-/

To expand on what I was trying to do - I've got samples of a non-stationary time series of varying lengths (actually the pitch of an aircraft). I've estimated the power spectrum using `welch_pgram()` to understand some vibration issues we're having. I don't do a lot of signal processing, so I got the windowing wrong to start with. In fact, I ended up having to implement my own `my_welch()` function to understand how to use the optimized and nicely abstracted version in this package.

So, what went wrong?
* It's quite likely that `n != nfft`, due to the way that defaults are chosen for the function, which means there will be some of zero padding on average; the amount of zero padding can *change* a lot as `n` is changed.
* In my case the signal has a long term trend, so chunks of the signal have a large DC offset, and sensible windowing is particularly important to avoid the spectrum being rubbish in the high frequencies.  In particular, this seems to mean choosing a window which drops to zero (ruling out the Hamming window as a reasonable choice, and certainly ruling out the default of no window).

I'll double check this tomorrow.

simonster:
It's likely that we should apply some window by default if none is specified, since, as you say, this is what other implementations do.

c42f:
The corresponding code in octave and scipy uses the Hamming and Hann(ing) windows respectively.  Octave has some good commentary on the effect of window choice, zero padding, etc, and also mentions removal of the mean by default, which seems like something that might matter a lot in some cases.

https://sourceforge.net/p/octave/signal/ci/default/tree/inst/pwelch.m#l197

It looks like choosing good default settings is a tricky matter.

PR #581, opened 2024-11-07
Deprecate `welch_pgram` without explicitly given `window`

martinholters:
In preparation of #152/#153 to be done in v0.9.

### DIFF
--- a/src/periodograms.jl
+++ b/src/periodograms.jl
@@ -530,11 +530,11 @@ end
 """
     WelchConfig(s::AbstractArray; n=size(s, ndims(s))>>3, noverlap=n>>1,
              onesided=eltype(s)<:Real, nfft=nextfastfft(n),
-             fs=1, window=nothing)
+             fs=1, window)
 
     WelchConfig(nsamples, eltype; n=nsamples>>3, noverlap=n>>1,
              onesided=eltype<:Real, nfft=nextfastfft(n),
-             fs=1, window=nothing)
+             fs=1, window)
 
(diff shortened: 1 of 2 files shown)


src/periodograms.jl
```diff
@@ -601,24 +607,26 @@ See [`periodogram`](@ref) for description of optional keyword arguments.
...
 2-element Vector{Float64}:
  0.9523809523809523
  0.04761904761904761
 
-julia> power(welch_pgram(x; onesided=false))    # 2-sided periodogram
+julia> power(welch_pgram(x; onesided=false, window=nothing))    # 2-sided periodogram
 3-element Vector{Float64}:
  0.9523809523809523
  0.023809523809523805
  0.023809523809523805
```
martinholters, line 620: Due to the short length of `x`, the default `n==3` and
```julia
julia> hanning(3)
3-element Vector{Float64}:
 0.0
 1.0
 0.0
```
is not at all better than the default rect window. In the next example, where `n=5` is explicitly set, I've changed the window to `hanning` as it somewhat reasonable there. I think from a pedagogical viewpoint this is ok, and I'd like to keep a complete rewrite of the example out of the scope of this PR.

Sidenote: a variant of the windows without the zeros at the edges might a useful here, e.g.
```julia
julia> hanning(5)[begin+1:end-1]
3-element Vector{Float64}:
 0.5
 1.0
 0.5
```
But that's also beyond the scope of this PR.

martinholters:
CC @JakeZW

PR #585, opened 2024-11-13
Update Estimation

wheeheee:
mainly remove repeated copies inside the loops

### DIFF
--- a/src/estimation.jl
+++ b/src/estimation.jl
@@ -71,7 +71,7 @@ function esprit(x::AbstractArray, M::Integer, p::Integer, Fs::Real=1.0)
     U, _ = svd(X)
     D, _ = eigen( U[1:end-1, 1:p] \ U[2:end, 1:p] )
 
-    return (angle.(D) .* Fs ./ 2π)
+    return angle.(D) .* (Fs / 2π)
 end
 
 """


src/estimation.jl
```diff
@@ -67,11 +67,11 @@ julia> esprit(x, M, p, Fs)
 function esprit(x::AbstractArray, M::Integer, p::Integer, Fs::Real=1.0)
     count(!isone, size(x)) <= 1 || throw(ArgumentError("`x` must be a 1D signal"))
     N = length(x)
-    X = x[ (1:M) .+ (0:N-M)' ]
+    X = getindex.((x,), (1:M) .+ (0:N-M)')
     U, _ = svd(X)
     D, _ = eigen( U[1:end-1, 1:p] \ U[2:end, 1:p] )
 
-    angle.(D)*Fs/2π
+    return (angle.(D) .* Fs ./ 2π)
```
martinholters, line 74:
```suggestion
    return angle.(D) .* (Fs / 2π)
```
? Otherwise, I think, the division is done for every value in `D` (which the compiler may or may not hoist out of the loop).

martinholters (approved):

martinholters:
> Is it intentional that `quinn` silently mutates the input signal?

I don't think so, no. Good catch!

src/estimation.jl
```diff
@@ -151,81 +150,72 @@ complex, the algorithm is [^Quinn2009].
...
+function quinn(v::Vector{<:Real}, f0::Real, Fs::Real; tol=1e-6, maxiters::Integer=20)
+    fₙ = Fs / 2
 
     # Run a quick estimate of largest sinusoid in x
-    ω̂ = π*f0/fₙ
+    ω̂ = π * f0 / fₙ
 
     # remove DC
-    x .= x .- mean(x)
+    x = v .- mean(v)
```
martinholters, line 164: Could have left the name of the parameter `x` and just make this like `x = x .- mean(x)` to the same effect with slightly less code churn.

martinholters (approved):

PR #591, opened 2024-11-17
Grant permissions to docs workflow

wheeheee:
Switch from using DOCUMENTER_KEY, following example in https://github.com/julia-actions/julia-docdeploy.


martinholters:
Shouldn't the permissions also include `pull-requests: read`?

PR #596, opened 2024-11-25
Fix output length of `resample`

### LINKED ISSUE #186: Resample doesn't resample at given rate
For example, simply trying to double the length of a signal doesn't actually work:
```
x = randn(4)
resample(x,4.0)
resample(x,4//1)
```
both of which return a vector of length 14. I tried various other lengths for x and got that it down by 2 for every case.

I tried altering the amount of zero padding, but increasing the padding by even 1 led to a length 18 signal. I suspect that something is wrong with the actual resample_filter design, but I'm not knowledgeable enough with the material to fix it on my own at this point.

And while we're at it, not being able to put in an integer as a resampling rate is mildly annoying. Is there any design reason for this?

martinholters:
This comprises a few changes that can be considered in on their own, but later ones depend on earlier ones for proper results. In commit order:
* `outputlength(::FIRRational, inputlength)` is fixed to correctly predict the number of valid samples produced by `filt!`. (For context: `filt!` expects a sufficiently large output buffer and returns the number of samples actually written.)
* `outputlength(::FIRArbitrary, inputlength)` is improved to do this correctly more often. Unfortunately, it can still be off-by-one sometimes, which seems almost inevitable. The reason is -- somewhat simplified -- that `filt!` increases the time incrementally, which may be numerically different from considering the whole time-span at once. More complicated in reality, but conceptually: Starting at sample 1 and going in steps of 0.1, how many steps do you need to reach sample 2? Ten, because `10 * 0.1 == 1`? No, one more, because `0.1 + 0.1 + 0.1 + 0.1 + 0.1 + 0.1 + 0.1 + 0.1 + 0.1 + 0.1 < 1`. Now, to be sure to have `outputlength` give the correct result, I think one would have to replicate the same iterative index/phase computations, which seems excessive.
* Add rounding mode support to `inputlength`, specifically for `RoundUp` and `RoundDown` (where the latter is the old behavior and still the default). With `RoundDown`, `inputlength` returns the largest input length such that the actual output length will be at most the given one, i.e. `outputlength(H, inputlength(H, yL)) <= yL < outputlength(H, inputlength(H, yL)+1)`. OTOH, with `RoundUp`,  `inputlength` returns the shortest input length such that the actual output length will be at least the given one, i.e. `outputlength(H, inputlength(H, yL, RoundUp)-1) < yL <= outputlength(H, inputlength(H, yL, RoundUp))`.
* In `resample`, use `inputlength` in `RoundUp` mode to pad the input to the minimum length to get at least `ceil(Int, length(x)*rate)` output samples, then trim the output to the required size as necessary.

Some points I'd especially appreciate feedback on:
* Are the current tests sufficient or should I add more? If so, which?
* Are we confident with the `@assert`s or should that be more defensive?
* Is the `RoundingMode` API for `inputlength` appropriate? Only `RoundUp` and `RoundDown` really make sense there, I guess, so maybe not? (Or should we just drop the `RoundDown` behavior which might not be needed at all?)

Fixes #186.

### DIFF
--- a/src/Filters/stream_filt.jl
+++ b/src/Filters/stream_filt.jl
@@ -702,11 +702,14 @@ function resample(x::AbstractVector, rate::Real, h::Vector)
 
     # Calculate the number of 0's required
     outLen      = ceil(Int, length(x)*rate)
-    reqInlen    = inputlength(self, outLen)
+    reqInlen    = inputlength(self, outLen, RoundUp)
     reqZerosLen = reqInlen - length(x)
     xPadded     = [x; zeros(eltype(x), reqZerosLen)]
 
-    filt(self, xPadded)
+    y = filt(self, xPadded)
+    @assert length(y) >= outLen
+    length(y) > outLen && resize!(y, outLen)
+    return y
 end
 
 function resample(x::AbstractVector, rate::Real)
--- a/src/Filters/stream_filt.jl
+++ b/src/Filters/stream_filt.jl
@@ -111,7 +111,7 @@ function FIRArbitrary(h::Vector, rate_in::Real, Nϕ_in::Integer)
     pfb          = taps2pfb(h,  Nϕ)
     dpfb         = taps2pfb(dh, Nϕ)
     tapsPerϕ     = size(pfb, 1)
-    ϕAccumulator = 1.0
+    ϕAccumulator = 0.0
     ϕIdx         = 1
     α            = 0.0
     Δ            = Nϕ/rate
@@ -228,8 +228,8 @@ function setphase!(kernel::FIRArbitrary, ϕ::Real)
     ϕ >= zero(ϕ) || throw(ArgumentError("ϕ must be >= 0"))
     (ϕ, xThrowaway) = modf(ϕ)
     kernel.inputDeficit += round(Int, xThrowaway)
-    kernel.ϕAccumulator = ϕ*(kernel.Nϕ) + 1.0
-    kernel.ϕIdx         = floor(kernel.ϕAccumulator)
+    kernel.ϕAccumulator = ϕ * kernel.Nϕ
+    kernel.ϕIdx         = 1 + floor(Int, kernel.ϕAccumulator)
     kernel.α            = modf(kernel.ϕAccumulator)[1]
     nothing
 end
@@ -257,7 +257,7 @@ function reset!(kernel::FIRDecimator)
 end
 
 function reset!(kernel::FIRArbitrary)
-    kernel.ϕAccumulator = 1.0
+    kernel.ϕAccumulator = 0.0
     kernel.ϕIdx         = 1
     kernel.α            = 0.0
     kernel.inputDeficit = 1
@@ -336,7 +336,7 @@ function outputlength(kernel::FIRRational, inputlength::Integer)
 end
 
 function outputlength(kernel::FIRArbitrary, inputlength::Integer)
-    ceil(Int, (inputlength-kernel.inputDeficit+1) * kernel.rate - (kernel.ϕAccumulator - 1) / kernel.Δ)
+    ceil(Int, (inputlength-kernel.inputDeficit+1) * kernel.rate - kernel.ϕAccumulator / kernel.Δ)
 end
 
 function outputlength(self::FIRFilter, inputlength::Integer)
@@ -382,7 +382,7 @@ end
 
 function inputlength(kernel::FIRArbitrary, outputlength::Integer, r::RoundingMode=RoundDown)
     d      = r == RoundUp || r == RoundFromZero ? 1 : 0
-    inLen  = floor(Int, (outputlength - d + (kernel.ϕAccumulator - 1) / kernel.Δ)/kernel.rate) + d
+    inLen  = floor(Int, (outputlength - d + kernel.ϕAccumulator / kernel.Δ)/kernel.rate) + d
     inLen += kernel.inputDeficit - 1
 end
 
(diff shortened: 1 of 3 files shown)


test/resample.jl
```diff
@@ -94,16 +94,16 @@ end
     idxLower  = round(Int, yLen/3)
     idxUpper  = idxLower*2
     yDelta    = abs.(f32_y[idxLower:idxUpper].-yy[idxLower:idxUpper])
-    @test all(map(delta -> abs(delta) < 0.005, yDelta))
+    @test all(map(delta -> abs(delta) < 0.0055, yDelta))
```
martinholters, line 97: I haven't looked into why this tolerance needs to be increased. Probably just bad luck because with different signal length, a slightly different excerpt is considered, but I should double-check that's all there is to it.

test/resample.jl
```diff
@@ -94,16 +94,16 @@ end
...
     # https://github.com/JuliaDSP/DSP.jl/issues/317
     @testset "Buffer length calculation" begin
-        @test length(resample(sin.(1:1:35546),  1/55.55)) == 641
-        @test length(resample(randn(1822), 0.9802414928649835)) == 1787
-        @test length(resample(1:16_367_000*2, 10_000_000/16_367_000)) == 20_000_001
-        @test resample(zeros(1000), 0.012) == zeros(13)
+        @test length(resample(sin.(1:1:35546),  1/55.55)) == 640
+        @test length(resample(randn(1822), 0.9802414928649835)) == 1786
+        @test length(resample(1:16_367_000*2, 10_000_000/16_367_000)) == 20_000_000
+        @test resample(zeros(1000), 0.012) == zeros(12)
```
martinholters, line 106: These lengths now all match expectations:
```julia
julia> 35546 / 55.55
639.8919891989199

julia> 1822 * 0.9802414928649835
1786.0

julia> 16_367_000*2 * 10_000_000/16_367_000
2.0e7

julia> 1000 * 0.012
12.0
```

martinholters:
> After some simplification, I think there is a way to make sure that the output length matches.

For ` FIRArbitrary` you mean, I suppose? That would be great. I'm curious to see it!

> Is there an obstruction to generating `ϕAccumulator` with a pre-calculated range?

Not sure what you're trying to say here. I guess we could consider changing the state from `ϕAccumulator` to a counter and then calculate `ϕAccumulator` from something like `mod(counter*Δ)`. Then it would be much easier to determine the final value for that counter, I think.

I tried to avoid putting too many changes in this PR for sake of reviewability, and rewriting the inner workings of `FIRArbitrary` is somewhat on the border. It's not immediately required for fixing #186, but surely closely related, so might make sense.

test/resample.jl
martinholters: Ah, this is hilarious: The test signal `y` is obtained by upsampling (using `resample`) an input `x` by `ratio = 3.141592653589793`, but then the reference signal `yy` is computed based on a time-step reduced by the factor `length(y)/length(u)` (where `length(u)==1000`), which only approximates the `ratio`. And this PR changes `length(y)` from 3141 to 3142 (ref. https://github.com/JuliaDSP/DSP.jl/pull/596#issuecomment-2505660227 on whether that is good) making that approximation worse. Arguably, it would be more reasonable to use `ratio` for the reference signal as well, which would allow this bound to be reduced to `0.00025` and make the impact of this PR on that error insignificant.

I'll open a separate PR to improve the test and then after a rebase, adjusting the test here should be unnecessary.

src/Filters/stream_filt.jl
```diff
@@ -565,18 +557,18 @@ end
...
 
 function update!(kernel::FIRArbitrary)
     kernel.ϕAccumulator += kernel.Δ
 
-    if kernel.ϕAccumulator > kernel.Nϕ
-        kernel.xIdx        += div(kernel.ϕAccumulator-1, kernel.Nϕ)
-        kernel.ϕAccumulator = mod(kernel.ϕAccumulator-1, kernel.Nϕ) + 1
+    if kernel.ϕAccumulator >= kernel.Nϕ
+        Δx, kernel.ϕAccumulator = divrem(kernel.ϕAccumulator, kernel.Nϕ)
+        kernel.xIdx += round(Int, Δx)
```
martinholters, line 567:
```suggestion
        kernel.xIdx += Int(Δx)
```
just to make it explicit that this should not actually round, but just be a type conversion?

src/Filters/stream_filt.jl
```diff
@@ -565,18 +557,18 @@ end
...
     end
 
-    kernel.ϕIdx = floor(Int, kernel.ϕAccumulator)
-    kernel.α    = kernel.ϕAccumulator - kernel.ϕIdx
+    kernel.α, foffset = modf(kernel.ϕAccumulator)
+    kernel.ϕIdx = 1 + round(Int, foffset)
```
martinholters, line 571:
```suggestion
    kernel.ϕIdx = 1 + Int(foffset)
```
likewise

martinholters:
> I suppose you mean `ceil(Int, length(input) * ratio)`?

Of course, yes.

> The only reason I can think of is if `length(input) * ratio < 0.5`, this still gives an output, and somebody wants that.

Hm... That's not a reason that convinces me. A single sample does not seem that much more helpful than none. Would `round` be more intuitive? (It is to me.) Change?

martinholters:
Good point. Looks like they both round up (for Matlab I found that to be documented as well), so reasonable for us to do the same.

I guess then this is good to go then.

martinholters:
Ah yes, turns out `inputlength` was incorrect for cases where no rounding in the length computation occurred. Fix incoming.

Should we reconsider this question: Are we confident with the `@assert`s or should that be more defensive?

martinholters:
Then let's leave the `@assert`s in. We can always reconsider if these actually do fire in practice.

Do you want to take this for another test flight and see if you find more problems or should we merge?

Issue #372, opened 2020-08-17
Stateful filtering fails with multi-dim array


hergum:
Using more than one dimension in the input array makes `filt` fail for stateful filter. 

`s = rand(10,3)`
`df = digitalfilter(Lowpass(0.25), Butterworth(4))`
`f = DF2TFilter(df)`
`filt(f,s)`

_ERROR: MethodError: no method matching filt(::DF2TFilter{SecondOrderSections{Float64,Float64},Array{Float64,2}}, ::Array{Float64,2})_

One-dimensional input works as expected. Two-dimensional inputs works along the first dimension, as expected, for stateless filtering `filt(df,s)`.

Any ideas on how to make this work according to the doc?

galenlynch:
You're right that the docs suggest that this should work. However, that doc string was written for normal filter coefficients, and not stateful filters. It's not really clear to me what the expected behavior would be when a stateful filter is applied to a matrix. For stateless filtering each column is treated as a separate signal, where the filter state is reset at the start of each column of the signal, and the filter state is discarded at the end of each column. If you were to do the same with a stateful filter, it's not clear what the filter state should be either at the start of each column, or after all the columns have been filtered (i.e. which column of the signal should the state reflect).

Depending on what you are expecting the filter state to do between columns of your signal, you could write an explicit loop accordingly. The only thing that really makes sense to me, given your decision to use a stateful filter, is that each column is a continuation of the same signal. In this case, there is no reason to use a matrix instead of a vector. If you instead want to maintain separate filter state for each column, then you should make separate filter objects for each column.

Could you explain your use case a little bit more? Are you processing very long signals in batches, or are you doing some on-line filtering of signals? Are you filtering multiple channels of data that should be treated separately? What is the reason that you want to use stateful filters in the first place?

hergum:
I would like the stateful filter to keep track of the state of the filter for each column.

My use-case is I get long 2D signals in batches. Each row is independent, and all rows should be processed the same way. The rows are continuous in time, and must be processed as if I had the full signal available, so I need to preserve the filter state. Its for a medical device, but the use case might as well be wanting to low-pass filter the audio track of 15 radio stations. In matlab I'm used to doing things like:

```
>> [b,a]=butter(4,0.25);
>> [s,state]=filter(b,a,rand(15,800),[],2); % [] means no state for the first batch
>> size(state)

ans =

     4    15
[...]

PR #599, opened 2024-12-02
Let `filt`/`filt!` return state if provided as input

martinholters:
As argued in https://github.com/JuliaDSP/DSP.jl/issues/372#issuecomment-2506259561, I don't think the possibility to pass in state to `filt` and `filt!` in the current form is very helpful, as there is no supported way of obtaining such state. There is the internal `_zerosi`, but explicitly passing the zero-state is equivalent to passing no state, so why bother. Otherwise, the filter state should probably be considered an implementation detail and it is undocumented, so hand-crafting it is not recommended. (With the exception of the use in DSPs own `filtfilt`, maybe.) The most likely use of passing in state is to resume a filtering operation, e.g. for blockwise processing. (Yes, that's what using a `DF2TFilter` allows, too.) But to enable that, one would need the final state to be provided by `filt`/`filt!`.

And that's what this PR does, but only if an input state is provided. I.e. one can do
```julia
(outblock1, state) = filt(H, inblock1, state)
(outblock2, state) = filt(H, inblock2, state)
```

The implementation is somewhat rough-and-dirty for now, as I first wanted to get a feel for this and see if we want to do this. There are some questions on the API worth thinking about:
* How should the initial state be obtained? Make `_zerosi` public (with a better name)? Or support some placeholder value for the `state` parameter that indicates a zero-state should be set up?
* Should `filt!` modify the `state` instead of returning it? Seems reasonable, but up to now, care was taken not to modify the state.
* For multi-column input, we support passing in a single state which will then be replicated for each column. Do we need to keep supporting that?

Putting on the milestone as this would be breaking.

### DIFF
--- a/src/Filters/filt.jl
+++ b/src/Filters/filt.jl
@@ -16,8 +16,7 @@ Same as [`filt()`](@ref) but writes the result into the `out`
 argument. Output array `out` may not be an alias of `x`, i.e. filtering may
 not be done in place.
 """
-filt!(out, f::PolynomialRatio{:z}, x::AbstractArray, si=_zerosi(f, x)) =
-    filt!(out, coefb(f), coefa(f), x, si)
+filt!(out, f::PolynomialRatio{:z}, x::AbstractArray, si...) = filt!(out, coefb(f), coefa(f), x, si...)
 
 """
     filt(f::FilterCoefficients{:z}, x::AbstractArray[, si])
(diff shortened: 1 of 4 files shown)


martinholters:
> Would it be too radical to remove `si` here and let those who care about state use `DF2TFilter`?

I've been wondering the same. I'll give it try and open an alternative PR if it looks promising.

> * I think something like a public `zerosi` would be better since it might help users get some idea of the dimensions of the state, although dispatching on say `::ZeroSI` does strike me as more aesthetic.

Another point in favor of a `zerosi` would be a use-case where you do block-wise processing in a `for`-loop. It would be inconvenient (or type-unstable) to have to special-case the first-block to feed in a place-holder state. Better to allocate the state using a dedicated function before the loop.

> * The change to returning a tuple with the state is sufficiently breaking, so preserving backwards-compatibility for not mutating state here isn't so important imo, but see below

Right. That would allow the return-type to remain the same, but then this might be silently breaking, i.e. the same code still works, but produces different results. We might want to avoid that. Also, `filt` should probably still not mutate the state, which asks for yet another entry-point which mutates the state, but contrary to `filt!` allocates the output. This sounds less attractive the more I think about it, maybe getting rid of `si` in favor of `DF2TFilter` really is the way to go.

> * I feel passing in a single state for multiple columns is convenient and should be kept. But if so, the state passed in has different dimensions and it doesn't make sense to mutate it, hence it might be inconsistent if we only mutate `si` in the case of multidimensional input + state...

We could make that orthogonal by introducing some state-setup function that could both play the role of `zerosi` or do the appropriate `repeat` dance to fan out a single input state for multiple signals.

martinholters:
Superseded by #603.

PR #601, opened 2024-12-02
Small fixups to #600

martinholters:
* Fix multi-column `_small_filt_fir!` for `ndims(x)>2`.
* Consistently require mutli-column state to match input layout. (A single state as input is still supported.) For a 2D input -- the only shape properly supported prior to #600 -- this doesn't change anything.

### DIFF
--- a/src/Filters/filt.jl
+++ b/src/Filters/filt.jl
@@ -69,7 +69,7 @@ function filt!(out::AbstractArray, f::SecondOrderSections{:z}, x::AbstractArray,
     initial_si = si
     si = similar(si, axes(si)[1:2])
     for col in CartesianIndices(axes(x)[2:end])
-        copyto!(si, view(initial_si, :, :, N > 2 ? col : 1))
+        copyto!(si, view(initial_si, :, :, N > 2 ? col : CartesianIndex()))
         _filt!(out, si, f, x, col)
     end
     out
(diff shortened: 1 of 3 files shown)


martinholters:
Indeed, I think the case of passing in `si` with per-column-state doesn't work even with these changes as it is unclear whether the state for a, say, $L \times M_1 \times M_2$ input should be $N \times (M_1 \cdot M_2)$ or $N \times M_1 \times M_2$, and it is assumed differently in different places. I'll add tests that cover these and extend the fixup to always use the latter.

src/Filters/filt.jl
```diff
@@ -99,10 +98,8 @@ end
 # filt! variant that preserves si
 function filt!(out::AbstractArray, f::Biquad{:z}, x::AbstractArray,
                     si::AbstractArray{S,N}=_zerosi(f, x)) where {S,N}
-    ncols = Base.trailingsize(x, 2)
-
     size(x) != size(out) && throw(DimensionMismatch("out size must match x"))
-    (size(si, 1) != 2 || (N > 1 && Base.trailingsize(si, 2) != ncols)) &&
+    (size(si, 1) != 2 || (N > 1 && size(si)[2:end] != size(x)[2:end])) &&
```
martinholters, line 102: The checks base on `trailingsize` were more permissive, but the indexing below would then fail if the sizes didn't match, so test up here. (Also, `trailingsize` was accessing a `Base` internal.)

PR #603, opened 2024-12-03
Deprecate `filt`/`filt!` with `si` parameter

martinholters:
The radical alternative to #599. If you need stateful filtering, use `DF2TFilter`. After #602, this shouldn't take away any functionality, just restrict the interface to enforce the use of `DF2TFilter`.

### DIFF
--- a/src/Filters/filt.jl
+++ b/src/Filters/filt.jl
@@ -6,33 +6,28 @@
 using ..DSP: _filt_fir!, _filt_iir!
 
 ## PolynomialRatio
-_zerosi(f::PolynomialRatio{:z,T}, ::AbstractArray{S}) where {T,S} =
-    zeros(promote_type(T, S), max(-firstindex(f.a), -firstindex(f.b)))
 
 """
-    filt!(out, f, x[, si])
+    filt!(out, f, x)
 
 Same as [`filt()`](@ref) but writes the result into the `out`
 argument. Output array `out` may not be an alias of `x`, i.e. filtering may
 not be done in place.
 """
-filt!(out, f::PolynomialRatio{:z}, x::AbstractArray, si=_zerosi(f, x)) =
-    filt!(out, coefb(f), coefa(f), x, si)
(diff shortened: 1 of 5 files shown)


martinholters:
Re-added that test.

martinholters:
> How well would this be received downstream? i.e. does anyone actually pass `si` to `filt`?

I don't know. I don't think it's particularly useful in its current form (hence #599 and this PR), so widespread use would surprise me. However, I don't want to rush this PR, so that more people get the chance to stumble upon it and raise any objections they might have. That said, I think I'll add deprecations so that code passing `si` to `filt` can continue to function and any users get more time to switch to `DF2TFilter`. (Or complain loud enough that we can add it back during 0.8.x, potentially with #599 reconsidered.)

martinholters:
Hm, that's a bit unfortunate indeed. We could also drop the `<:FilterCoefficients{:z}` from the `DF2TFilter` type parameter (or relax it) to allow `(b,a)::Tuple{Vector,Vector}` to be stored as `coef`. Or we make `coefa`/`coefb` non-allocating. The latter would be nice anyway, but is certainly trickier, too. I see two options:
* Implement a small `AbstractVector` subtype that forwards `getindex` to the underlying `LaurentPolynomial` after the necessary index transformation. But that would need some benchmarking how bad this index transformation is in tight loops like filtering.
* Revert `PolynomialRatio` back to storing the coefficients directly as `Vector`s. I switched these to `LaurentPolynomial`s a while ago because there was some confusion within DSP.jl how the coefficients had to be interpreted, and by representing them as `Polynomial`s, that was made clear. But I don't think anything critically depends on these being `Polynomial`s.

None of these options strikes me as ideal. But I'm fairly confident there is some way forward, so no reason to block this PR I'd hope.

martinholters:
Deprecations are in the works. Stay tuned.

martinholters:
> I guess for two AbstractVectors the repeat in the deprecated functions basically just copies

That should be covered by the more specific deprecations for that case, or am I misunderstanding?

PR #607, opened 2024-12-06
Simplify (and slightly speed up) `coefb`/`coefa`

martinholters:
Noticed this the other day. Main benefit is readability. Only in cases where trailing/leading zeros had to be re-appended/prepended, one allocation is eliminated. Otherwise, performance is almost unchanged.


martinholters:
Well, yes, the `getindex` branches on whether the requested index is inside the stored coefficients, but `Polynomials.coeffs` also calls `getindex` internally, so not much difference there.

Issue #605, opened 2024-12-04
Improve online filter functions


ufechner7:
If you want to filter online data, you can use the following function:
```
"""
    apply_filter(tfilter::DF2TFilter, measurement)

Apply the filter to the measurement.

# Arguments
- `tfilter`: The filter, created with `create_filter` and converted to a DF2TFilter
- `measurement`: The measurement value
"""
function apply_filter(tfilter::DF2TFilter, measurement)
    results = zeros(1)
    measurements = ones(1) * measurement
    @views filt!(results[1:1], tfilter, measurements)
    return results[1]
end
```
This is an ugly workaround. We need a function that does the same, but:
- does not require the allocation of one element arrays
- does not allocate at all (important for real-time control systems)

In addition, the fact that this function does not accept a filter of type `ZeroPoleGain` is annoying.

martinholters:
> * does not require the allocation of one element arrays

Yes, this is quite unfortunate. Would be nice if we could tack this on `filt` as accepting a single number as the input "signal". However, we could not do that too broadly as we have
`filt(h::Vector, x::AbstractVector, rate::AbstractFloat)` (FIR resampling) and `filt(b::Union{Number, AbstractVector}, a::Union{Number, AbstractVector}, x::AbstractArray{T})`. Now if we allowed `x` to be a number in the latter, that would cause ambiguities. Arguably, this single-sample processing only makes sense if state is preserved, i.e. when using `DF2TFilter`. So adding a method `filt(f::DF2TFilter, x::Number)` returning a single number seems reasonable to me and the inconsistency of not doing this for the other `filt` variants can be justified. (Oh, just noticed that `filt(self::FIRFilter{<:FIRStandard}, x)` would be another candidate; but that also is unambiguous when a `x::Number` method is added.)


> * does not allocate at all (important for real-time control systems)

You get that if you base the `DF2TFilter` on `SecondOrderSections`:
```julia
julia> typeof(H)
DF2TFilter{SecondOrderSections{:z, Float64, Float64}, Matrix{Float64}}

julia> @ballocated filt!(y, H, x)
0
```
It surely would be nice for `PolynomialRatio` to also allow this, but that is not so easily solved.
(This is for `filt!`, of course, but should translate to a potential future single-sample `filt`.)
 
> In addition, the fact that this function does not accept a filter of type `ZeroPoleGain` is annoying.

[...]

PR #608, opened 2024-12-12
Prettier tests

wheeheee:


test/FilterTestHelpers.jl
```diff
@@ -58,33 +58,22 @@ function tffilter_accuracy(f1, f2, accurate_f)
     accuracy_check(loss(a1, accurate_a), loss(a2, accurate_a), "a")
 end
 
-function zpkfilter_accuracy(f1, f2, accurate_f; relerr=1, compare_gain_at=nothing, eps=nothing)
+function zpkfilter_accuracy(f1, f2, accurate_f; relerr=1, compare_gain_at=nothing, eps=0.0)
```
martinholters, line 61:
```suggestion
function zpkfilter_accuracy(f1, f2, accurate_f; relerr=1, compare_gain_at=nothing, atol=0.0)
```
?

test/estimation.jl
```diff
@@ -22,77 +22,58 @@ end
...
-    (f_est_real, maxiter) = quinn(sr, 50, fs)
-    @test maxiter == false
-    @test isapprox(f_est_real, fr, atol = 1e-3)
+    sr = cospi.(2 * fr * t .+ 1 / 4.2)
+    function test_quinn(f, s, args...)
+        (f_est_real, maxiter) = quinn(s, args...)
+        @test maxiter == false
+        @test isapprox(f_est_real, f; atol=1e-3)
+        return nothing
+    end
```
martinholters, line 65: Move the function definition above the variables specific to only a subset of the the tests?

test/filt_stream.jl
```diff
@@ -102,8 +102,9 @@ function test_singlerate(h::AbstractVector{T}, x::AbstractVector) where T
     @printfifinteractive( "\n\tfilt filt. Piecewise for first %d inputs\n\t\t", length(x1) )
     reset!(myfilt)
     @timeifinteractive begin
-        for i in 1:length(x1)
-            y1[i] = filt(myfilt, x1[i:i])[1]
+        one_buffer = similar(x1, 1)
+        for (i, p) in enumerate(x1)
+            y1[i] = filt(myfilt, setindex!(one_buffer, p, 1))[1]
```
martinholters, line 107: I'm not sure these `@timeinteractive` should be taken seriously at all (I never look at them), so this optimization might be nothing more than an unnecessary complication. But even if care about the allocations here, wouldn't
```suggestion
        for i in eachindex(x1)
            y1[i] = filt(myfilt, @view(x[i:i]))[1]
```
cut it? (And likewise below.)

test/filt_stream.jl
```diff
@@ -312,16 +316,15 @@ function test_arbitrary(Th, x, resampleRate, numFilters)
...
     end
 
-    commonLen = minimum(length.((naiveResult, statelessResult, statefulResult, piecewiseResult)))
-    resize!(naiveResult, commonLen)
-    resize!(statelessResult, commonLen)
-    resize!(statefulResult, commonLen)
-    resize!(piecewiseResult, commonLen)
+    results = (naiveResult, statelessResult, statefulResult, piecewiseResult)
+    commonLen = minimum(length, results)
+    resize!.(results, commonLen)
```
martinholters, line 327: I'd prefer the more explicit form here. This line changing all of the `xxxResult`s is a bit non-obvious. (Using `minimum(length, ...)` instead of `minimum(length.(...))` is good, though.)

src/diric.jl
```diff
@@ -36,7 +36,7 @@ julia> diric(0, 4)
 ```
 """
 function diric(Ω::T, n::Integer) where T <: AbstractFloat
-    n > 0 || throw(ArgumentError("n=$n not positive"))
+    n > 0 || throw(DomainError(n, "n not positive"))
```
martinholters, line 39: I wouldn't file this under "prettier tests", but still a good change.

test/filt_stream.jl
martinholters: Ah, that's unfortunate. Rework after fixing #605, then :smile:

martinholters (approved):

PR #609, opened 2024-12-12
Reduce macro usage

wheeheee:
also use more `tanpi`

### DIFF
--- a/src/Filters/filt_order.jl
+++ b/src/Filters/filt_order.jl
@@ -380,7 +380,8 @@ for (fcn, filt) in ((:ellipord, "Elliptic (Cauer)"), (:cheb1ord, "Chebyshev Type
 
         Integer and natural frequency order estimation for $($filt) Filters. `Wp` and `Ws`
         indicate the passband and stopband frequency edges, and `Rp` and `Rs` indicate the
-        maximum ripple loss in the passband and the minimum ripple attenuation in the stopband, (in dB.)\n
+        maximum loss in the passband and the minimum attenuation in the stopband, (in dB.)
+
         Based on the ordering of passband and stopband edges, the Lowpass or Highpass filter type is inferred.
         `N` indicates the smallest integer filter order that achieves the desired specifications,
         and `ωn` contains the natural frequency of the filter, (in this case, simply the passband edge.)\n


src/Filters/filt_order.jl
```diff
@@ -327,76 +327,73 @@ end
...
+for (fcn, filt) in ((:ellipord, "Elliptic (Cauer)"), (:cheb1ord, "Chebyshev Type I"))
     @eval begin
         """
             (N, ωn) = $($fcn)(Wp::Real, Ws::Real, Rp::Real, Rs::Real; domain::Symbol=:z)
+            (N, ωn) = $($fcn)(Wp::Tuple{Real, Real}, Ws::Tuple{Real, Real}, Rp::Real, Rs::Real; domain::Symbol=:z)
 
         Integer and natural frequency order estimation for $($filt) Filters. `Wp` and `Ws`
         indicate the passband and stopband frequency edges, and `Rp` and `Rs` indicate the
-        maximum loss in the passband and the minimum attenuation in the stopband, (in dB.)\n
+        maximum ripple loss in the passband and the minimum ripple attenuation in the stopband, (in dB.)\n
```
martinholters, line 380: No... I think neither "ripple loss" nor "ripple attenuation" are correct here. We're not loosing or attenuating the ripple. Loss and non-ideal attenuation are consequences of the frequency response ripple. If we want to mention ripple, it has to go somewhere else. But I'm not sure we have to.

src/Filters/filt_order.jl
```diff
@@ -327,76 +327,73 @@ end
...
+    if (domain == :z)
+        ωn = (2 / π) * atan(Ωp)
+    else
+        ωn = Ωp
+    end
+    N, ωn
+end
+
+function ordfreq_est(order_estimate, domain::Symbol,
+        Wp::Tuple{Real,Real}, Ws::Tuple{Real,Real}, Rp::Real, Rs::Real)
```
martinholters, line 351:
```suggestion
function ordfreq_est(order_estimate::F, domain::Symbol,
        Wp::Tuple{Real,Real}, Ws::Tuple{Real,Real}, Rp::Real, Rs::Real) where {F}
```

src/windows.jl
```diff
@@ -632,33 +632,32 @@ const IntegerOr2 = Union{Tuple{Integer, Integer}, Integer}
 const RealOr2 = Union{Tuple{Real, Real}, Real}
 const BoolOr2 = Union{Tuple{Bool, Bool}, Bool}
 
+function matrix_window(func, dims::Tuple{Integer,Integer}, arg::Union{RealOr2,Nothing}=nothing;
+        padding::IntegerOr2=0, zerophase::BoolOr2=false)
```
martinholters, line 636:
```suggestion
function matrix_window(func::F, dims::Tuple{Integer,Integer}, arg::Union{RealOr2,Nothing}=nothing;
        padding::IntegerOr2=0, zerophase::BoolOr2=false) where {F}
```

src/windows.jl
```diff
@@ -632,33 +632,32 @@ const IntegerOr2 = Union{Tuple{Integer, Integer}, Integer}
...
+    length(dims) == 2 || throw(ArgumentError("`dims` must be length 2"))
```
martinholters, line 637: `dims::Tuple{Integer,Integer}`, so its length has to be two.
```suggestion
```

martinholters:
Any special reason those `@inline`s?

martinholters:
Ah, yes, makes sense.

martinholters (approved):

PR #610, opened 2024-12-12
Simplify 3-arg `conv`, remove unnecessary inbounds

wheeheee:


src/dspbase.jl
```diff
@@ -794,13 +794,13 @@ function conv(u::AbstractVector{T}, v::Transpose{T,<:AbstractVector}, A::Abstrac
...
-    n = length(v)+size(A,2)-1
+    m = length(u) + size(A, 1) - 1
+    n = length(v) + size(A, 2) - 1
     B = zeros(T, m, n)
-    B[1:size(A,1),1:size(A,2)] = A
-    u = fft([u;zeros(T,m-length(u))])
-    v = fft([v transpose(zeros(T,n-length(v)))])
-    C = ifft(fft(B) .* (u * v))
+    B[CartesianIndices(A)] = A
+    u, vt = fft.(_zeropad.((u, transpose(v)), (m, n)))
```
martinholters, line 801: I think this is easier both on the compiler and for anyone reading the code if done more explicitly instead of broadcasting:
```suggestion
    u = fft(_zeropad(u, m))
    vt = fft(_zeropad(transpose(v), n))
```

src/dspbase.jl
```diff
@@ -794,13 +794,13 @@ function conv(u::AbstractVector{T}, v::Transpose{T,<:AbstractVector}, A::Abstrac
...
+    p = plan_fft(B)
+    C = inv(p) * ((p * B) .*= u .* transpose(vt))
```
martinholters, line 803: Are you sure using an FFT plan is beneficial here? For a somewhat reduced example, `ifft(fft(B))` seems to have fewer allocations than `p = plan_fft(B); inv(p) * (p * B)`. I think explicitly creating a plan is only beneficial if it is used more than once, and inverting it does not count as a use in this sense.
```suggestion
    C = ifft(fft(B) .*= u .* transpose(vt))
```

martinholters (commented):
Nice to get rid of those `@inbounds`. I trust you to have done some benchmarking to confirm they are not needed.

martinholters (approved):

Issue #598, opened 2024-11-29
Tag DSP v1


dlfivefifty:
There hasn't been a breaking change in 3 years... probably a good time to tag 1.0

dlfivefifty:
I would say v0.8 should just be v1.0 in that case.

Using v0.x longer than necessary has two problems: (1) it marks the package as unstable, which in this case it clearly isn't and (2) you cannot take fulll advantage of semantic versioning.

dlfivefifty:
And (3) it annoys me slightly when my Project.toml files for a v1.x package depend on a v0.x package 😅

martinholters:
The absence of breaking changes has primarily been due to overall slow development. I'm not sure that should be taken as a sign of stability.

That said, I agree that we should aim at v1.0 in the foreseeable future. My preferred path forward would be to tag v0.8 soon(!), wait for a few months, say, whether anything severe shows up, and if everything seems calm, tag v1.0 after removing deprecations.

dlfivefifty:
I would take the lack of breaking changes as the _definition_ of stability. I don't think "v1.0" means the finished product, and if there are major issues there can always be a v2.0...

martinholters:
> I would take the lack of breaking changes as the _definition_ of stability.

What about breaking changes we are aware must happen but haven't come around to executing?

Also, there has been a breaking change at least almost a year ago (#458), which has been sitting around for almost three years. Just because we waited for a long time to do a breaking release doesn't mean there were no breaking changes in the pipeline. Would have been funny to release v1.0 at some point knowing the release after would probably have to be v2.0.

dlfivefifty:
I don’t think it’s strange at all if v2 is the next release after v1, according to the principles of semantic versioning.

there will almost always be breaking changes in the pipeline (unless the package is mothballed)

martinholters:
If I got the regex right, there are a total of 64 Julia packages registered with v2.0.0 immediately following v1.0.0. So while perfectly legal, it's at least unusual.

What would you consider a criterion for _not_ calling a release v1.0?

dlfivefifty:
I would say the criteria is that the interface is unstable in the short term. I.e. if v2.0 will be released within weeks or a month. 

But if v1.0 is going to last 3 years and v2.0 is the next release in 2028, then that is probably fine 😀

martinholters:
[...]

PR #613, opened 2024-12-15
Move convolution code to Convolutions.jl submodule

wheeheee:
Not sure if `xcorr` belongs in there, and currently the `deconv` function calls `filt`, so that can't be moved into convolutions.jl as-is either.


martinholters:
Not that I'm necessarily opposed to this, but what's the benefit?

dlfivefifty:
I wonder about whether FFTW.jl should be a hard dependency. Alternatively,  a simpler package ConvolutionsBase.jl which only defines `conv` and `conv!` without implementing them might be a better solution.

For example, in InfiniteArrays.jl overloads of `conv` I don't actually need the fft at all:

https://github.com/JuliaArrays/InfiniteArrays.jl/blob/master/ext/InfiniteArraysDSPExt.jl

martinholters:
Right, if we want to move the convolutions stuff to a seperate repo, there are a bunch of questions to answer. But I don't think this PR or #598 are the best places to discuss that. Maybe open a dedicated issue or discussion?

WRT this PR, I think we should wait until that discussion has settled before actually setting into anything in motion. But as a draft to try things out, it might indeed be helpful.

dlfivefifty:
https://github.com/JuliaDSP/DSP.jl/issues/615

wheeheee:
Should probably make a decision about whether this moves out before v0.9

PR #614, opened 2024-12-16
Reduce allocations for IIR filtering

martinholters:
Avoid the necessity to re-allocate one of the coefficient vectors if they have different lengths by letting `_filt_iir!` handle different lengths.
~(The second commit is some performance-neutral clean-up. I'm not 100% sure we should do it.)~

For equal-length coefficient vectors performance is practically the same:
```
julia> @btime filt!($(zeros(100)), $(rand(2)), $([1; rand(1)]), $(rand(100))); # a and b both length 2
  344.546 ns (1 allocation: 64 bytes) # PR
  348.047 ns (1 allocation: 64 bytes) # master

julia> @btime filt!($(zeros(100)), $(rand(3)), $([1; rand(2)]), $(rand(100))); # a and b both length 3
  319.090 ns (1 allocation: 80 bytes) # PR
  317.876 ns (1 allocation: 80 bytes) # master

julia> @btime filt!($(zeros(100)), $(rand(2)), $([1; rand(2)]), $(rand(100)));
  294.293 ns (1 allocation: 80 bytes) # PR
  330.063 ns (2 allocations: 160 bytes) # master

julia> @btime filt!($(zeros(100)), $(rand(3)), $([1; rand(1)]), $(rand(100)));
  285.819 ns (1 allocation: 80 bytes) # PR
  332.861 ns (2 allocations: 160 bytes) # master

julia> @btime filt!($(zeros(100)), $(rand(100)), $([1; rand(1)]), $(rand(100)));
  1.247 μs (1 allocation: 896 bytes) # PR
  1.495 μs (2 allocations: 1.75 KiB) # master
```

### DIFF
--- a/src/Filters/filt.jl
+++ b/src/Filters/filt.jl
@@ -162,14 +162,7 @@ function filt!(out::AbstractArray{<:Any,N}, f::DF2TFilter{<:PolynomialRatio,Arra
         mul!(out, x, b[1])
     else
         a = coefa(f.coef)
-        as = length(a)
-        bs = length(b)
-        if as != 1
-            if as < n
-                append!(a, zero(eltype(a)) for _ in 1:(n-as))
-            elseif bs < n
-                append!(b, zero(eltype(b)) for _ in 1:(n-bs))
-            end
+        if length(a) != 1
             for col in CartesianIndices(axes(x)[2:end])
                 _filt_iir!(out, b, a, x, view(si, :, col), col)
             end
--- a/src/dspbase.jl
+++ b/src/dspbase.jl
@@ -45,9 +45,6 @@ function filt!(out::AbstractArray, b::Union{AbstractVector, Number}, a::Union{Ab
         a = @noinline broadcast(/, a, norml)
         b = @noinline broadcast(/, b, norml)
     end
-    # Pad the coefficients with zeros if needed
-    bs<sz   && (b = copyto!(zeros(eltype(b), sz), b))
-    1<as<sz && (a = copyto!(zeros(eltype(a), sz), a))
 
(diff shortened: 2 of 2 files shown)


martinholters:
Hm, I've tried with the same benchmarks and the results are somewhat hard to interpret, as it's close to measurement noise, but to me it looks like this:
* On Julia 1.10.7, the PR is slightly better than master.
* On Julia 1.11.2, master is slightly better than this PR.
* But not doing the views thing, the PR is slightly better on Julia 1.11.2, too.

All of that might be dominated by randomness, though. Anyway, I've removed the rewrite to views to have this more focused. Can you re-do the benchmarks and see whether it's now an improvement for you, too?

martinholters:
So here's my hypothesis to what's going on: The run-time of the loops only scales approximately linear wrt. iteration count, depending on how well the iteration count matches the vectorization width. So if `max(length(a), length(b))` (before any padding) is a "good" length, but `min(length(a), length(b))` is a "bad" length (and their difference is a "bad" length for the second loop, too), then this PR can lead to inferior performance.
This also seems to be system specific (not unexpected if my suspicion is correct), as for the example you have provided, I only see about 10% run-time increase compared to your 40%. 

Overall, my feeling is that we shouldn't over-value these benchmarks. Users who really want to squeeze the last bits of performance out of `filt!` on their specific system can always just zero-pad their coefficients until they reach an optimum. For your example, I see improved performance replacing `b` with `[rand(15); zeros(2)]`. YMMV.

martinholters:
@wheeheee I'm not too excited of the commit you've pushed. It accesses internals of `LaurentPolynomial` which AFAICT are non-public/undocumented, and the gain is not that much. Are you ok with me undoing that and then merging?

PR #617, opened 2024-12-28
unwrap: Initialize image reliability in deterministic order (Fixes #616)

### LINKED ISSUE #616: `unwrap` is not stable when supplied with stable RNG
I have been doing some signal analysis and happened to notice when applying unwrap in the below manner that it provides unstable results, even when applied with a stable RNG.

```julia
using STFT, LibSndFile, FileIO, Test, Plots
using DSP: unwrap

wk_data = Array{Float64}(load("anything.wav")[:, 1])

fft(sig) = 
	STFT.analysis(wk_data, DSP.hanning(64), 32, 192)

Test.@test sum(rand(StableRNG(10), 100) - rand(StableRNG(10), 100)) == 0
Test.@test sum(fft(wk_data) - fft(wk_data)) == 0
Test.@test sum(angle.(fft(wk_data)) - angle.(fft(wk_data))) == 0

ft = fft(wk_data)
ϕs = angle.(ft)
	
measure() = 
	sqrt(mean((DSP.unwrap(ϕs; dims=1:2, rng=StableRNG(10))
		     - DSP.unwrap(ϕs; dims=1:2, rng=StableRNG(10))) .^ 2))

es = [measure() for _ in 1:20]
plot(es, label="RMSE")
```

image

And so forth.

I am trying to measure phase information for comparison between samples, which requires the most (ideally completely) stable measurements as are possible. The error between applications is both significant and well-enough distributed as to propagate to other metrics,

Here is one measurement of a signal:
image

And another of the same signal - note the visually distinct changes to both the phase heatmap and the troughs at either end of the entropy measurement as well as the fact that the range of the unwrapped data is now (roughly) `[-175, 75]` as opposed to (again, roughly) `[-25, 175]` :
image

I am only noticing this fairly late, so have not scrutinized the implementation to see if this is to be expected, is caused by some issue with the wrapped phase information, or is an actual issue. I'll also accept the response that I am using `DSP.unwrap` in an unintended manner, or that I should not expect stability, should it include some better suggestion - I usually work much lower in the stack, where the most complex math tends to be memory alignment (humor attempt complete), so I am still becoming acquainted with some of the stuff here.

RomanHargrave:
`init_image()` presently calls `rand()` once per pixel to initialize reliability values. As this is done inside an `@threads` loop, this does not happen in a deterministic order, in turn causing `unwrap` to produce measurably different outputs when executed for the same input and with a prepared PRNG. To remedy this, I moved reliability value generation outside of the loop, which stabilized the output - now there is zero variation between applications (with a fresh PRNG).

Unrelated, but the use of `@threads` here seems like it is of limited benefit - there does not seem to be any computationally intense work inside the loop.


src/unwrap.jl
```diff
@@ -146,8 +146,13 @@ end
...
     pixel_image = similar(wrapped_image, Pixel{T})
-    Threads.@threads for i in eachindex(wrapped_image, pixel_image)
-        pixel_image[i] = Pixel(wrapped_image[i], rng)
+
+    # Initialize reliability values before going parallel. This ensures that
+    # reliability values are generated in a deterministic order.
+    rels = rand(rng, Float64, size(wrapped_image))
+
+    Threads.@threads for i in eachindex(wrapped_image, pixel_image, rels)
+        pixel_image[i] = Pixel(wrapped_image[i], rels[i])
```
martinholters, line 155: Is this `@threads` even worth it or could we just keep the old code but with `Threads.@threads` removed?

martinholters (commented):
> Unrelated, but the use of @threads here seems like it is of limited benefit - there does not seem to be any computationally intense work inside the loop.

Agree. Now it's just the allocation of `Pixel`s that happens threaded, I believe? Is that even beneficial, or does the allocation then serialize things again by using locks anyway, and it's just the trivial initialization that runs in parallel?

src/unwrap.jl
martinholters: Then I'd tend to the simpler implementation, but no strong opinion here.

martinholters (approved):
No idea about the other test changes, but the main point of this PR looks good to me.

PR #618, opened 2024-12-31
Remove recursion from `Base.:^`

wheeheee:
Prevents stack overflow, in the admittedly unlikely case that `-e == e`, for `e == typemin(Int8)` etc.
Also moves some definitions around for readability.

### DIFF
--- a/src/Filters/coefficients.jl
+++ b/src/Filters/coefficients.jl
@@ -48,11 +48,8 @@ Base.inv(f::ZeroPoleGain{D}) where {D} = ZeroPoleGain{D}(f.p, f.z, inv(f.k))
 
 function Base.:^(f::ZeroPoleGain{D}, e::Integer) where {D}
     ae = uabs(e)
-    z, p, k = repeat(f.z, ae), repeat(f.p, ae), f.k
-    if e < 0
-        z, p, k = p, z, inv(k)
-    end
-    return ZeroPoleGain{D}(z, p, k^ae)
+    z, p = repeat(f.z, ae), repeat(f.p, ae)
+    return e < 0 ? ZeroPoleGain{D}(p, z, inv(f.k)^ae) : ZeroPoleGain{D}(z, p, f.k^ae)
 end
 
 #
@@ -331,8 +328,12 @@ Base.inv(f::SecondOrderSections{D}) where {D} = SecondOrderSections{D}(inv.(f.bi
 
 function Base.:^(f::SecondOrderSections{D}, e::Integer) where {D}
     ae = uabs(e)
-    pf = e < 0 ? inv(f) : f
-    return SecondOrderSections{D}(repeat(pf.biquads, ae), pf.g^ae)
+    if e < 0
+        inv_f = inv(f)
+        return SecondOrderSections{D}(repeat(inv_f.biquads, ae), inv_f.g^ae)
+    else
+        return SecondOrderSections{D}(repeat(f.biquads, ae), f.g^ae)
+    end
 end
 
 function Base.:^(f::Biquad{D}, e::Integer) where {D}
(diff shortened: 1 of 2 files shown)


src/Filters/coefficients.jl
```diff
@@ -44,12 +44,16 @@ Base.promote_rule(::Type{ZeroPoleGain{D,Z1,P1,K1}}, ::Type{ZeroPoleGain{D,Z2,P2,
 
 Base.inv(f::ZeroPoleGain{D}) where {D} = ZeroPoleGain{D}(f.p, f.z, inv(f.k))
 
+function checked_abs(e::Integer)
```
martinholters, line 47: Why not use `Base.Checked.checked_abs` instead of rolling our own?

martinholters: Good guess at the name, then :smile:

martinholters: Heh, just stumbled upon https://github.com/JuliaLang/julia/pull/53503 and indeed, `uabs` would be even better, making these cases just work instead of throwing. Says there "`Base.uabs` is already being used by a few packages". So maybe we can join those in the expectation of `uabs` becoming part of the public API.

test/filter_conversion.jl
```diff
@@ -236,6 +236,11 @@ end
             @test filt(H^0, x) ≈ x
         end
     end
+    # test that ^ doesn't cause stack overflow
+    @test_nowarn PolynomialRatio([1.0], [2.0])^typemin(Int8)
+    @test_nowarn ZeroPoleGain([1], [2], 3)^typemin(Int8)
+    @test_nowarn Biquad(1:5...)^typemin(Int8)
+    @test_nowarn SecondOrderSections(Biquad(1:5...))^typemin(Int8)
```
martinholters, line 243: Can we check for the correct result for at least some of those? E.g. something like
```julia
let H = PolynomialRatio([1.0], [2.0])^typemin(Int8)
    @test coefb(H) == [2.0^128]
    @test coefa(H) == [1.0]
end
```
For the `Biquad`/`SecondOrderSections` cases, that might be unwieldy, but at least `PolynomialRatio` and `ZeroPoleGain` should be doable.

test/filter_conversion.jl
```diff
@@ -236,6 +236,24 @@ end
...
+    # test that ^ doesn't cause stack overflow
+    let H = PolynomialRatio([1.0], [2.0])^typemin(Int8)
+        @test coefb(H) == [2.0^128]
+        @test coefa(H) == [1.0]
+    end
+    let zpg = ZeroPoleGain([1], [2], 3)^typemin(Int8)
+        @test length(zpg.z) == length(zpg.p) == 128
+        @test all(==(2), zpg.z)
+        @test all(==(1), zpg.p)
+    end
```
martinholters, line 248: Also test `k`?
```suggestion
    let zpg = ZeroPoleGain([1], [2], 3)^typemin(Int8)
        @test length(zpg.z) == length(zpg.p) == 128
        @test all(==(2), zpg.z)
        @test all(==(1), zpg.p)
        @test zpg.k ≈ inv(3)^128
    end
```

martinholters (approved):
Haven't verified the code movement in detail, but the rest LGTM. Maybe the one additional test. Apart from that, good to go from my side.

src/Filters/coefficients.jl
```diff
@@ -45,11 +47,9 @@ Base.promote_rule(::Type{ZeroPoleGain{D,Z1,P1,K1}}, ::Type{ZeroPoleGain{D,Z2,P2,
...
 
 function Base.:^(f::ZeroPoleGain{D}, e::Integer) where {D}
-    if e < 0
-        return inv(f^-e)
-    else
-        return ZeroPoleGain{D}(repeat(f.z, e), repeat(f.p, e), f.k^e)
-    end
+    ae = uabs(e)
+    res = ZeroPoleGain{D}(repeat(f.z, ae), repeat(f.p, ae), f.k^ae)
+    return e < 0 ? inv(res) : res
```
martinholters, line 52: Would this fix the overflow in computing `k` in the test?
```suggestion
    if e < 0
        return ZeroPoleGain{D}(repeat(f.p, ae), repeat(f.z, ae), inv(f.k)^ae)
    else
        return ZeroPoleGain{D}(repeat(f.z, ae), repeat(f.p, ae), f.k^ae)
    end
```

PR #621, opened 2025-01-23
Expose `resample_filter` optional arguments in `resample`, `FIRFilter`

wheeheee:
Just passes on arguments in the most obvious manner.
Additionally, make some `FIRKernel` fields const, and removed unnecessary `@inbounds`, and added restrictions to `pad_length` in `filtfilt` to prevent oob.
Also wrote some minimal documentation for `resample_filter`. Need some help here, I just guessed the meaning of `rel_bw`...

Should help with #620

PR #623, opened 2025-01-28
Buffer array resample and `reference_data`

wheeheee:
On top of #621

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "DSP"
 uuid = "717857b8-e6f2-59f4-9121-6e50c889abd2"
-version = "0.8.0"
+version = "0.8.1"
 
 [deps]
 Bessels = "0e736298-9ec6-45e8-9647-e4fc86a2fe38"


test/runtests.jl
```diff
@@ -17,6 +17,8 @@ module_tests = [
     :Unwrap => ["unwrap.jl"]
 ]
 
+!(dirname(@__FILE__) in LOAD_PATH) && push!(LOAD_PATH, dirname(@__FILE__))
```
martinholters, line 20:
```suggestion
!(@__DIR__ in LOAD_PATH) && push!(LOAD_PATH, @__DIR__)
```
?
But why is this needed anyway?

src/Filters/stream_filt.jl
```diff
@@ -623,31 +623,39 @@ function filt!(
...
     #        is updated to properly account for pathological arbitrary rates.
+    outLen = outputlength(sf, length(x))
     if Tk <: FIRArbitrary
-        bufLen += 1
+        outLen += 1
     end
-    buffer         = Vector{promote_type(Th,Tx)}(undef, bufLen)
-    samplesWritten = filt!(buffer, self, x)
+    out = Vector{promote_type(Th, Tx)}(undef, outLen)
+    return out
```
martinholters, line 649:
```suggestion
    return Vector{promote_type(Th, Tx)}(undef, outLen)
```

martinholters (commented):
I generally like DRYing up code by pulling out common code to functions, but those then need a descriptive name to maintain readability. I'm not sure that's always the case here. That aside, this PR increases LOC by 49, 6 of which are a new tests, some are additional `using`s in the tests (to make those file runable by themselves, I suppose). But that still leaves quite a few additional lines due to NFC code re-organization, where I'm not 100% sure it's a net improvement. Need to let this sink in a bit more.

Meanwhile, some minor remarks.

src/Filters/stream_filt.jl
```diff
@@ -623,31 +623,39 @@ function filt!(
     return bufIdx
 end
 
-function filt(self::FIRFilter{Tk}, x::AbstractVector{Tx}) where {Th,Tx,Tk<:FIRKernel{Th}}
-    bufLen         = outputlength(self, length(x))
+function filt(self::FIRFilter, x::AbstractVector)
+    buffer = allocate_output(self, x)
+    bufLen = length(buffer)
+    samplesWritten = filt!(buffer, self, x)
+    return checked_ff_output(buffer, self, bufLen, samplesWritten)
```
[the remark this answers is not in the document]
martinholters: But doesn't `length(buffer) == bufLen` always hold when calling `checked_ff_output`? Then it shouldn't be necessary to pass it explicitly.

src/Filters/stream_filt.jl
```diff
@@ -689,24 +694,35 @@ function resample(x::AbstractVector, rate::AbstractFloat, h::Vector, Nϕ::Intege
     _resample!(x, rate, FIRFilter(h, rate, Nϕ))
 end
 
-function _resample!(x::AbstractVector, rate::Real, self::FIRFilter)
+function _resample!(x::AbstractVector{T}, rate::Real, sf::FIRFilter) where T
```
martinholters, line 697:
```suggestion
function _resample!(x::AbstractVector, rate::Real, sf::FIRFilter)
```
no longer needed

src/Filters/stream_filt.jl
```diff
@@ -689,24 +694,35 @@ function resample(x::AbstractVector, rate::AbstractFloat, h::Vector, Nϕ::Intege
...
+end
 
-    y = filt(self, xPadded)
-    @assert length(y) >= outLen
-    length(y) > outLen && resize!(y, outLen)
+function checked_resample_output!(y::AbstractVector, outLen, samplesWritten, ::FIRFilter{Tk}) where Tk<:FIRKernel
+    if !(Tk <: FIRArbitrary)
+        samplesWritten == length(y) || throw(AssertionError("Length of resampled output different from expectation."))
+    end
+    # bufLen == length(y) is the size of the allocated output.
```
martinholters, line 721:
```suggestion
```
`bufLen` is gone, `length(y)` is obvious

martinholters (commented):
LGTM except for minor nits.

test/runtests.jl
martinholters: But shouldn't the individual test files (all) do this? Those that have `using FilterTestHelpers`, that is. You've amended some to be runnable individually, I think, but that would also require this `LOAD_PATH` manipulation, no? But then it doesn't have to go into `runtests.jl`. Or am I missing something?

martinholters (approved):

PR #624, opened 2025-02-06
Update Zenodo badge

wheeheee:
The webhook triggered 3 different deliveries on the most recent release, failed twice after the first. But it's on zenodo anyhow so it doesn't matter.

This Zenodo badge will point to the most recent release as suggested previously so no further need to update this badge afterwards.


martinholters:
> The webhook triggered 3 different deliveries on the most recent release, failed twice after the first. But it's on zenodo anyhow so it doesn't matter.

Creating a release apparently triggers "released", "published", and "created" webhooks. Zenodo correctly acts on the first one it receives and then rejects the other two because "The release has already been received." So just a cosmetic problem.

martinholters (approved):
I meant to do this, but forgot... Thanks for fixing this.

PR #626, opened 2025-02-08
implement `reset!` for FIRInterpolator and tests

wheeheee:
Also bumps version to 0.8.2. Or I don't know, should 0.8.1 be pulled?


martinholters (approved):

martinholters:
> Also bumps version to 0.8.2. Or I don't know, should 0.8.1 be pulled?

A quick 0.8.2 with the regression fixed seems appropriate.

PR #627, opened 2025-02-20
Consistently adopt column signal convention

wheeheee:
Currently for `filt` each column of an array is treated as a separate signal. This isn't the case for `mt_cross_power_spectra` and `mt_coherence`, and I'm not sure if there's a reason for that, so this is just a draft. The way they are implemented now means it is probably more performant going down the columns than rows, but it isn't very impactful other than consistency with the rest of the package. The change is kind of breaking.
Also, the documentation for signal dimensions in `mt_coherence` and `mt_coherence!` was inconsistent.

### DIFF
--- a/src/multitaper.jl
+++ b/src/multitaper.jl
@@ -436,9 +436,16 @@ struct MTCrossSpectraConfig{T,T1,T2,T3,T4,F,T5,T6,C<:MTConfig{T}}
     function MTCrossSpectraConfig{T,T1,T2,T3,T4,F,T5,T6,C}(
         n_channels::Int, normalization_weights::T1, x_mt::T2,
         demean::Bool, mean_per_channel::T3, demeaned_signal::T4, freq::F, freq_range::T5,
-        freq_inds::T6, ensure_aligned::Bool, mt_config::C
+        freq_inds::T6, ensure_aligned::Bool, mt_config::C, override_warn::Bool
     ) where {T,T1,T2,T3,T4,F,T5,T6,C<:MTConfig{T}}
         check_onesided_real(mt_config)  # this restriction is artificial; the code needs to be generalized
+        if n_channels > mt_config.n_samples && !override_warn
+            Base.depwarn("n_channels > n_samples; this is likely a mistake.
+                From v0.9 onwards, each column is interpreted as a separate signal,
+                whereas in v0.8 and earlier, each row was interpreted as a separate signal.
+                To suppress this warning, pass `override_warn=true` to `MTCrossSpectraConfig`.",
+                :mt_cross_power_spectra; force=true)
+        end
         return new{T,T1,T2,T3,T4,F,T5,T6,C}(
             n_channels, normalization_weights, x_mt,
             demean, mean_per_channel, demeaned_signal, freq, freq_range,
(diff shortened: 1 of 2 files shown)


src/multitaper.jl
```diff
@@ -576,26 +575,27 @@ end
     else
         mt_fft_tapered_multichannel!(x_mt, signal, config)
     end
-    x_mt[1, :, :] ./= sqrt(2)
+    @views x_mt[1, :, :] ./= sqrt(2)
```
martinholters, line 578:
```suggestion
    x_mt[1, :, :] ./= sqrt(2)
```
The broadcast uses a `dotview` anyway, and the extra `@views` seems to make things worse, actually. (Apart from the that, this is one of those unrelated changes I'm not exactly happy about.)

src/multitaper.jl
```diff
@@ -576,26 +575,27 @@ end
...
     if iseven(config.mt_config.nfft)
-        x_mt[end, :, :] ./= sqrt(2)
+        @views x_mt[end, :, :] ./= sqrt(2)
```
martinholters, line 580:
```suggestion
        x_mt[end, :, :] ./= sqrt(2)
```
Same here.

martinholters (changes requested):

martinholters:
I like the consistency here. I wonder whether we can do anything to ease the transition. I guess it's quite unlikely that the input is a square matrix, so if one were to update DSP to a version after this PR without adjustments to `mt_*` callees, one would probably see errors due to output dimensions not matching expectations. That is at least better than silently getting different results. But if we could come up with some form of deprecation that would tell people what to do, that would be better. Any ideas?

src/multitaper.jl
martinholters: You're right. Don't know what I benchmarked yesterday, but indeed the `@views` version is faster and doesn't allocate. Sorry for the noise.

martinholters:
> Hm. Well, typically `n_channels` is quite small compared to `n_samples`. We could write a heuristic to detect if it's the other way round and emit a warning if so

Yeah, that would work for the vast majority of cases. Downside: it will be really annoying for any use-case with an unexpected `n_channels` to `n_samples` ratio. First you get a relatively-hard-to-track-down bug after updating DSP.jl, and when you've finally fixed it, you get warnings you can't suppress. Is this case sufficiently unlikely that we don't care? (Serious question. I tend to think so, but cannot quite convince myself.)

> But if we don't like warnings in functions we could put a warning on first launch/precompile if that's possible. Failing that, a bright red/yellow warning in release notes...

I don't think that's going to work. These warnings will go unnoticed too often.

If we really want to avoid breakage, we should support both argument orders, to be selected with a kwarg, say `argument_order`, defaulting to `:legacy`. If it is `:legacy`, we accept the old order, but emit a deprecation warning asking to switch the argument order and specify `argument_order=:consistent`. If it is `:consistent`, we accept the new order. In the release after that, we change the default to `argument_order=:consistent` and stop supporting `:legacy`, and eventually, we remove the kwarg completely. But that's probably over-engineered and we should go with the simpler warning outlined above. Or maybe there is some viable middle ground?

martinholters:
> However, we probably have to force the depwarn because it's off by default and users won't see it...

Ah, true, that would just postpone the user's grief... What if we require the kwarg? Well, then one would need to adjust the call-sites anyway and could switch the argument order, so we only would need to support `argument_order=:consistent`. That would actually be simpler. It's bit like having to pass `i_have_fixed_this=true`. Seems quite stupid at first glance, but might actually not be that bad.

martinholters:
That's actually what I've tried to say :smile: 

> Another alternative: an override for the heuristic warnings?

Hm... Yes, that would allow most new code to get away without passing a redundant kwarg. Downside is that a warning is easier missed than an error. But if you suddenly get bad results, you probably look closer, and then notice the warning. So a clear warning and an `i_know_what_Im_doing=true` kwarg to silence it might be best. Only the probably very few current uses with fewer channels than samples will get bad results without a warning.

martinholters:
No need to rush anything here, but I've added it to the 0.9 milestone so we don't forget about it.

PR #628, opened 2025-03-05
Fix `conv` for empty arrays

### LINKED ISSUE #168: conv corner case
Moved from https://github.com/JuliaLang/julia/issues/20539

`conv( randn(5), randn(0) )`, say
should be, i'd think, should be a vector of length 0
right now it's an error

`conv(randn(0), randn(0))` is trickier, but a vector length 0 is not unreasonable

wheeheee:
Fixes #168 and adds tests

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "DSP"
 uuid = "717857b8-e6f2-59f4-9121-6e50c889abd2"
-version = "0.8.2"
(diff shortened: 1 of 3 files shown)


martinholters (approved):
LGTM, just could maybe use a test for `conv` of two empty arrays. `@test conv(B, B; algorithm) == B` might do the trick?
Should we include a version bump here and release?

PR #634, opened 2025-05-26
Fix type stability

wsmoses:


wsmoses:
@SarahWilliamson7 do you have the error log offhand?

swilliamson7:
I don't think we printed this error to a text file, so no error log offhand that I can locate :/

swilliamson7:
and this will produce the above error (I didn't quite understand if the error output or an example was needed)

```julia
using Enzyme
using DSP

function f(x)

    dsp_x = power(periodogram(x; radialavg=true))
    J = sum(dsp_x)

    return J

end

x = 0.01 .* randn(10, 10)
dx = zeros(10,10)
J = autodiff(
    set_runtime_activity(Enzyme.ReverseWithPrimal),
    f,
    Duplicated(x, dx)
)
```

swilliamson7:
I can confirm it does fix it.

PR #652, opened 2026-01-10
Add window comparison plot to documentation

thimotedupuch:
For the documentation I added a plot with the window functions as well as the frequency response, to give an overwiew of what's available.

---
Thank you very much,
T.D.


wheeheee:
Do you have a screenshot of that part of the docs?

ViralBShah:
I wonder if it is possible to generate this as part of the documentation rather than check it in - so that it is reproducible.

PR #653, opened 2026-01-28
[WIP] Replace FFTW with FFTA.jl as default FFT backend

asinghvi17:
## Summary
Replace FFTW (GPL licensed) with FFTA (MIT).  This still needs a bunch of work and probably a new release of FFTA to bring it up to speed with AbstractFFTs.jl and what we need here, so we can remove hacks in this package

## Changes
- New `src/fft/FFTBackend.jl`: Abstract FFT backend interface
- New `src/fft/FFTAImpl.jl`: FFTA backend implementation  
- New `ext/FFTWExt.jl`: FFTW package extension
- Updated source files to use `FFTBackend` instead of direct FFTW imports

## Usage
```julia
using DSP          # Uses FFTABackend by default
using FFTW         # Automatically switches to FFTWBackend for better Float64 performance
```

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/src/fft/FFTAImpl.jl
+++ b/src/fft/FFTAImpl.jl
@@ -1,0 +1,114 @@
+module FFTAImpl
+
+using FFTA
+using ..FFTBackend
+import ..FFTBackend: AbstractFFTBackend, fft, ifft, ifft!, bfft, rfft, irfft, brfft,
+                     plan_fft, plan_ifft, plan_bfft, plan_rfft, plan_irfft, plan_brfft,
+                     plan_fft!, plan_bfft!
+
+export FFTABackend
+
+"""
+    FFTABackend <: AbstractFFTBackend
+
+FFT backend using FFTA.jl. This is the default backend and works with arbitrary numeric types
+including BigFloat. For better performance with Float32/Float64, use FFTWBackend by loading FFTW.jl.
+"""
+struct FFTABackend <: AbstractFFTBackend end
+
+# Transform implementations
+fft(::FFTABackend, x) = FFTA.fft(x)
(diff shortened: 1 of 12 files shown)


src/fft/FFTAImpl.jl
```diff
@@ -0,0 +1,114 @@
...
+fft(::FFTABackend, x) = FFTA.fft(x)
+bfft(::FFTABackend, x) = FFTA.bfft(x)
+
+function ifft(::FFTABackend, x)
+    result = FFTA.bfft(x)
+    return result ./ length(x)
+end
+
+function ifft!(::FFTABackend, x)
+    # FFTA doesn't have in-place operations, so we compute and copy back
```
baggepinnen: This is not quite true, the in-place version is `mul!` on the corresponding plan object

asinghvi17:
Yeah, that's probably the better approach.  I started this before I realized the API mismatch 😅

ViralBShah:
It would be good to get @stevengj's thoughts on this. I think a lot of people like having non-GPL dependencies as the default, but we should only do it if those are the best. Otherwise, we may want a more flexible system to pick different defaults.

stevengj:
It's perfectly reasonable to have a pluggable backend, and really the whole AbstractFFTs.jl structure needs to be re-done so that this can be done in a composable way (https://github.com/JuliaMath/AbstractFFTs.jl/issues/32)

However, as a default method, I currently have serious accuracy concerns about FFTA.jl: https://github.com/JuliaMath/FFTA.jl/issues/118 (aside from performance questions).

PR #658, opened 2026-04-20
Feature/new windows

### LINKED ISSUE #657: Add more typical window functions (i.e. Blackman-Harris) to the defaults
I recently wanted to use some of the less often used filter windows (in this case Blackman-Harris) by using the DSP.jl library. I'm coming from MATLAB and was used to having it directly available there. Now I found that in general some of the commonly known windows are not available here (Blackman-Harris, Flattop). My first reference for a list of windows would be the corresponding [wikipedia entry](https://en.wikipedia.org/wiki/Window_function). 

Of course one can implement them with the makewindow function, but I guess it would be a nice feature to have them readily available.

Before I proceed with a potential pull request, I wanted to check if this is something the maintainers would like to include and if there are any specific guidelines or preferences for how it should be implemented (else I would just go the same way the other windows are implemented).

daniel-s-w:
As suggested in #657 this PR adds two conventional windows to DSP (`blackmanharris` and `flattop`), so they do not need manual user implementation with `makewindow`. The flattop as well as the Blackman-Harris 4-term window coefficients are the same as the ones in MATLAB. Additionally, the `blackmanharris` function supports the selection of the 3-term minimum sidelobe version (not given in MATLAB).
The `gen_winplots.jl` has been updated such that the negative values for the flattop window are shown.

closes #657


src/windows.jl
```diff
@@ -456,6 +458,58 @@ function blackman(n::Integer; padding::Integer=0, zerophase::Bool=false)
...
+window is a linear combination of three or four trigonometric terms, optimized
+for minimum sidelobe level (at the expense of a wider main lobe). The number of
+terms can be selected with `term` ∈ [3,4]. For `term = 3`, the maximum sidelobe
+level is about -69 dB, while for `term = 4` (the default), it improves to -92
+dB.
+
+The 3-term window `w3(x)` and the 4-term window `w4(x)` are defined by sampling
+the following continuous functions in the range `[-0.5, 0.5]`:
+
+    w3(x) = 0.42323 + 0.49755*cos(2pi*x) + 0.07922*cos(4pi*x)
```
wheeheee: These values differ from those in the Wikipedia entry. Is there something I'm missing, or is the IEEE publication correct and you're using those values?

mbaz: Would it be worthwhile to use `cospi` here?

wheeheee: I think it's already used down below.

daniel-s-w: Yes. The implementation itself is with cospi. The documentation shows the pure function term.

daniel-s-w: > These values differ from those in the Wikipedia entry. Is there something I'm missing, or is the IEEE publication correct and you're using those values?

Ah, right. I missed an important publication by [Nuttall](https://ieeexplore.ieee.org/document/1163506) on this topic. The coefficients currently implemented by me are those used by [Harris in his publication](https://ieeexplore.ieee.org/document/1455106) where he names these windows _Blackman-Harris windows_. He claims that the coefficients lead to the smallest sidelobes for **3- and 4-term** windows (where the terms consist of harmonic cosine functions only).

Nuttall refutes that shortly after and provides his coefficients for the actual lowest sidelobes in a **3- or 4-term** window (together with a bunch of other windows with differently optimized tradeoffs).

Wikipedia now lists under _Blackman-Harris_ window the coefficients of the **3-term** window from Nuttall's publication, which is [referenced](https://ccrma.stanford.edu/~jos/sasp/Three_Term_Blackman_Harris_Window.html) on Wikipedia as the "improved Blackman-Harris window design". Furthermore, the **4-term** window from Nuttall's publication is listed as the _Blackman-Nuttall_ window on Wikipedia.

So there's a lot of potential for confusion :smile: 

Looking at MATLAB, you can see that they only implement the **4-term** windows from
Harris and Nuttall, but they are labeled as such (`blackmanharris` and `nuttallwin`).

daniel-s-w: My suggestion would be to clearly distinguish between the publications and do it similarly to MATLAB:

Keep the coefficients as implemented for both the **3-term* window and the **4-term** window under the `blackmanharris` function. Additionally, I would add a `nuttall` function that also implements both window types (**4-term** and **3-term**) from Nuttall and then refers to the respective publications in the documentation.
This way the default behavior of those functions would be the same as the one from MATLAB (**4-term** only), but with the option to switch to the 3-term version of the windows, which allows the 66% overlap-add.

daniel-s-w:
Done

src/windows.jl
```diff
@@ -456,6 +459,113 @@ function blackman(n::Integer; padding::Integer=0, zerophase::Bool=false)
...
+terms can be selected with `term` ∈ [3,4]. For `term = 3`, the maximum sidelobe
+level is about -70.83 dB, while for `term = 4` (the default), it improves to
+-92.01 dB. Note, that Nuttall proved those coefficients to not be the most
+optimized ones in [Nuttall, A. H. (1981). Some windows with very good sidelobe
+behavior. IEEE Transactions on Acoustics, Speech, Signal Processing, 29,
+84-91](https://ieeexplore.ieee.org/document/1163506).
+
+
+The 3-term window `w3(x)` and the 4-term window `w4(x)` are defined by sampling
+the following continuous functions in the range `[-0.5, 0.5]`:
```
wheeheee:
```suggestion
the following continuous functions in the domain `[-0.5, 0.5]`:
```

Probably carried over from the documentation for other windows, so a quick find and replace should suffice
This also appears to be the reason our coefficients (not just the new ones) don't have alternating signs, unlike those on Wikipedia.
It might be helpful to document that one day.

wheeheee (approved):
Looks fine. Maybe the other maintainers have further comments, so I'll just leave this lying around for 2 weeks or so.

test/windows.jl
```diff
@@ -103,10 +124,14 @@ end
...
+
+    @test_throws ArgumentError blackmanharris(128;term=2)
+    @test_throws ArgumentError nuttall(128;term=2)
 end
 
 zeroarg_wins = [rect, hanning, hamming, cosine, lanczos,
-                bartlett, bartlett_hann, blackman, triang]
+    bartlett, bartlett_hann, blackman, blackmanharris,
+    nuttall, triang, flattop]
 onearg_wins = [gaussian, kaiser, tukey]
```
wheeheee, line 134: Ok uh, maybe that was too hasty. There should only be two kwargs, `zerophase` and `padding` for window functions, so that you can provide the argument `term` for `matrix_window`s. 

So `term` should be made a normal arg, or alternatively rewrite these into separate functions e.g. `nuttall4c`, `nuttall3b` as in https://holometer.fnal.gov/GH_FFT.pdf

wheeheee (changes requested):

test/windows.jl
daniel-s-w: Ah. Yes, that is true. Right now only the 4-term windows would have been *matrixable* :)
Thanks for the nice reference - that's a good overview. I would like to adapt the concept of all the variants for the Nuttall windows, since they all have their individually optimized trade-off between sidelobe level and sidelobe fall-off per decade. But I have to admit I do not like the view of 8 different nutall-function-names.
What about the following solution:
- For the `blackmanharris()`, I would change the kwarg `term` to a normal positional parameter with a default value of 4, such that one can call the function without this parameter and have the behavior as in MATLAB.
- For the `nutall()`, I would again go with a term parameter, but it should be a symbol and the documentation then refers to the pdf-report from Heinzel from above. The symbols would be listed as `:_3`, `:_3a`, `:_3b` etc. and again the default would be set to the `:_4c` window to be MATLAB compatible. Then all 8 windows could be included within one `nutall()` function.

wheeheee: That's fine, although it would be kinda ugly and even fewer people would have need for these niche functions...scipy doesn't provide them too. As far as I can see, it's mainly one variant of 3 and 4-terms that are used, 3b and 4c.

daniel-s-w: Ok. Then sticking to the 3b and 4c variants might be less chaotic. But then I would not go for the symbol solution but rather just put the same integer `term` as for `blackmanharris()`.

src/windows.jl
```diff
@@ -632,7 +785,7 @@ const IntegerOr2 = Union{Tuple{Integer, Integer}, Integer}
 const RealOr2 = Union{Tuple{Real, Real}, Real}
 const BoolOr2 = Union{Tuple{Bool, Bool}, Bool}
 
-function matrix_window(func::F, dims::Tuple{Integer,Integer}, arg::Union{RealOr2,Nothing}=nothing;
+function matrix_window(func::F, dims::Tuple{Integer,Integer}, arg::Union{RealOr2,IntegerOr2,Nothing}=nothing;
```
wheeheee:
```suggestion
function matrix_window(func::F, dims::Tuple{Integer,Integer}, arg::Union{RealOr2,Nothing}=nothing;
```

Seems extraneous, as `Integer <: Real`

daniel-s-w: Ah. Well, then the whole new `for` loop afterward is redundant. I'll update this.

test/windows.jl
```diff
@@ -131,6 +156,16 @@ onearg_wins = [gaussian, kaiser, tukey]
         @test winf(9, 0.5, zerophase=true) == ifftshift(winf(19, 0.5)[2:2:end])
     end
 
+    # Test the window functions that have the term arg (number of ceofficients)
```
wheeheee:
```suggestion
    # Test the window functions that accept the `term` arg (number of coefficients)
```

test/windows.jl
```diff
@@ -131,6 +156,16 @@ onearg_wins = [gaussian, kaiser, tukey]
...
+    for winf in termarg_wins
+        @test winf(8, 4, zerophase=true) == ifftshift(winf(9, 4)[1:8])
+        @test winf(9, 4, zerophase=true) == ifftshift(winf(19, 4)[2:2:end])
+    end
+    # 3-term blackmanharris and nuttall are a spocial case because of their odd
```
wheeheee:
```suggestion
    # 3-term blackmanharris and nuttall are special cases because of their odd
```

wheeheee (commented):
Some small typos.

I'm not sure if it's in there somewhere, but how did you generate the reference data? It would be good to put that in comments near where they are loaded.

test/windows.jl
daniel-s-w: > I'm not sure if it's in there somewhere, but how did you generate the reference data? > It would be good to put that in comments near where they are loaded.

This is explained in the _common window_ test.
https://github.com/daniel-s-w/DSP.jl/blob/36e7536d56a1ec4dec28cdfc28dec9f522b89bed/test/windows.jl#L47-L51
I'm adding this line to the _zero-phase windows_ test as well.
Basically there is no real reference, apart from the own calculations. But I'm pretty sure they are correct.

wheeheee: Right, missed that. Final thing (hopefully): v0.8.4 is the current release, so the reference data would actually be generated by the next one (0.8.5). Could you update that? And bump the Project.toml version to "0.8.5", of course.

daniel-s-w:
I'm not 100% sure about the codecov complaint, as the mentioned lines have been there before. Is this due to some coverage checker update?

wheeheee:
Yeah, seems to be. It's the same on master, after bumping setup-julia from v2 to v3. You can rebase onto master and add on from there, hopefully it all goes away.

wheeheee (approved):
LGTM

PR #665, opened 2026-07-09
Reland #639 and minor housekeeping

### LINKED ISSUE #318: `resample` is type-unstable with rational ratios
This is on Julia 1.1 and DSP 0.6.0:

```
julia> @code_warntype resample(randn(100), 1//5)
Body::Any
1 ─ %1 = DSP.Filters.resample_filter::Core.Compiler.Const(DSP.Filters.resample_filter, false)
│   %2 = invoke %1(_3::Rational{Int64}, 1.0::Float64, 60::Int64)::Array{Float64,1}
│   %3 = invoke DSP.Filters.resample(_2::Array{Float64,1}, _3::Rational{Int64}, %2::Array{Float64,1})::Any
└──      return %3
```

Notice the final `resample` call has return type `Any`

wheeheee:
Turns out the `@inline` annotation there was for `FIRRational` and not `FIRInterpolator`, so I benchmarked the wrong thing.
Now the inlining is done for all functions that use that `unsafe_dot` method. The performance boost from this is bigger on 1.13.

Also fixes #318 for constant `ratio` by annotating with `@constprop :aggressive`, which is probably the upper limit of what can be accomplished with the current API.

Plus a few other minor modifications (`cld`) and housekeeping.

### DIFF
--- a/src/Filters/design.jl
+++ b/src/Filters/design.jl
@@ -707,7 +707,7 @@ function _resample_filter(f_nyq, Nϕ, rel_bw, attenuation)
 
     # Round the number of taps up to a multiple of Nϕ.
     # Otherwise the missing taps will be filled with 0.
-    hLen = Nϕ * ceil(Int, hLen / Nϕ)
+    hLen = Nϕ * cld(hLen, Nϕ)
 
     # Ensure that the filter is an odd length
     if iseven(hLen)
--- a/src/Filters/remez_fir.jl
+++ b/src/Filters/remez_fir.jl
@@ -114,12 +114,10 @@ end
 
 Returns `grid`, `des`, and `wt` arrays
 """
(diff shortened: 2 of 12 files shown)


wheeheee:
There is actually a significant jump in performance from optimizing `update!` (and the memory layout of `FIRArbitrary`?).
```julia-repl
julia> out = zeros(Float64, 8 * 10^6);

julia> @btime filt!($out, FF, x) setup=begin
           x = rand(Float64, 2*10^6)
           FF = FIRFilter(2.56)
       end ;
  208.528 ms (0 allocations: 0 bytes)    # master
  166.064 ms (0 allocations: 0 bytes)    # b762d1d
```

Apart from the latest commit, I think the others are quite safe to merge even without reviews.

wheeheee:
Merging in a day or two if there aren't any objections.

